PR #6, opened 2023-06-03
Computing job wait time

benjamin-wilson:
Changed the job wait time to a (simplified) computed value based on the cgminer gekko driver.

Simplified code handling ASIC response 

Added stats to the logging for screenless debugging

### DIFF
--- a/main/miner.c
+++ b/main/miner.c
@@ -25,10 +25,7 @@
 #include "serial.h"
 #include "bm1397.h"
 #include <pthread.h>
-#include "EMC2101.h"
-#include "INA260.h"
 #include <stdint.h>
-#include <math.h>
 
 
 #define PORT CONFIG_STRATUM_PORT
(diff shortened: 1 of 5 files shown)


main/miner.c
```diff
@@ -114,55 +115,61 @@ static void AsicTask(void * pvParameters)
...
 
-                if (nonce_diff > active_jobs[nonce.job_id]->pool_diff)
-                {
-                    //print_hex((uint8_t *)&job, sizeof(struct job_packet), sizeof(struct job_packet), "job: ");
-                    submit_share(sock, STRATUM_USER, active_jobs[nonce.job_id]->jobid, active_jobs[nonce.job_id]->ntime,
-                                 active_jobs[nonce.job_id]->extranonce2, nonce.nonce);
-                }
-            }
+        uint16_t fan_speed = EMC2101_get_fan_speed();
+        float chip_temp = EMC2101_get_chip_temp();
```
johnny9, line 163: Lets not do any additional I/O operations within the asic loop

johnny9 (changes requested):
I think we shouldn't be reading from the i2c devices within the asic loop. Keep that in the system task and avoid extra I/O in the asic loop

PR #10, opened 2023-06-09
Cleanup?

benjamin-wilson:


main/tasks/asic_task.c
```diff
@@ -7,28 +7,16 @@
...
 
     GlobalState *GLOBAL_STATE = (GlobalState*)pvParameters;
 
-
     uint8_t buf[CHUNK_SIZE];
     memset(buf, 0, 1024);
 
     uint8_t id = 0;
 
-    active_jobs = malloc(sizeof(bm_job *) * 128);
```
johnny9, line 23: active jobs is used to maintain all of the recent jobs sent to the asic and the nonce packets returned can then be mapped back to this list.

johnny9 (changes requested):
We need to maintain the list of "active" or recent jobs sent to the asic since the nonce packets returned are not always the last job sent

PR #13, opened 2023-06-22
Nonce checking

developeralgo8888:
update


benjamin-wilson:
I think this probably belongs in a JaBIT repo along with any binaries you choose to build, not in the Bitaxe repo.

johnny9:
Binary files aren't appropriate for a github repo. If you can transcribe it to markdown or another text based format that would be much better.

johnny9:
Also, rebase your view so that we don't have all of those merge commits

johnny9 (changes requested):
Please use a text-based format like markdown for documentation if you would like to include it in a git repo. Binary files can't be reviewed or diffed in a human readable way.

PR #14, opened 2023-08-18
Idf 5.1 migration

Georges760:


johnny9 (approved):
Tested and verified with my Bitaxe

Georges760:
tested OK here too

PR #25, opened 2023-09-04
WIFI Failure

SatoshyBoy:
The firmware halts when it loses the wifi connection.


johncarl-te (approved):
Only suggestion going forward is to be a bit more clear with your commit message. A good commit message title should be in the "imperitive" mood. In this case it should be something like "Call esp_restart on connection failure". A good way to think about it is say "This change will..." in your head and the following words should be your commit title. 
The body of the commit message should include a longer explanation which in this case can include the issue that is solved by this fix.

SatoshyBoy:
thanks for the advice.

PR #27, opened 2023-09-05
Add retries if the stratum hostname is not found

SatoshyBoy:
After restoring the Wi-Fi connection, the router may still try to connect to the internet. Therefore, it will fail to solve the IP address, and then the *ipaddr returns a NULL pointer.

### DIFF
--- a/main/tasks/stratum_task.c
+++ b/main/tasks/stratum_task.c
@@ -18,7 +18,7 @@
 
 static const char *TAG = "stratum_task";
 static ip_addr_t ip_Addr;
-static bool bDNSFound = false;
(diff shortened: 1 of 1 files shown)


johnny9 (changes requested):
Squash your commits and be more clear with the title. Advice is to use "imperative mood". Basically finish the sentance "This change will...". Something like "add retries if the stratum hostname is not found"

johnny9:
Squash your commits with `git rebase -i origin/master`

SatoshyBoy:
I find another issue in my commit. I will push a fix later

PR #29, opened 2023-09-07
Add retries if the stratum hostname is not found

SatsForFreedom:
After restoring the Wi-Fi connection due to a power outage, the router may still try to connect to the internet. Therefore, it will fail to solve the IP address, and then the *ipaddr returns a NULL pointer. This will crash the firmware when it reads from a NULL pointer.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,13 +21,25 @@ jobs:
       uses: espressif/esp-idf-ci-action@v1
       with:
         esp_idf_version: v5.1
-        target: esp32s2
+        target: esp32s3
         command: GITHUB_ACTIONS="true" idf.py build
         path: '.'
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.9'
+        cache: 'pip' # caching pip dependencies
+    - run: pip install esptool
+    - name: "Create factory/merged image"
+      run: "./merge_bin.sh ./esp-miner-merged.bin"
+    - name: upload esp-miner-merged.bin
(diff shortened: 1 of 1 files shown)


benjamin-wilson:
Why was bDNSFound changed from a bool to an int?

SatoshyBoy:
Because I need 3 states: waiting, success and failure.

main/tasks/stratum_task.c
```diff
@@ -18,7 +18,7 @@
 
 static const char *TAG = "stratum_task";
 static ip_addr_t ip_Addr;
-static bool bDNSFound = false;
+static int8_t bDNSFound = 0;
```
johnny9, line 21: this variable is no longer a bool, give it a clearer name and use and enum for the different states or have a second bool for tracking dns progress

components/bm1397/bm1366.c
```diff
@@ -704,6 +704,11 @@ task_result *BM1366_proccess_work(void *pvParameters)
 
     GlobalState *GLOBAL_STATE = (GlobalState *)pvParameters;
 
+    if (GLOBAL_STATE->valid_jobs[rx_job_id] == 0) {
```
johnny9, line 707: create a new PR for this

johnny9 (changes requested):
Don't create merge commits. When updating your PR rebase instead.

components/bm1397/bm1366.c
johnny9: I merged this piece already. Can you update the PR without it?

PR #30, opened 2023-09-07
Implement BM1366 set  hash frequency command

johnny9:


johnny9:
Update from[ 5f9877f ](https://github.com/skot/ESP-Miner/commit/5f9877f3bff4e957cfa46fc45178b5a33f52e2d0)to[ fdb66fe](https://github.com/skot/ESP-Miner/commit/fdb66fe2ac53b31166e7322ef340a52cf7a2a612)
 - Rewrote function to use an iterator to find the dividers. This is much closer to the actual frequency calculation on the s19xp.

PR #35, opened 2023-09-21
Get ASIC model from NVS

johnny9:
By having the asic model written into flash, we can distribute a single release binary. This will prevent users from uploading the wrong firmware type. The ASIC model can be written into flash using bitaxetool.

```
pip install --upgrade bitaxetool==0.4
cp config.cvs.example config.cvs
vim config.cvs
bitaxetool --port /dev/ttyACM1 --config ./config.cvs
```

### DIFF
--- a/components/stratum/include/stratum_api.h
+++ b/components/stratum/include/stratum_api.h
@@ -57,7 +57,7 @@ void STRATUM_V1_initialize_buffer();
 
 char *STRATUM_V1_receive_jsonrpc_line(int sockfd);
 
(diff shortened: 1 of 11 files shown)


benjamin-wilson:
The asic model option should be removed from the menuconfig

johnny9:
> The asic model option should be removed from the menuconfig

I removed it from menuconfig. I might need to remove from sdkconfig.defaults

johnny9:
Update from[ bfdd307 ](https://github.com/skot/ESP-Miner/commit/bfdd30783b2d5cb254b3a8133b09a85f0b4ac139)to 45d7d59:
  - Split into two commits for clarity

Issue #24, opened 2023-09-03
Serial Synchronization


SatoshyBoy:
Eventually the UART can miss a package and lost the header 0xAA55. Today, I need to manually reset the device to fix that issue.

This is a rare event.

![Screenshot_20230903_155051](https://github.com/skot/ESP-Miner/assets/132848052/6710df77-e2a0-4f60-86cf-51a88b67660b)

PR #48, opened 2023-10-20
Serial Synchronization #24

tommywatson:
Fix for issue #24 

I am unable to test it as I don't have any real hardware, I have attached a test program that runs it through short reads and invalid data.

[main.c.txt](https://github.com/skot/ESP-Miner/files/13048786/main.c.txt)

### DIFF
--- a/components/bm1397/serial.c
+++ b/components/bm1397/serial.c
@@ -94,3 +94,41 @@ void SERIAL_clear_buffer(void)
 {
     uart_flush(UART_NUM_1);
 }
+
+/**
+ * @brief recieve packet with 0xaa 0x55 header
+ * @param data - buffer for received serial data
+ * @param length - length of expected packet
+ */
+void *SERIAL_rx_aa55(uint8_t *data,const int length) {
+    for(int len=0; len < length;) {
+        // wait for a response, wait time is pretty arbitrary
+        int received = SERIAL_rx(data+len, length-len, 60000);
+        if (received < 0) {
+            ESP_LOGI(TAG, "Error in serial RX");
+            return NULL;
+        } else if (received == 0) {
+            // Didn't find a solution, restart and try again
+            return NULL;
+        }
+
+        if (len+received > 2) {
+            // valid start
+            if (data[0] == 0xAA && data[1] == 0x55) {
+                len += received;
+            } else {
+                for(int count = 1; count < len + received; ++count) {
+                    if(*(data+count) == 0xAA) {
+                        // move to head and adjust read length
+                        memmove(data, data+count, len+received-count);
+                        len+=received-count;
+                        break;
+                    }
+                }
+            }
+        } else {
+            len+=received;
+        }
+    }
+    return data;
+}
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,5 +1,5 @@
 {
-    "version": "2.0.0",
+    "version": "2.0.1",
     "tasks": [
         {
             "label": "Build - Build project",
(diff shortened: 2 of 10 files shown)


components/bm1397/serial.c
```diff
@@ -94,3 +94,43 @@ void SERIAL_clear_buffer(void)
...
+                for(int count=1; count < len+received; ++count) {
+                    if(*(data+count) == 0xAA) {
+                        // move to head and adjust read length
+                        memmove(data, data+count, len+received-count);
+                        len+=received-count;
+                        break;
+                    }
+                }
+            }
+        }
```
johnny9, line 130: Dont break lines for else

components/bm1397/serial.c
```diff
@@ -94,3 +94,43 @@ void SERIAL_clear_buffer(void)
...
+            return NULL;
+        }
+
+        if(len+received>2) {
+            // valid start
+            if(data[0] == 0xAA && data[1] == 0x55) {
+                len+=received;
+            }
+            else {
+                for(int count=1; count < len+received; ++count) {
```
johnny9, line 121: spaces between operators

components/bm1397/serial.c
```diff
@@ -94,3 +94,43 @@ void SERIAL_clear_buffer(void)
...
+            ESP_LOGI(TAG, "Error in serial RX");
+            return NULL;
+        } else if (received == 0) {
+            // Didn't find a solution, restart and try again
+            return NULL;
+        }
+
+        if(len+received>2) {
+            // valid start
+            if(data[0] == 0xAA && data[1] == 0x55) {
```
johnny9, line 117: space after if

johnny9 (commented):
I don't believe that this will resolve the issue. I think you still have an issue when receiving a smaller response everything will be off by a couple of bytes. i think potentially an easier solution might be to tighten the timeout to reduce the chance of getting multiple nonce responses but we also need to handle when we receive partial nonce response.

tommywatson:
> I don't believe that this will resolve the issue. I think you still have an issue when receiving a smaller response everything will be off by a couple of bytes. i think potentially an easier solution might be to tighten the timeout to reduce the chance of getting multiple nonce responses but we also need to handle when we receive partial nonce response.

The test program does random short reads/0 reads and it resyncs with the incoming data.

components/bm1397/serial.c
johnny9, line 121:
```suggestion
                for(int count = 1; count < len + received; ++count) {
```

components/bm1397/serial.c
johnny9: if ( ) {
} else {
}

components/bm1397/serial.c
```diff
@@ -94,3 +94,41 @@ void SERIAL_clear_buffer(void)
...
+            return NULL;
+        } else if (received == 0) {
+            // Didn't find a solution, restart and try again
+            return NULL;
+        }
+
+        if (len+received > 2) {
+            // valid start
+            if (data[0] == 0xAA && data[1] == 0x55) {
+                len+=received;
```
johnny9, line 118:
```suggestion
                len += received;
```

johnny9:
Concept ACK.

I will try to replicate the issue on actual hardware to verify before merging.

johnny9:
Please squash commits as well and give a better commit message.  Commit header should be clear with what was worked on and in the imperative mood. So something like, Ensure bm1366 serial data is the expected size.  Then the description can include more details about the error and why. Issue number should go after the description/body of the commit.

Something like this for commit template.
```
<scope>: <short summary>

<optional body/description>

<related issue numbers>
```

johnny9:
Sorry for the delays here. I'll do my best to get this tested soon. This shouldnt have been stuck in PR hell this long.

skot:
This fix got implemented a while back. I'll close.

Issue #50, opened 2023-10-22
Stratum V1 mining.configure sent out of order


checksum0:
Per BIP0310, mining.configure SHOULD be the first message sent to the pool after the connection.

While it should not cause too many problem, it might cause pool to ignore the mining.configure message, creating issue with ASICBoosted shares being submitted.

checksum0:
PR #51 fixes this.

checksum0:
I remarked also that the method only send the mining.configure message and discard the pool reply. Assuming all pools will support 1fffe000 rolling mask is a dangerous assumption, not all pool currently support that mask to begin with.

PR #51, opened 2023-10-22
Fixes mining.configure order #50

checksum0:
Fixed mining.configure order as it should be sent first, per BIP0310


johnny9 (approved):
ACK a7d2bb5cb8b8d2cf56229429976555012b8079f0

Issue #58, opened 2023-11-22
Automatic Fan Control causing BM1366 overheating


ozbibi:
Introduced in v2.0.2, the formula used for automatic fan control incorrectly calculates the fan speed, resulting in potential overheating of the BM1335 on the Bitaxe Ultra.

Looking at the [automatic_fan_speed()](https://github.com/skot/ESP-Miner/blob/v2.0.2/main/tasks/power_management_task.c#L147) function, the minimum fan speed is based on `result` set to 25 when the chip temperature is below 50 degrees c.

In my case, this set the fan speed to ~ 2150RPM (for a Noctua NF-A4x10 5V PWM) on a cold start of the Bitaxe. As the chip temperature iincreases (say from starting at 36c), it reaches 50c at which point `result` is set to 0, causing the fan to stop spinning, and the chip temperature to rise rapidly to the threshold temperature before the fan starts spinning fast enough to cool it down.

Formula should be changed to `result = ((chip_temp - min_temp) / range) * 75 + 25;` for the expected fan speeds to be calculated correctly

PR #59, opened 2023-11-22
Fix formula for automatic fan control and adjust minimum fan speed

ozbibi:
Fixed the formula for the  automatic fan control #58.

Adjusted the minimum fan speed to 33 (vs 25 previously), and min temp to 40 degrees c (vs 50 previously), as tests with a Noctua NF-A4x10 5V PWM at 24 degree c ambient, show that this keeps the chip temperature cooler (~42 degrees c) than the previsous minimums (~51 degrees c).

### DIFF
--- a/main/tasks/power_management_task.c
+++ b/main/tasks/power_management_task.c
@@ -144,11 +144,11 @@ void POWER_MANAGEMENT_task(void * pvParameters)
 }
 
 // Set the fan speed between 33% min and 100% max based on chip temperature as input.
-// The fan speed increases from 33% to 100% proportionally to the temperature increase from 40 and THROTTLE_TEMP
+// The fan speed increases from 33% to 100% proportionally to the temperature increase from 50 and THROTTLE_TEMP
 static void automatic_fan_speed(float chip_temp)
 {
     double result = 0.0;
-    double min_temp = 40.0;
+    double min_temp = 50.0;
     double min_fan_speed = 33.0;
 
     if (chip_temp < min_temp) {
--- a/main/tasks/power_management_task.c
+++ b/main/tasks/power_management_task.c
@@ -161,4 +161,3 @@ static void automatic_fan_speed(float chip_temp)
 
     EMC2101_set_fan_speed((float) result / 100);
 }
-
--- a/main/tasks/power_management_task.c
+++ b/main/tasks/power_management_task.c
@@ -143,20 +143,22 @@ void POWER_MANAGEMENT_task(void * pvParameters)
     }
 }
 
-// Fan speed between 33 min and 100 max based on input,
-// proportional to temp increase between 50 and THROTTLE_TEMP
+// Set the fan speed between 33% min and 100% max based on chip temperature as input.
+// The fan speed increases from 33% to 100% proportionally to the temperature increase from 40 and THROTTLE_TEMP
 static void automatic_fan_speed(float chip_temp)
 {
     double result = 0.0;
     double min_temp = 40.0;
+    double min_fan_speed = 33.0;
 
     if (chip_temp < min_temp) {
-        result = 33;
+        result = min_fan_speed;
     } else if (chip_temp >= THROTTLE_TEMP) {
         result = 100;
     } else {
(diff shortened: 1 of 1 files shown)


main/tasks/power_management_task.c
```diff
@@ -143,20 +143,21 @@ void POWER_MANAGEMENT_task(void * pvParameters)
     }
 }
 
-// all values input below 60 result in 38 but increases exponentially such that an input of 70 results in 100
+// Fan speed between 33 min and 100 max based on input,
+// proportional to temp increase between 50 and THROTTLE_TEMP
```
johnny9, line 147: Does this comment make sense? Did you mean "proportional temp increase between 40 and THROTTLE_TEMP"

main/tasks/power_management_task.c
```diff
@@ -143,20 +143,21 @@ void POWER_MANAGEMENT_task(void * pvParameters)
...
 static void automatic_fan_speed(float chip_temp)
 {
     double result = 0.0;
-    double min_temp = 50.0;
+    double min_temp = 40.0;
 
     if (chip_temp < min_temp) {
-        result = 25;
+        result = 33;
```
johnny9, line 154: replace 33 with a local variable as well

main/tasks/power_management_task.c
```diff
@@ -143,20 +143,21 @@ void POWER_MANAGEMENT_task(void * pvParameters)
...
     } else if (chip_temp >= THROTTLE_TEMP) {
         result = 100;
     } else {
         double range = THROTTLE_TEMP - min_temp;
-        result = ((chip_temp - min_temp) / range) * 100;
+        result = ((chip_temp - min_temp) / range) * 67 + 33;
```
johnny9, line 159: maybe define 67 as fan_range and set it equal to 100 - minimum_fan_speed.

johnny9 (commented):
Concept ACK

The original code does have an error with values at the minimum temp and you code fixes it and appears to do what the original code intended. Have some minor comments. The main thing is to at least fix the comment since that can be very confusing if you have a comment that doesnt match the logic.

benjamin-wilson:
Please change the min temp back to 50 as it is still very cool for the chip and reduces fan noise.

main/tasks/power_management_task.c
ozbibi: I reworded the comment, hopefully it makes more sense

main/tasks/power_management_task.c
ozbibi: Done

main/tasks/power_management_task.c
ozbibi: Done

benjamin-wilson:
Great, i'll test it tonight

ozbibi:
I have tested this successfully with a Noctua NF-A4x10 5V PWM:
- keeping reported chip temperature at 44 degrees c with min fan speed 
- reducing reported chip temperature from a starting chip temperature of 56 degrees c down to 44 degrees c , with an increased fan speed to start with, then gradually slowing down until chip temperature below 50 degrees c.

@benjamin-wilson my Bitaxe ultra feels warmer than the reported chip temperature: at 44 degrees c reported, an infrared thermometer shows 55 degree c for the heating surface temp, and 59 degree c at the back of the board under the BM1366.
The Bitaxe appeared unstable at 55+ degrees chip temperatures.

benjamin-wilson:
@ozbibi Which version of bitaxe ultra do you have? Sounds like V1? V1 won't work with automatic control as the temperature sensor is not in the correct location.

ozbibi:
@benjamin-wilson I indeed have a V1. I'll use manual fan control for it then.

benjamin-wilson:
Checked this out tonight. @ozbibi I think 33% starting is a little aggressive, I tested with 20% and it seemed to work well settling the chip temp at 55C. Could you change it to 20? Then I think it's good to go.

ozbibi:
@benjamin-wilson I have lowered the min fan speed to 20 as requested.

PR #139, opened 2024-03-17
Self_test

benjamin-wilson:
Self test that runs once lifetime (if passing) to test the ASIC, power and fan. TODO is check a returned nonce instead of chip cout. Doesn't work on MAX.


egaytan922:
@benjamin-wilson my LV06 is stuck on the self test: Asic=Pass, Power=Fail, Fan=Pass but now it wont allow me to log in.  I cant log into IP address since it does not connect. Any solutions or how to factory reset this?

Issue #141, opened 2024-03-18
Web UI aways show 100% fan


m6olz:
Web UI aways show that fan is running at 100%

![a](https://github.com/skot/ESP-Miner/assets/122750052/d64f9975-9888-4ab4-87e5-06de99dbd5b8)

dadofsambonzuki:
I think this is a wider issue.

I'm seeing it as "Auto fan always shows the last saved value of the manual setting".

e.g. If you set it manually to 50%, restart, set it to auto, restart, it will show 50%.

PR #147, opened 2024-03-21
Issue #141 resolved: Web UI always show 100% fan

MoellerDi:


benjamin-wilson:
This does appear to be fixed now.
image

PR #155, opened 2024-03-29
fix Best Difficulty can not be > 4.29G

MoellerDi:
best_nonce_diff used to be defined as uint32_t (max value 4294967295). IMHO it should (at least) be defined as double otherwise Bitaxe UI might show a far too low 'Best Difficulty' value (can not be > 4.29G (uint32_t max value)).

I noticed this as public-pool is showing "Your Best Difficulty: 15.46G" but Bitaxe UI claims just "Best Difficulty: 4.29G"


benjamin-wilson:
I think it would make more sense for this to be a uint64_t as that's what is stored in NVS, we also don't need the precision for display. Good catch though!

PR #163, opened 2024-04-13
fix stratum parsing not always counting rejected shares

MoellerDi:
I noticed that IMHO the code was not always counting rejected shares correctly; at least when using public-pool.io.
The UI was always reporting 0 rejected shares even when shares got rejected (as per the log)

I noticed public-pool is responding to an accepted share with something like: 
`{"id":803,"error":null,"result":true}`

and a response to a rejected share is something like: 
`{"id":873,"result":null,"error":[21,"Job not found",""]}`

Reading the code, it seems it was expected to get "result":false in case of rejected shares. However as per https://braiins.com/stratum-v1/docs is seems be normal for the protocol to response with "result":null instead of "result":false for an error.

This PR change the code to simply increase the counter of rejected shares whenever the value of "error:" is not null. It also add some logging for some unhandled stratum messages (like unhandled methods and result parsing).

In order to not confuse users with high numbers of rejected shares it might be a good idea to show just "x percent rejected" instead of the actual number.

### DIFF
--- a/components/stratum/stratum_api.c
+++ b/components/stratum/stratum_api.c
@@ -118,25 +118,32 @@ void STRATUM_V1_parse(StratumApiV1Message * message, const char * stratum_json)
             result = MINING_SET_DIFFICULTY;
         } else if (strcmp("mining.set_version_mask", method_json->valuestring) == 0) {
             result = MINING_SET_VERSION_MASK;
+        } else {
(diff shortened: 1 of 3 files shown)


skot:
this looks good to me! I think we should pull into the next esp-miner release

PR #185, opened 2024-05-26
Avoid miner reboots due to wifi disconnection

BitMaker-hub:
**Improvements**
- Detect a wifi down situation and initiate reconection process 
- Avoid reboot if socket can't be created or had a forced close (5 retries)
- Reset retries when conection is released
- Check returned jsonrpc_line to detect socket force closing situations and reconnect socket

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ html_doc/
 components/components/*
 config.bin
 config.cvs
+.vscode/settings.json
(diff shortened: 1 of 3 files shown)


skot (changes requested):
I think I resolved the DNS issue in #204 .. the rest of these improvements look good to me though! Can you rebase on master so we can merge?

BitMaker-hub:
Done ;)

skot:
@Georges760 does this work with your changes?

PR #189, opened 2024-05-27
refactor: unify merge_bin scripts

tdb3:
Noticed that there was a bit of duplication in the merge_bin scripts.  This PR combines the functionality of merge_bin_update.sh and merge_bin_with_config.sh into merge_bin.sh.

Now running `merge_bin.sh -u <output_file>` is the same as running `merge_bin_update.sh <output_file>` previously.
Running `merge_bin.sh -c <output_file>` is the same as running `merge_bin_with_config.sh <output_file>` previously.

Some considerations/rationale:
 - `merge_bin_update.sh` and `merge_bin_with_config.sh` are simplified to call merge_bin.sh for now (they can be removed later if desired, kept for deprecation path).
 - The default behavior of running `merge_bin.sh <output_file>` was kept the same to minimize disruption to existing developer workflow.  The argument behavior could easily be changed if desired.
 - Additional usage info and error printing were added to increase clarity for developers/users.
 - Addresses and binary names were modularized to increase re-use.


skot:
this LGTM. thanks for the tooling help!

Issue #190, opened 2024-05-29
Pool difficulty is not respected


djkazic:
**Describe the bug**
I see logs that would indicate the miner is not respecting pool diffs.

I was mining to ocean pool (ocean.xyz) which has a minimum of 16384 diff.

Logs:
```
₿ (473198) asic_result: Nonce difficulty 16000.40 of 8192.
₿ (473198) stratum_api: tx: {"id": 6, "method": "mining.submit", "params": ["[redacted].bitaxe", "17169890446666395", "7709000000000000", "66572c74", "1199c280", "00006000"]}
₿ (473258) stratum_task: rx: {"error": [23, "high-hash", null], "id": 6, "result": null}
```

Why is it 8192?

https://github.com/skot/ESP-Miner/blob/00d5a7386f259967c6fdccfcd0cb432d2008e756/main/tasks/asic_result_task.c#L42C120-L42C129

Suggests that the job's `pool_diff` member is being printed.

**To Reproduce**
1. Setup bitaxe
2. Mine to ocean.xyz
3. Watch logs

**Expected behavior**
The diff should be 16834.

**Hardware (please complete the following information):**
 - Bitaxe HW version: 2.2
 - Bitaxe HW vendor: self built
 - ESP-Miner FW version: v2.1.6
 - Hash Frequency: 500mhz
 - Voltage: 1.4
 - Pool URL, Port, User: ocean.xyz

djkazic:
I did contact Ocean mining support, they stated that the first message should be a difficulty setting stratum message.

However, when I do a reboot and watch the websocket, I see that we parse a `mining.notify` only. Perhaps there's an issue with parsing the first message over stratum or something to that effect?

djkazic:
OK yeah I think that's the issue!

I wrote a script to simulate the stratum connection and the set difficulty message is there:
```
> $ python3 fakeasic.py
Connecting to mine.ocean.xyz:3334
Sending: {"id": 1, "method": "mining.subscribe", "params": []}
Sending: {"id": 2, "method": "mining.authorize", "params": ["[redacted].mock", "password"]}
Received: {"error": null, "id": 1, "result": [[["mining.notify", "c60000001"], ["mining.set_difficulty", "c60000002"]], "c6000000", 8]}
Received: {"id": null, "method": "mining.set_difficulty", "params": [16383.75]}
```

djkazic:
[...]

PR #192, opened 2024-05-31
Stratum message ordering fix

skot:
Removed all dependencies on stratum responses coming in order. 
for issue: Pool difficulty is not respected #190


skot:
tested with;
- public-pool
- OCEAN
- CKPool
- AntPool
- BraiinsPool
     - sends `rx: {"id":2,"result":null,"error":[21,"ConfigurationOutOfOrder",null]}` in response to mining.configure
- Luxor
- NiceHash
     - sends `{"id":null,"method":"mining.set_extranonce","params":["3263b370affeb0d550",3]}` later on, which we don;t handle

PR #205, opened 2024-06-06
refactor: split vcore out from ds4432 driver, to make it an abstracti…

Georges760:
…on of whatever hardware is regulating vcore (making room for TPS546)

no functional impact, only code refactoring for improved modularity


WantClue (approved):
looks good to me @skot @benjamin-wilson

PR #209, opened 2024-06-09
Update bm1397.c to increase the max frequency to 650Mhz

yanir99:
The original version was setting everything above 500Mhz to 500Mhz, the update increases the limit to 650Mhz. Added changes to the web interface also.

### DIFF
--- a/components/bm1397/bm1397.c
+++ b/components/bm1397/bm1397.c
@@ -132,7 +132,7 @@ void BM1397_send_hash_frequency(float frequency)
     float deffreq = 200.0;
 
     float fa, fb, fc1, fc2, newf;
-    float f1, basef, famax = 0xf0, famin = 0x10;
+    float f1, basef, famax = 0x104, famin = 0x10;
     int i;
 
     // bound the frequency setting
@@ -142,9 +142,9 @@ void BM1397_send_hash_frequency(float frequency)
     {
         f1 = 50;
     }
-    else if (frequency > 500)
+    else if (frequency > 650)
     {
-        f1 = 500;
+        f1 = 475;
     }
     else
     {
@@ -154,7 +154,7 @@ void BM1397_send_hash_frequency(float frequency)
     fb = 2;
     fc1 = 1;
     fc2 = 5; // initial multiplier of 10
-    if (f1 >= 500)
+    if (f1 > 650)
     {
         // halve down to '250-400'
         fb = 1;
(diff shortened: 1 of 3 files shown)


Georges760:
i am interested to understand better these new value, but no time for it now. I will review it soon.

components/bm1397/bm1397.c
```diff
@@ -142,9 +142,9 @@ void BM1397_send_hash_frequency(float frequency)
     {
         f1 = 50;
     }
-    else if (frequency > 500)
+    else if (frequency > 650)
     {
-        f1 = 500;
+        f1 = 475;
```
Georges760, line 147: this should be 650 also, not 475, e are bounding the value here

components/bm1397/bm1397.c
```diff
@@ -154,7 +154,7 @@ void BM1397_send_hash_frequency(float frequency)
     fb = 2;
     fc1 = 1;
     fc2 = 5; // initial multiplier of 10
-    if (f1 >= 500)
+    if (f1 > 650)
```
Georges760, line 157: i think we should keep the >=500 here, to halve down from [500-650] to [250-325]

readme.md
```diff
@@ -1,6 +1,7 @@
 [![](https://dcbadge.vercel.app/api/server/3E8ca2dkcC)](https://discord.gg/3E8ca2dkcC)
 
 # ESP-Miner
+# Forked from v2.1.6
```
Georges760, line 4: please remove this

Georges760 (changes requested):

Georges760:
Thanks for the changes, before approving, could you tell us how welll your BitaxeMAx can handle overclocking above 500MHz ? Do you have to do some heatsink improvement to handle the over heat ?

components/bm1397/bm1397.c
```diff
@@ -132,7 +132,7 @@ void BM1397_send_hash_frequency(float frequency)
     float deffreq = 200.0;
 
     float fa, fb, fc1, fc2, newf;
-    float f1, basef, famax = 0xf0, famin = 0x10;
+    float f1, basef, famax = 0x104, famin = 0x10;
```
Georges760, line 135: this is not enough, for fa > 0xff, on line 188 below, the value will be truncated to 8bits and stored in freqbuf[3], so you should also adapt the line 188 to not loose the 9th bit (msb)

Georges760 (changes requested):

components/bm1397/bm1397.c
yanir99: updates as discussed

Georges760 (approved):
very good

Georges760 (approved):

PR #210, opened 2024-06-09
feature: add Zabbix template for Bitaxe monitoring

tdb3:
Adds a template and basic instructional readme so that users can monitor the health of Bitaxe devices using the Zabbix NMS.

image


WantClue (approved):
LGTM

tdb3:
Pushed an update which adjusts `MINIMUM_FREE_HEAP` to align with current free heap expectations. Otherwise, intermittent alarm is seen (with a Gamma running 2.4.2). After adjustment, no intermittent alarming was seen.

Issue #218, opened 2024-06-12
No connection to wifis with SSID length of 32 bytes


kisenberg:
**Describe the bug**
If a wifi has a SSID with a length of 32 bytes, Bitaxe is not able to connect. Only connections to wifis with SSID up to a length of 31 bytes are possible.

**To Reproduce**
1. Go to AxeOS settings.
2. Enter 32 bytes long SSID into field of "WiFi SSID".
3. Click "Save".
4. Click "Restart".
5. Bitaxe displays try to connect the wifi forever.

**Expected behavior**
A connection to wifis with SSID of 32 bytes length should be possible, because an SSID can be 32 bytes long.

**Screenshots & Photos**
n/a

**Hardware (please complete the following information):**
 - Bitaxe HW version: Supra 400
 - Bitaxe HW vendor: D-Central
 - ESP-Miner FW version: 2.1.8
 - Hash Frequency: 490 MhZ
 - Voltage: 1166mV
 - Pool URL, Port, User:

**Additional context**
n/a

tommywatson:
Unfortunately it looks like we are limited to 31 characters by the esp-idf wifi libraries, the structure defined the ssid attribute as

`uint8_t ssid[32];                         /**< SSID of target AP. */`

Including the null terminator that limits the SSID to 31 characters.

https://github.com/espressif/esp-idf/blob/8760e6d2a7e19913bc40675dd71f374bcd51b0ae/components/esp_wifi/include/esp_wifi_types_generic.h#L354

kisenberg:
So, there will be no chance to change that from 32 bytes to 33 bytes with the null terminator?

PR #220, opened 2024-06-12
Code clean resulting from looking into #218

tommywatson:
Looking into #218 I found a few issues in the code, fixes include;

- stop crashes when developing on non miner boards
- fix some warnings/errors from invalid strings / char vs const char

### DIFF
--- a/main/global_state.h
+++ b/main/global_state.h
@@ -20,14 +20,16 @@
 
 typedef enum
 {
-    DEVICE_MAX = 0,
+    DEVICE_UNKNOWN = -1,
+    DEVICE_MAX,
     DEVICE_ULTRA,
     DEVICE_SUPRA,
 } DeviceModel;
 
 typedef enum
 {
-    ASIC_BM1397 = 0,
+    ASIC_UNKNOWN = -1,
+    ASIC_BM1397,
     ASIC_BM1366,
     ASIC_BM1368,
 } AsicModel;
@@ -60,7 +62,7 @@ typedef struct
     char best_session_diff_string[DIFF_STRING_SIZE];
     bool FOUND_BLOCK;
     bool startup_done;
-    char ssid[20];
+    char ssid[32];
     char wifi_status[20];
     char * pool_url;
     uint16_t pool_port;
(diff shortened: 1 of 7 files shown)


main/global_state.h
```diff
@@ -60,7 +62,7 @@ typedef struct
     char best_session_diff_string[DIFF_STRING_SIZE];
     bool FOUND_BLOCK;
     bool startup_done;
-    char ssid[20];
```
Georges760, line 63: hooo shit, did we had this bad bug !!!! I hate C letting you down on this simple yet very common problem.

good spotted there

main/global_state.h
```diff
@@ -20,14 +20,16 @@
 
 typedef enum
 {
-    DEVICE_MAX = 0,
+    DEVICE_UNKNOWN = -1,
```
Georges760, line 23: that's a very good improvement, I wanted to do it, thx

main/main.c
```diff
@@ -47,6 +47,10 @@ void app_main(void)
     } else {
         ESP_LOGE(TAG, "Invalid DEVICE model");
         // maybe should return here to now execute anything with a faulty device parameter !
+        // this stops crashes/reboots and allows dev testing without an asic
+        GLOBAL_STATE.device_model = DEVICE_UNKNOWN;
+        GLOBAL_STATE.asic_count = 1;
```
Georges760, line 52: why not putting 0 default value here, instead of assuming there is a actual chip in a actual voltage domain, better shutting everything down by using 0 default value

Georges760 (approved):

Georges760:
i did approve even if we still can discuss the 0 default value for asiccount and voltagedomain parameter in case the devicemodel is unknown

tommywatson:
> i did approve even if we still can discuss the 0 default value for asiccount and voltagedomain parameter in case the devicemodel is unknown, because it is mergeable with this assumed 1 value for now...

My bad, it should have been -1 as a canary, 0 is fine with me too.
Thanks for the review.

main/main.c
tommywatson: My bad, it should have been -1 as a canary, 0 is fine with me too.

PR #222, opened 2024-06-13
Fixed fan speed web update #141

tommywatson:
These changes fix fan rpm/percent requested and update both on the web


Georges760 (approved):
LGTM

skot:
this isn't working on my 401. On first loading up AxeOS I see the following; (my fan is definitely spinning)
image

If I turn off auto fan speed it shows this;
image

skot (changes requested):
I flashed this PR to my 401 and loaded up AxeOS. everything looks good except the fan speed percentage gauge is showing 0%

skot:
nice, this seems to be working on 204 and 401

skot:
we need to hide the Voltage Regulator Temperature on everything but 207, 30x, and 402, but maybe that's another issue. 
image

tommywatson:
> we need to hide the Voltage Regulator Temperature on everything but 207, 30x, and 402, but maybe that's another issue. image

I think that should be separate, I don't know what the max rpm is on the fans, so with some testing we should hone the RPM in to closer match the "Fan %" reading.

skot (approved):
LGTM

skot:
I added a separate issue for the vreg temp; https://github.com/skot/ESP-Miner/issues/237

Yeah, I was noticing that the fan RPM doesn't match the percentage. My fan from the screenshot is 5k max RPM, so 4503RPM should be 90%

I think the issue is that 90% refers to the PWM duty cycle, which doesn't really map to fan speed percent directly. Pretty sure the EMC2101 fan controller we use has some features to sort this out.

tommywatson:
It's the 20000 on this line, it sounds like it should be set to something just above 5k?
https://github.com/skot/ESP-Miner/blob/a3f381862e3496dc0f68aecebbc8fd39fb10747d/main/http_server/axe-os/src/app/components/home/home.component.html#L148

skot:
I think that will make the UI line up a little bit better. But it doesn't change the fact that 4503 isn't 58% of 5000

PR #228, opened 2024-06-17
try to explain nonce space duration from paramters

Georges760:
try to calculate the asic_job_frequency_ms dynamicaly, to be discussed

### DIFF
--- a/components/bm1397/bm1366.c
+++ b/components/bm1397/bm1366.c
@@ -684,7 +684,7 @@ task_result * BM1366_proccess_work(void * pvParameters)
 
     uint8_t job_id = asic_result->job_id;
     uint8_t rx_job_id = job_id & 0xf8;
-    ESP_LOGI(TAG, "RX Job ID: %02X", rx_job_id);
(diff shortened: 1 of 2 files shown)


skot (approved):
This is looking good to me. Tested on 204, 401 and 402. Way less invalid work and rejected shares

Georges760:
I will try it on 302 @WantClue sent me, then merge it if OK

skot:
My bitaxeMax (v2.2) isn't working. It never starts hashing. It's been a _long_ time since I tested this, so it might not be this branch causing the problem..

skot:
> My bitaxeMax (v2.2) isn't working. It never starts hashing. It's been a _long_ time since I tested this, so it might not be this branch causing the problem..

fixed this with e7a29f8. 
Tested on 102 (v2.2), 204, 401, and 402

Georges760:
regression came from #192

Georges760:
we can merge it now as it fix BitaxeMax, the Hex will be tested on top of it

PR #231, opened 2024-06-19
Changes efficiency metric display in AxeOS

dadofsambonzuki:
Fixes https://github.com/skot/ESP-Miner/issues/230


Georges760:
I wanted to do it for long time too... well done

To the none physicist out there, when you divide a Power in W (Voltage in V * Current in A) unit by a Hashrate in H/s (Hash per second) unit, you end up with an efficiency in Ws/H and 1 Watt during 1 second is equal to 1 Joule.

Issue #213, opened 2024-06-10
AxeOS based firmware updates fail


skot:
In some cases the AxeOS dashboard firmware updater fails, leaving the Bitaxe in a state where it does not boot. 
Usually the failure happens during the `www.bin` image update. 
In all of these cases (afaik) the bitaxe can be recovered with a USB firmware flash.

tdb3:
Do users encountering the issue see a particular error response (e.g. one of the 500 reason phrases returned)?  Could help see what error is encountered frequently.

If this is the code being exercised (e.g. on a www.bin update), then it would make sense that the Bitaxe is left in a state where it doesn't boot.  The partition is being erased optimistically with the assumption that the subsequent write/update will work.  If it doesn't, there wouldn't be a www.bin to load on next boot.

https://github.com/skot/ESP-Miner/blob/11107a3d321e6d69b85cf6889d11ce9961446ed3/main/http_server/http_server.c#L422-L457

Not sure if we have enough space for this, but some devices handle this by having two partitions and flipping between the two.  Updates are written to the partition not currently being used.  On next boot, the updated partition is tried.  If loading is unsuccessful, execution falls back / fails safe to the partition that still functions, and the update can be tried again.

One alternative, if space is a luxury, would be to have three partitions, two that are very minimal in size to support bare-bones recovery, and the last to contain the bulk of the app.

WantClue:
With recovery page this is been closed as fixed.

PR #232, opened 2024-06-19
fix: add recovery page

tdb3:
This PR enhances safety for www.bin updates by providing a failsafe web interface in the event of an update failure occurring in the normal AxeOS settings page.  This enables a user to attempt another, or different, www.bin update without having to connect with USB.

Aims to provide at least a partial fix for Issue #213.
The recovery page is accessible at `/recovery` when there is a working www partition and `/` and `/recovery` when the www partition is borked.

recovery_page

Rationale/Notes:
 - Updating with esp-miner.bin is currently safer than updating with www.bin (partitions `factory`, `ota_0`, and `ota_1` include failsafe mechanisms).
 - Currently, when www.bin is updated, it's all or nothing.  The update either works or there is no longer a working web interface (since the sole `www` partition is overwritten during the update).
 - Rather than try to create multiple `www` partitions (space being a luxury), the approach was to implement "just enough" of a minimal interface to enable the user to recover.
 - Decided not to include esp-miner.bin uploading in the recovery page for now because if the device's factory/ota_0/ota_1 is messed up, the web HTTP server may not work at all in the first place.
 - Why include the recovery page in C rather than the AxeOS webapp?  If the www partition is messed up, the web interface is unavailable.  This implementation enables the recovery page to be shown even when any/all of the AxeOS webapp is broken
 - A future enhancement for updates in general would be integrity checking the update files (e.g. update files include a signature of a hash of the partition contents).  This recovery page would still be compatible (since other types of issues can occur, such as power bumps during update).

Initially tested by:

```
mv www.bin www.bin.backup
echo "a messed up www.bin file" > www.bin
```
Uploaded the broken www.bin through the AxeOS settings page, then restarted.
At this point the default route presents the recovery page.
```
mv www.bin.backup www.bin
```
The valid www.bin was then uploaded, and once complete, restart button clicked.
The original AxeOS web interface is restored.


skot:
Looked over the changes, tried the whole process with the messed up www.bin file and then recovering it... works very well! I support merging this in for the 2.1.9 release. @Georges760 want to take a look?

WantClue:
Tested works very well.

tdb3:
If I get time today, I might try a riskier test like purposefully interrupting a www.bin update (e.g. turn off PC WiFi).

skot:
> If I get time today, I might try a riskier test like purposefully interrupting a [www.bin](http://www.bin) update (e.g. turn off PC WiFi) then seeing that recovery still works.

I yanked bitaxe power mid-update. Recovery page came up and I was able to recover the bitaxe.. this is sweet!

PR #250, opened 2024-07-02
Several changes and fixes!

rvdgracht:
Hi!

I've designed and build my own miner based on a ESP32C6 with an asic chain consisting of 4 x BM1366. I've
ben running ESP-Miner for a while. I had to make some changes, and I want to contribute them back to the community.

This is about half of the changes I've made. The other half moves all Bitaxe (board) specific code in a separate file with a generic API so more boards/hardware support can more easely be added. Let me know if you're interested in that
part as well.


rvdgracht:
I  found a bug in my changes to the DS4432 driver which I fixed in the last commit, and a merge fault which is now fixed as well.

BeeEvolved:
Awesome work! Thank you for the contribution

pixeldoc2000:
@rvdgracht 
This might be OT, but do you have any documentation about your miner design with 4 x BM1366 ?

rvdgracht:
> @rvdgracht This might be OT, but do you have any documentation about your miner design with 4 x BM1366 ?

@skot and @pixeldoc2000, I just created a Github repository for the Hardware. You'll find some details, photos and schematics there. Let me know what you think :)
The repository is called [Bitcuboid](https://github.com/rvdgracht/Bitcuboid).

Issue #252, opened 2024-07-10
stratum_api in undesired state after wifi timeout


mutatrum:
When the wifi haa a timeout `wifi:bcn_timeout,ap_probe_send_start`, it tries to restart the socket and stratum connection, but some bits are left in a wrong state resulting in the device hashing but almost no shares accepted. There are several issues hiding that prevent a proper reconnection.

It starts with:

```
I (118764501) bm1368Module: Job ID: 08, Core: 77/6, Ver: 0146C000
I (118764501) asic_result: Ver: 2146C000 Nonce D317009A diff 794.2 of 502.
I (118764511) stratum_api: tx: {"id": 32242, "method": "mining.submit", "params": ["bc1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq.bitaxe", "64851638001239e9", "0800000000000000", "668d76e2", "d317009a", "0146c000"]}
I (118766321) bm1368Module: Job ID: 50, Core: 0/4, Ver: 06C08000
I (118766321) asic_result: Ver: 26C08000 Nonce A48C0000 diff 391.6 of 502.
I (118766661) bm1368Module: Job ID: 68, Core: 6/15, Ver: 03F7E000
I (118766661) asic_result: Ver: 23F7E000 Nonce 4CA5000C diff 378.1 of 502.
I (118766731) bm1368Module: Job ID: 68, Core: 72/15, Ver: 0549E000
I (118766741) asic_result: Ver: 2549E000 Nonce 9EC90290 diff 497.3 of 502.
I (118768811) bm1368Module: Job ID: 48, Core: 38/10, Ver: 06A34000
I (118768811) asic_result: Ver: 26A34000 Nonce B50E034C diff 465.9 of 502.
I (118770191) wifi:bcn_timeout,ap_probe_send_start
I (118771741) bm1368Module: Job ID: 58, Core: 49/2, Ver: 05604000
I (118771741) asic_result: Ver: 25604000 Nonce 3F3F0062 diff 339.7 of 502.
I (118772581) bm1368Module: Job ID: 08, Core: 35/0, Ver: 02A60000
I (118772581) asic_result: Ver: 22A60000 Nonce C8CC0146 diff 270.4 of 502.
I (118772701) wifi:ap_probe_send over, resett wifi status to disassoc
I (118772701) wifi:state: run -> init (0xc800)
I (118772701) wifi:pm stop, total sleep time: 103045108713 us / 118771732338 us

[...]

PR #253, opened 2024-07-10
Implement client.reconnect and reset stratum uid on reconnect

mutatrum:
Initial code change for #252, it's running now.

The parsing of the stratum message is a bit messy, with hardcoded id's. I'll see if I can clean up that logic if the current PR works.


mutatrum:
And what's missing is that shares are still being sent even though the socket is not live. This could interfere with the hardcoded id's above. I'm not sure if a share submit message can interleave with the setup messages and mess up the `send_uid` value.

mutatrum:
Not sure why the build on GitHub is failing while it was working on my machine.

mutatrum:
I've verified several wifi timeouts with reconnects, and the pool connection is properly restored. There are still shares submitted while reconnecting, which the pool doesn't really like. I've also seen an `Invalid job nonce found` error, so it needs a guard against sending stale data. But the general pool communication is working over a reconnect now.

PR #259, opened 2024-07-18
Add a blurb about web based administration to the readme

pRizz:
Makes the included web server more discoverable


Georges760 (commented):
I think you can remove the port 80, any web browser will find it ! KISS

PR #276, opened 2024-08-08
TCH - 2.1.9 All-In-One PR

aaron3481:
2.1.9 base for universail build for ultra, supra and hex. with some UI fix and fan control improvement. Bug fix for EMC2302 for fan speed setting.

### DIFF
--- a/main/EMC2302.c
+++ b/main/EMC2302.c
@@ -52,7 +52,7 @@ void EMC2302_set_fan_speed(uint8_t devicenum, float percent)
     uint8_t speed;
 	uint8_t FAN_SETTING_REG = EMC2302_FAN1_SETTING + (devicenum * 0x10);
 
-    speed = (uint8_t) (255.0 * percent);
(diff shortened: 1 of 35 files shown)


main/EMC2302.c
```diff
@@ -0,0 +1,89 @@
...
+    if (invertPolarity) {
+        ESP_ERROR_CHECK(register_write_byte(EMC2302_PWM_POLARITY, 0b00011111));
+    }
+}
+
+// Sets the fan speed to a given percent
+void EMC2302_set_fan_speed(uint8_t devicenum, float percent)
+{
+    uint8_t speed;
+	uint8_t FAN_SETTING_REG = EMC2302_FAN1_SETTING + (devicenum * 0x10);
```
mutatrum, line 53: Don't do arithmetic with constants, this is bound to cause problems. Use ifs or switch and handle out of bound devicenum.

PR #283, opened 2024-08-11
Fix precision for devices reaching TH/s

eandersson:
This fixes the precision for the graph when reaching 1 TH/s+.
image


eandersson:
Updated this to only apply to TH and above.

skot:
Did this functionality get added in #359 ?

eandersson:
> Did this functionality get added in #359 ?

Yep

PR #292, opened 2024-08-16
Gamma support

skot:
Adding support for gamma hardware


skot:
task timing issues aside, this is working pretty well on my 600x. I need to test this on more hardware then I'll merge

skot:
Tested on 102, 201, 202, 204, 401, 402 and 600x

PR #300, opened 2024-08-17
fixed best difficulty

shufps:
Fixes the best difficulty after hashrate improvment


WantClue (approved):
fixes it

skot:
I think this fix uses the pool difficulty to calculate the hashrate and not the ASIC difficulty, which we decided not to do, right?

skot:
@shufps Is this PR no longer relevant? Should I close?

PR #305, opened 2024-08-20
nvs memory fix

shufps:
fixes memory leak when web UI is opened.

web UI requests `/info` which loads strings from `nvs` everytime it is called.

`nvs_close` was not called in all cases.


Georges760 (approved):
LGTM

WantClue (approved):

PR #317, opened 2024-09-04
adding a log lock for the first 25 seconds

WantClue:
adding a lock on the webui Logging function to only propagate data after 25seconds uptime


benjamin-wilson:
Nack, start-up is often the most important to log

WantClue:
> Nack, start-up is often the most important to log

It stilll logs to esp-idf but not to the websocket. The websocket causes so many crashes

benjamin-wilson:
We should fix the crashes then

benjamin-wilson:
Were these crashes fixed?
https://github.com/skot/ESP-Miner/pull/318

Issue #320, opened 2024-09-05
Wifi never goes back to connected after WiFi connection failed


eandersson:
In the UI the status is stuck at something like this after being disconnected from the WiFi and then successfully re-establishing the connection again.
> WiFi Status: Retrying 100

![image](https://github.com/user-attachments/assets/cc0332ee-9add-488c-a7a4-914682e80fe4)

Steps to reproduce
- Connect device to WiFi.
- Restart WiFi Router / AP.
- Reconnect to device. WebUI under Logs should say `Connected!`, but instead says `Retrying: X`.

andersonsilva87:
Give this a shot
https://bit.ly/4eaAnWN
Passcode: changeme
you must have a gcc compiler

Dinar813:
after upgrading to version 2.1.10, I received the same error.
![Снимок экрана 2024-09-06 в 12 06 24](https://github.com/user-attachments/assets/c8d264ae-a0e0-40a5-8caf-632829237cc3)

eandersson:
This is a duplicate of https://github.com/skot/ESP-Miner/issues/289

PR #321, opened 2024-09-05
Fix wifi status not updating after disconnected #320

eandersson:
This should fix https://github.com/skot/ESP-Miner/issues/320

Steps to reproduce
1) Connect device to WiFi.
2) Restart WiFi Router / AP.
3) Reconnect to device. WebUI should now say `Connected!`.

### DIFF
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -75,6 +75,7 @@ static void event_handler(void * arg, esp_event_base_t event_base, int32_t event
         ESP_LOGI(TAG, "Bitaxe ip:" IPSTR, IP2STR(&event->ip_info.ip));
         s_retry_num = 0;
         xEventGroupSetBits(s_wifi_event_group, WIFI_CONNECTED_BIT);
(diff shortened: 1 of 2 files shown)


skot:
This looks great! Is it possible to make it to that the "reveal" button doesn't show up if you still have the default password of `*****` entered?

Can you also add this reveal feature to the WiFi password with the same caveat?

eandersson:
> This looks great! Is it possible to make it so that the "reveal" button doesn't show up if you still have the default password of `*****` entered?
> 
> Can you also add this reveal feature to the WiFi password with the same caveat?

Not fully sure what feature you are referring to here. Can you share a screenshot :D

skot:
> > This looks great! Is it possible to make it so that the "reveal" button doesn't show up if you still have the default password of `*****` entered?
> > Can you also add this reveal feature to the WiFi password with the same caveat?
> 
> Not fully sure what feature you are referring to here. Can you share a screenshot :D

derp I meant to add this comment to #358

skot:
tested on my 401, LGTM.

PR #326, opened 2024-09-06
Protect against negative frequency and voltage values

eandersson:
Testing the new freq code I tried to input some silly values, e.g. `-1` and it caused frequency to ramp up endlessly (potentially bricking the device) due to an int underflow.


skot:
LGTM, thanks.

PR #327, opened 2024-09-06
Increase the max number of log lines stored in browser

eandersson:
I think it's worth a few more kb browser memory to store some additional logs. This can be useful when trying to browse startup logs.

### DIFF
--- a/main/http_server/axe-os/src/app/components/logs/logs.component.ts
+++ b/main/http_server/axe-os/src/app/components/logs/logs.component.ts
@@ -56,7 +56,7 @@ export class LogsComponent implements OnDestroy, AfterViewChecked {
       this.websocketSubscription = this.websocketService.ws$.subscribe({
         next: (val) => {
           this.logs.push(val);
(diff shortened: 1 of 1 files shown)


skot:
have you noticed the scrolling gets all wonky? 

https://github.com/user-attachments/assets/b5619d48-4e84-4530-9e48-9c91201b9638

eandersson:
> have you noticed the scrolling gets all wonky? It looks like the whole thing is scrolling, but it's just adding 2 lines. This is on Safari
> 
>  log_scrolling.mov

I haven't seen this. Is this before and/or after this patch?

eandersson:
I wonder if we need to tune the buffert a bit.

skot:
> > have you noticed the scrolling gets all wonky? It looks like the whole thing is scrolling, but it's just adding 2 lines. This is on Safari
> > log_scrolling.mov
> 
> I haven't seen this. Is this before and/or after this patch?

This was after this patch.. but I think I've seen it in the past too. Might be a Safari bug?

Issue #329, opened 2024-09-07
A pool DNS record won't work if it has a zero in one of its IP octets (e.g. 51.81.0.15)


eandersson:
The current DNS code checks if we have one or more zeros in the IP octets (e.g. 51.81.0.15) and if so fails. Having one or more zero in the IP octets is perfectly valid. This means that any pool that either already has a zero in their IP octet, or in the future change IP is at risk at stop working for all ESP-Miner devices.
```
        if (ip4_addr1(&ip4addr) != 0 && ip4_addr2(&ip4addr) != 0 && 
            ip4_addr3(&ip4addr) != 0 && ip4_addr4(&ip4addr) != 0) {
            ESP_LOGI(TAG, "IP found : %d.%d.%d.%d",ip4_addr1(&ip4addr),ip4_addr2(&ip4addr),ip4_addr3(&ip4addr),ip4_addr4(&ip4addr));
            ip_Addr = *ipaddr;
         }
```

PR #330, opened 2024-09-07
Remove invalid IPv4 validation in DNS code #329

eandersson:
This is a fix for the issue with pool records that have a zero in one or more of their IP octets (e.g. `51.81.0.15`). https://github.com/skot/ESP-Miner/issues/329

The issue is that we are checking if one of the octets is a zero and if we get at least one we return without setting the DNS valid flag. If we don't get a zero we pass, but a zero in an IP octet is a perfectly valid value, so we should be passing as long as we aren't getting a null value.

Of course we could be more pedantic here as there are invalid values we could be looking for, but those are very unlikely to be received as valid DNS results, unless they are intentional (e.g. 127.0.0.1 or 0.0.0.0).

tl;dr this logic isn't really necessary and it should be enough to just remove it. Tested with a few pools and custom DNS records.


skot:
looks good to me!

Issue #41, opened 2023-10-08
stratum method client.reconnect not supported


PSLLSP:
I assume that `BitAxe` miner doesn't support stratum method [**client.reconnect**](https://en.bitcoin.it/wiki/Stratum_mining_protocol#client.reconnect). It is used in pool [MiningRigRentals](https://www.miningrigrentals.com/), to redirect miner from default TCP port (like 3333) to dynamically allocated TCP port. Most well known miners supports this method, for example `cgminer`, `bfgminer`, `gminer` and many others...

Demo configuration, for test. [`cpuminer`](https://github.com/pooler/cpuminer) tested at Linux PC:
```bash
$ cat mrr-test-sha256d-1.sh
#!/bin/sh

POOL="stratum+tcp://eu-01.miningrigrentals.com:3311"
USER="droidMiner.296231"
PASS="BITAXE"

./minerd -a "sha256d" -o "$POOL" -u "$USER" -p "$PASS" $@
```
Show miner version:
```
$ ./minerd -V
cpuminer 2.5.1
 built on Oct  5 2023
 features: x86_64 PHE SSE2 AVX AVX2 XOP
libcurl/7.81.0 OpenSSL/3.0.2 zlib/1.2.11 brotli/1.0.9 zstd/1.4.8 libidn2/2.3.2 libpsl/0.21.0 (+libidn2/2.3.2) libssh/0.9.6/openssl/zlib nghttp2/1.43.0 librtmp/2.3 OpenLDAP/2.5.16
```
Start miner:
```
$ sh mrr-test-sha256d-1.sh

[2023-10-05 22:45:53] Starting Stratum on stratum+tcp://eu-01.miningrigrentals.com:3311
[2023-10-05 22:45:53] 1 miner threads started, using 'sha256d' algorithm.
[2023-10-05 22:45:53] Server requested reconnection to stratum+tcp://eu-01.miningrigrentals.com:52211
[2023-10-05 22:45:54] Stratum requested work restart
[2023-10-05 22:45:55] thread 0: 2097152 hashes, 8222 khash/s
...
```

Notice that miner was redirected from TCP port 3311 to port 52211; this port was allocated dynamically and will change in the future...
Details about protocol are visible when script is started with `-P` parameter:
```
$ sh mrr-test-sha256d-1.sh -P

[2023-10-05 22:57:51] 1 miner threads started, using 'sha256d' algorithm.
[...]

PR #333, opened 2024-09-10
Add support for client.reconnect method in Stratum V1 protocol

bears-fan:
#41 

This adds support for the client.reconnect method. This allows the miner to handle client.reconnect messages sent by pools, setting the miner to reconnect to a different host and port, optionally after a specified wait time (defaults to 1 second). 

Changes:
- added fields to stratum_api.h
- added parsing the json msgs from the CLIENT_RECONNECT method
- on reconnect, close the current connection and connect to the new one
- improved socket handling by marking the socket as invalid (-1) whenever it's closed

Test instructions:

1. Setup the [mock server](https://github.com/bears-fan/mock-stratum-v1-server) and start the server with default settings `python3 server.py`
2. Point the bitaxe at the mock server
3. While the server is running, use the command `reconnect`, enter host `public-pool.io` and port `21496`
4. Follow the bitaxe logs to verify expected usage. Should see a log entry similar to `Received client.reconnect: host=public-pool.io, port=21496, wait=10
`
5. The miner should gracefully close the connection and connect after the wait time specified

### DIFF
--- a/main/tasks/stratum_task.c
+++ b/main/tasks/stratum_task.c
@@ -210,10 +210,33 @@ void stratum_task(void * pvParameters)
                     GLOBAL_STATE->extranonce_str = stratum_api_v1_message.extranonce_str;
                     GLOBAL_STATE->extranonce_2_len = stratum_api_v1_message.extranonce_2_len;
                 } else if (stratum_api_v1_message.method == CLIENT_RECONNECT) {
-                    ESP_LOGE(TAG, "Pool requested client reconnect...");
+                    ESP_LOGI(TAG, "Pool requested client reconnect to %s:%d (wait: %d seconds)",
+                        stratum_api_v1_message.new_host ? stratum_api_v1_message.new_host : "same host",
+                        stratum_api_v1_message.new_port ? stratum_api_v1_message.new_port : port,
+                        stratum_api_v1_message.wait_time);
+
+                    if (stratum_api_v1_message.new_host) {
+                        free(stratum_url);
+                        stratum_url = strdup(stratum_api_v1_message.new_host);
+                    }
+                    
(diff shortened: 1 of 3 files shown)


main/tasks/stratum_task.c
```diff
@@ -210,10 +210,33 @@ void stratum_task(void * pvParameters)
                     GLOBAL_STATE->extranonce_str = stratum_api_v1_message.extranonce_str;
                     GLOBAL_STATE->extranonce_2_len = stratum_api_v1_message.extranonce_2_len;
                 } else if (stratum_api_v1_message.method == CLIENT_RECONNECT) {
-                    ESP_LOGE(TAG, "Pool requested client reconnect...");
+                    ESP_LOGI(TAG, "Pool requested client reconnect to %s:%d (wait: %d seconds)",
+                        stratum_api_v1_message.new_host ? stratum_api_v1_message.new_host : "same host",
+                        stratum_api_v1_message.new_port ? stratum_api_v1_message.new_port : port,
+                        stratum_api_v1_message.wait_time);
+
+                    if (stratum_api_v1_message.new_host) {
```
johnny9, line 218: Should probably check that its actually a new url. Do we know if the new stratum url should persist?.

johnny9:
Have a better commit message. Include the context/module and issue # after the body

johnny9:
Any instructions on how to test?

johnny9 (commented):
Have a better commit message. Include the context/module and issue # after the body.

Also, please include instructions on how to verify.

johnny9:
You might need to break one more loop so that DNS resolution can happen again.

skot:
I don;t think we want to support client.reconnect. This seems like it just makes it way too easy to hijack a miner.

PR #336, opened 2024-09-11
Add stratum fallback support

eandersson:
- Added fallback support.
- DNS will automatically update now in case the pool changes DNS.
- Tuning to TCP code to improve error handling.


eandersson:
Rebased against trunk.

neofutur:
A fallback wifi config could be useful to keep mining while my main wifi is down :)

19201003080114:
> A fallback wifi config could be useful to keep mining while my main wifi is down :)

This is rather pointless. 
It already supports wifi reconnect after disconnecting. 
If your wifi fails, fix your wifi or change ISP. 
No point to make AxeOS a bloatware.

PR #338, opened 2024-09-14
Set default cpu freq to 240mhz

eandersson:
This will increase the performance of the ESP32 CPU slightly by increasing the cpu frequency from 160mhz to the advertised 240mhz.


WantClue:
Any description on why or what this would benefit would be nice.
Also revert the overheat activation.

skot:
@eandersson do you know what ESP32 clock freq we were at before?

Also, can you remove the THROTTLE_TEMP change from this PR?

eandersson:
> @eandersson do you know what ESP32 clock freq we were at before?
> 
> Also, can you remove the THROTTLE_TEMP change from this PR?

Yep. I'll remove it. Not sure how that snuck in. This updates it from 160Mhz to 240Mhz.

eandersson:
We could also expose this to the Web UI.

Issue #342, opened 2024-09-19
Overheat mode should be more obvious in AxeOS


skot:
Right now it is easy to miss that the Bitaxe is in overheat mode in AxeOS. I suggest we put a big red banner at the top of every AxeOS tab.

Barnminer:
I agree. When I assist people with issues, they did not even notice this.

skot:
I'd also like to have the overheat mode checkbox **checked** by default. Then the user can uncheck it to "Disable overheated mode", causing the whole thing to disappear

skot:
@eandersson do you have any time to help with this? I'm hopeless with this front end stuff 😅
the [overheat_mode_improvements](https://github.com/skot/ESP-Miner/tree/overheat_mode_improvements) branch might be a good one to work off

eandersson:
> @eandersson do you have any time to help with this? I'm hopeless with this front end stuff 😅 the [overheat_mode_improvements](https://github.com/skot/ESP-Miner/tree/overheat_mode_improvements) branch might be a good one to work off

Haha I am too, but I can take a stab at it this weekend.

Issue #341, opened 2024-09-19
Overheat mode should show IP on OLED.


skot:
Overheat mode should show the IP address on the OLED so users can load AxeOS for more information.

PR #344, opened 2024-09-20
Overheat mode improvements

skot:
goals;
- #341
    - This is done
- #342
    - This is not done


skot:
#342 was addressed in #359

PR #347, opened 2024-09-24
Allow the password... password

eandersson:
I don't think anyone should use the password... password in anything but non-testing setups, but it is commonly used for testing so we should allow it.


skot:
Do we need to also remove the same lines in main/http_server/axe-os/src/app/components/settings/settings.component.ts ?

skot:
lgtm!

PR #348, opened 2024-09-24
Allow connecting to open WiFi networks

skot:
If the saved password is 0 length, switch authmode to `WIFI_AUTH_OPEN` which will allow us to connect to open / unsecured WiFi networks.

Also print some more debug info when the WiFi connection fails.


skot:
this is working well for me with 600x, 402 and 401

PR #349, opened 2024-09-24
Version mask now based on stratum msg

adammwest:
What: 
addresses https://github.com/skot/ESP-Miner/issues/54
version mask is now adhered to the stratum message

How:
creates set_version_mask functions for all bm chips
introduce new asic function set_asic_difficulty in global_state
during init the stratum version message has not come through so set a initial value 0x1fffe000
add a flag version_mask_given when a new stratum message comes in with a version change
the next jobs views the flag, and if true calls the set version mask and resets the flag
prints a log line with the new roll length

Testing
Compiled
tested on BM1368 
with different values of the version_rolling from the stratum msg and saw the nonces paired with versions rolled over at the correct version limit

### DIFF
--- a/components/asic/bm1366.c
+++ b/components/asic/bm1366.c
@@ -126,7 +126,7 @@ static void _set_chip_address(uint8_t chipAddr)
     _send_BM1366((TYPE_CMD | GROUP_SINGLE | CMD_SETADDRESS), read_address, 2, BM1366_SERIALTX_DEBUG);
 }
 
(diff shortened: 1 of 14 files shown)


Georges760 (approved):
LGTM

skot:
This is cool @adammwest thanks for digging into this! I'm going to try on my 600x and 401 right now.

skot:
this has been working well on 600x and 401 overnight. I'm going to merge.

PR #352, opened 2024-09-25
WebUI fix for Power/Watts circle visualization

WhiteyCookie:
Plz ignore the commits, was uploading into the wrong directory at first


skot:
should the min be 0?

Issue #357, opened 2024-09-27
Settings -  Stratum Password field - feature request


STSMiner1:
Settings - Stratum Password

Some pools allow the user to enter options here (Zergpool) with how you can mine on the pool (solo or pooled) and the coin(s) to mine etc.

When entering text in this field have it so that the user can see what is being entered into this field instead of masking each character with a '*' so that it can be checked for any typo errors before it is saved via the "Save" button.

![stratumpassword](https://github.com/user-attachments/assets/407f75b8-6e98-4d5b-b436-642cef3f8418)

The content of this field would then be hidden when the "Save" button is clicked on followed by "Restart" to update the pool of these changes for the device.

skot:
this has been fixed in #358 and will be in the next release. thanks y'all!

PR #358, opened 2024-09-27
fix: Password visibility toggle #357

mrv777:
Fix for issue #357


skot:
This looks great! Is it possible to make it so that the "reveal" button doesn't show up if you still have the default password of ***** entered?

Can you also add this reveal feature to the WiFi password with the same caveat?

PR #359, opened 2024-09-27
fix: Add overheat alert, knob colors & max, chart dots & label #342

### LINKED ISSUE #342: Overheat mode should be more obvious in AxeOS

mrv777:
Fix #342 and some fixes to the dashboard knobs and chart


skot:
this is beautiful 🙏
One more request;
When the overheat flag is set, we have this checkbox in settings; "Disable Overheat Mode. Make sure to reset Frequency and Voltage." can you make that a red button instead of a checkbox? So people can't miss it.

benjamin-wilson:
@mrv777 just a high level comment - you should avoid calling functions in the templates so that the change detection can run efficiently

Issue #217, opened 2024-06-12
Some Ideas, bugs and hints...


chillingbee:
I have a few ideas or improvements for the Bitaxe firmware in general:

1. A sorting function in the Swarm overview

2. The restart button in the Swarm overview is not working (at least for me)

3. Let the Bitaxe (AxeOs) run without it hashing. So that you can check on the software side whether everything is working (Wifi etc.) or as a cooldown before you unplug it, etc. As an extension of this idea, you can only let the Bitaxe hash for a certain period of time, for example because electricity is cheaper at night or in relation to solar because the sun only shines during the day or whatever. Of course you can regulate this with a timer, but I think that the software solution would be more elegant here.

4. Right at the moment you are able to update the Hex and the Nerdaxe via AxeOs wich is not a good idea at all!!!

Since I am unfortunately not able to implement this myself and don't know how extensive it would all be, I am writing it here as a possible suggestion! Maybe you also think it's good/useful and one of you could implement it...

WantClue:
most of it has been fixed, would be nice if you can open up a new issue on the asic stop enhancement

PR #360, opened 2024-09-28
fix: Add sorting of swarm table

mrv777:
fixes item 1 in #217


mrv777:
This one was a bit more complicated then I thought and needed AI to help with, so may be good to be reviewed well

skot:
I _think_ in the future we're going to 86 swarm in favor of external tools.

mrv777:
Sounds good, I'll leave the swarm page alone for now.  I was just looking through some old issues

eandersson:
Lets drop this one as we are planning on removing this feature.

PR #365, opened 2024-09-28
Enable WiFi 802.11k 802.11v

eandersson:
This should improve AP selection when having multiple APs.


skot:
nice catch. more compatible WiFi is a great thing. 
LGTM!

skot:
- 802.11k helps clients discover neighboring APs, reducing the need for the client to perform off-channel scanning as it makes the decision on the next Access Point to roam to.
- 802.11v BTM allows the infrastructure to make solicited or unsolicited roaming recommendations to the client.  Which the client is free to ignore, and they often do.

PR #367, opened 2024-09-28
Uptime added to Dashboard

stealth2600:
This pull request adds the uptime to the dashboard page under the best difficulty since system session boot.

### DIFF
--- a/main/http_server/axe-os/src/app/app.module.ts
+++ b/main/http_server/axe-os/src/app/app.module.ts
@@ -22,6 +22,7 @@ import { DateAgoPipe } from './pipes/date-ago.pipe';
 import { HashSuffixPipe } from './pipes/hash-suffix.pipe';
 import { PrimeNGModule } from './prime-ng.module';
 import { MessageModule } from 'primeng/message';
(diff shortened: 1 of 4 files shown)


mrv777:
I'm wondering if this would be better in the header?  The header has a lot of empty space.  Then it wouldn't need to be on the logs page too.

Now I'm wondering if most/all of that overview data could be moved to the header.  The wifi status could be an icon with a color (Not sure when you are seeing this screen though if wifi isn't connected so maybe even remove that?).  Model and board could probably be grouped together.  Version is already on the settings page

stealth2600:
> I'm wondering if this would be better in the header? The header has a lot of empty space. Then it wouldn't need to be on the logs page too.
> 
> Now I'm wondering if most/all of that overview data could be moved to the header. The wifi status could be an icon with a color (Not sure when you are seeing this screen though if wifi isn't connected so maybe even remove that?). Model and board could probably be grouped together. Version is already on the settings page

Hi, 

I agree and per our discussion on Discord I made some additional changes. Please review and provide your feedback.

-stealth2600

PR #368, opened 2024-09-29
fix: Misc UI Tweaks

mrv777:
# Misc UI Improvements

## Changes
- Added degree symbol (°) to temperature readings
- Implemented consistent use of hash pipe and capital 'H' for hashrate
- Added expected efficiency for consistent card displays
- Reduced menu width for better UI
- Added average hashrate to chart as a horizontal line
- Only use 2 decimals < 10, 1 decimal < 100, else 0
- Format the shares numbers
- Reduce max data points on chart

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.ts
+++ b/main/http_server/axe-os/src/app/components/home/home.component.ts
@@ -28,6 +28,8 @@ export class HomeComponent {
   public maxTemp: number = 75;
   public maxFrequency: number = 800;
 
+  private dataCounter: number = 0;
+
   constructor(
     private systemService: SystemService
   ) {
@@ -62,10 +64,9 @@ export class HomeComponent {
           backgroundColor: textColorSecondary,
           borderColor: textColorSecondary,
           tension: 0.2,
-          pointRadius: 0,
+          pointRadius: 1,
           pointHoverRadius: 5,
           borderWidth: 2,
-          borderDash: [5, 5]
         }
       ]
     };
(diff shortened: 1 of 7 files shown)


mrv777:
Okay, so the average hashrate line became a lot of commits.  In the end I just made it simply a horizontal line as it looked less cluttered with dots and I think this makes it more clear.  Can explore the backend average hashrate, but I think this is a good solution in the mean time

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -107,27 +121,34 @@ export class HomeComponent {
     };
 
 
-    this.info$ = interval(5000).pipe(
+    this.info$ = interval(2000).pipe(
```
eandersson, line 124: Is this adding a lot of value increasing the frequency of api calls?

mrv777: Hmm, yeah maybe its not worth the extra load/work.  I thought about just making the UI feel more responsive, but unless you bring it down even more I don't think it feels enough different to change it

eandersson:
LGTM only **_nit_** comment is that not sure if I prefer Mhz being moved. I assume that was for setups where we run it at 1Ghz or higher?

mrv777:
> LGTM only **_nit_** comment is that not sure if I prefer Mhz being moved. I assume that was for setups where we run it at 1Ghz or higher?

Yeah, once you go over 999mhz it hits the knob around it
Screenshot 2024-09-30 at 8 25 19 AM

If you want instead, we can format the number to Ghz then.  It just looked crowded to me, but over 999 is the real problem

eandersson (approved):

WantClue:
please add more information to this PR

mrv777:
> please add more information to this PR

I updated the description, let me know if more is needed

PR #370, opened 2024-10-01
Added Uptime Card To Dashboard Page

stealth2600:
Added a new Uptime card to the Dashboard Page that display the uptime in the format Xd Xh Xm Xs.

image


main/http_server/axe-os/src/app/pipes/uptime-since.pipe.ts
```diff
@@ -16,9 +16,9 @@ export class UptimeSincePipe implements PipeTransform {
       const s = Math.floor(seconds % 60)
       
       const dDisplay = d > 0 ? d + ( "d" ) : ""
-      const hDisplay = h > 0 ? h + ( "h" ) : ""
-      const mDisplay = m > 0 ? m + ( "m" ) : ""
-      const sDisplay = s > 0 ? s + ( "s" ) : ""
+      const hDisplay = h > 0 || d > 0 ? h + ( "h" ) : ""
+      const mDisplay = m > 0 || h > 0 ? m + ( "m" ) : ""
+      const sDisplay = s > 0 || m > 0 ? s + ( "s" ) : ""
```
mrv777, line 21: Hmm, not 100% what you are going for here, but you might want something like this:

```
     const dDisplay = d > 0 ? d + ( "d" ) : ""
      const hDisplay = h > 0 || d > 0 ? h + ( "h" ) : ""
      const mDisplay = m > 0 || h > 0 || d > 0 ? m + ( "m" ) : ""
      const sDisplay = s > 0 || m > 0 || h > 0 || d > 0 ? s + ( "s" ) : ""
      ```
      
Unless you do want to hide minutes when it and hours are 0

Comment (mrv777) [reply main/http_server/axe-os/src/app/pipes/uptime-since.pipe.ts:21]: Actually, this may be a little cleaner:

```
      const dDisplay = d > 0 ? d + ( "d" ) : ""
      const hDisplay = h > 0 || dDisplay.length > 0 ? h + ( "h" ) : ""
      const mDisplay = m > 0 || hDisplay.length > 0 ? m + ( "m" ) : ""
      const sDisplay = s > 0 || mDisplay.length > 0 ? s + ( "s" ) : ""
```

stealth2600: > const dDisplay = d > 0 ? d + ( "d" ) : ""
>       const hDisplay = h > 0 || dDisplay.length > 0 ? h + ( "h" ) : ""
>       const mDisplay = m > 0 || hDisplay.length > 0 ? m + ( "m" ) : ""
>       const sDisplay = s > 0 || mDisplay.length > 0 ? s + ( "s" ) : ""

Good suggestion, using .length would work better. I will update the PR to use this logic. Thank you!

mrv777:
Only other comment would be whether you really want to show seconds or not.  Without it, it would be a little less crowded and since the UI only updates every 5 seconds, the delay is noticeable when showing the seconds.  However, maybe people like to know the exact seconds its been up too?  If you don't show seconds, if its <60 seconds, I would just say "Recently" or maybe "Just Now"

stealth2600:
> Only other comment would be whether you really want to show seconds or not. Without it, it would be a little less crowded and since the UI only updates every 5 seconds, the delay is noticeable when showing the seconds. However, maybe people like to know the exact seconds its been up too? If you don't show seconds, if its <60 seconds, I would just say "Recently" or maybe "Just Now"

Yes, I've been debating this myself and I haven't been able to decide which way to go. What other opinion do folks have out there?

skot:
I say leave the seconds out. It's implying a precision we don't actually have.

stealth2600:
> I say leave the seconds out. It's implying a precision we don't actually have.

@skot Yeah, that's a valid point. I'll go ahead and remove the seconds from the uptime indicator and I'll have it say "Just Now" if the system has been up for less than 1 minute. A couple other questions for the team.

1. Do we want to update the "Uptime" value on the "Logs" and "Swarm" pages to use the new Uptime-Since pipe that I created? If yes, then I will delete the DateAgo pipe since it will no longer be needed.

2. Do we all feel that firmware version is the right secondary piece of information to have in the Uptime Card? Part of me feels like free heap memory might be a better choice. Thoughts? If we choose to stay with the firmware version, I'm thinking I will change the label to "Firmware:" as the word "version" is implied; ex. v.2.2.2 in order to save space. Thoughts?

skot:
I do not want to have any automated "phoning home" on the Bitaxe, even if it's totally innocent like a firmware update check.

Issue #374, opened 2024-10-06
Stratum Fallback - Two issues


STSMiner1:
Test Device :  Bitaxe 205

Merge (336) - https://github.com/skot/ESP-Miner/commit/3d15c25a0707f1a7993332554aaba30d8ed7b3ca

Issue 1
Fallback Stratum Password

Not able to view typo errors in this field before saving.

![fallback-issue1](https://github.com/user-attachments/assets/52fad1ca-e0ed-43cc-aa6a-4d9a049372a0)

Issue 2
Fallback Stratum User

Fallback is using the wallet address from Stratum User and not the one set in Fallback Stratum User.

![fallback-issue2](https://github.com/user-attachments/assets/e90e7137-d80e-4041-afb8-c6aeb8b793dd)

STSMiner1:
Commits (both) tested.

Both issues resolved with these commits, now waiting for these commits to be approved and merged with the master branch, thanks guys !

PR #375, opened 2024-10-06
Fix fallback user

eandersson:
Fixes the second issue mentioned in #374 while the first issue is fixed in #379


skot:
This fix is for "issue 2" in #374 and not "issue 1"?

eandersson:
> This fix is for "issue 2" in #374 and not "issue 1"?

Yea - the other fix is here https://github.com/skot/ESP-Miner/pull/379

skot:
this looks good to me

PR #376, opened 2024-10-06
Update gamma config

adammwest:
What
addresses https://github.com/skot/ESP-Miner/issues/369

How
Update  gamma config default
asicvoltage to 1100mv


benjamin-wilson:
1150 is the minimum stable voltage and shouldn't be lowered

skot:
> 1150 is the minimum stable voltage and shouldn't be lowered

At what frequency? My 600x works quite well at 525MHz and 575MHz @ 1100mV

zcht:
The unstable performance is probably due to a poor cooling solution. The Gamma cannot run at the default clock speed of the BM1370 on the Bitaxe, but it runs well at 525 Mhz@1100mV. 

It's really annoying that many retailers ship the Gamma with really poor cooling and then the users complain to the DEVs, who have to provide first level support.

skot:
okay, lets go with 525MHz/1150mV as the default. This needs to be adjusted in the cvs file as well as the AxeOS settings drop downs.

WantClue:
Has been closed due to already adjusted config on gamma and agreement on no need to change

PR #378, opened 2024-10-06
add warnings for consecutive timeout responses (no rx) from the chip

adammwest:
What
fix for https://github.com/skot/ESP-Miner/issues/69

How
change rx timeout to 10s from 60s
when 20s or 2 timeouts occour in a row, present a warning  
rename error handling variables

Compiled, Built and Tested on BM1368
no hashing problems for 10m+
no warning messages during this time


skot:
I did some small cleanup to where those counters and constants are stored, but otherwise this LGTM

PR #380, opened 2024-10-07
http_server: Add a way to set raw NVS parameters

luke-jr:


benjamin-wilson:
This should be under it's own rest endpoint rather than reusing the existing one

luke-jr:
Why? What if a user wants to change a normal one at the same time?

benjamin-wilson:
This is best practice for designing RESTful interfaces and maintaining a single responsibility for a given resource. I suggest putting this under PATCH /api/system/raw

skot:
I'd like to address #385 first.

main/http_server/http_server.c
```diff
@@ -333,6 +327,44 @@ static esp_err_t PATCH_update_settings(httpd_req_t * req)
...
+    }
+
+    cJSON * root = cJSON_Parse(buf);
+    cJSON * item;
+    cJSON_ArrayForEach(item, root) {
+        cJSON * const type_j = cJSON_GetObjectItem(item, "type");
+        if (!type_j) continue;
+        const char * const type = type_j->valuestring;
+        cJSON * const val = cJSON_GetObjectItem(item, "val");
+        if (!val) continue;
```
terratec, line 350: Maybe make the value parameter optional for the erase command?

```
curl -X PATCH http://bitaxe/api/system/raw -H "Content-Type: application/json" -d '{"testabc": {"type": "u16", "val": 123}}'
curl -X PATCH http://bitaxe/api/system/raw -H "Content-Type: application/json" -d '{"testabc": {"type": "erase"}}'
```

PR #381, opened 2024-10-07
Fix compiler warnings

skot:
- get rid of all the compiler warnings.
- Switch to the non-depreciated esp-idf ADC library


skot:
migrated to the non-depreciated I2C library. (That was more complicated, mostly because of TPS546 I2C complexity)

PR #383, opened 2024-10-07
fix: Power section UI on mobile improvement

mrv777:
### Description:

The Power section in the home component currently uses a fixed column layout that may not be optimal for all screen sizes. We need to update the column classes to improve responsiveness and maintain consistency with other sections.

### Current behavior:
The Power section uses `<div class="col-4">` for its column layout.

### Desired behavior:
Update the Power section to use `<div class="col-6 sm:col-4">` for improved responsiveness on small screens


skot:
beautiful!

PR #394, opened 2024-10-08
fix: Simply display just the active pool

mrv777:
### Description:

The Pool section in the home component currently shows both pools' information and not which one is active.  Changing it to just show the active pool and if its the primary or fallback pool.

If primary pool is active
image

If the fallback pool is active
image


WantClue (commented):
Any specific reason for changing the system.service.ts hashrate preset ?

mrv777:
Oh, I can set it back.  It was for my own testing to actually have the chart change instead of a straight line

eandersson (approved):
LGTM

PR #397, opened 2024-10-09
Remove swarm support

eandersson:


skot:
People aren't going to like this. But I still think it's the way to go. Hopefully removing Swarm will encourage contributors to make replacements

19201003080114:
I suggest providing a replacement app/function/whatever BEFORE removing Swarm completely. 
Bitaxe Viewer Chrome extension good but is not the same.

19201003080114:
> People aren't going to like this. But I still think it's the way to go. Hopefully removing Swarm will encourage contributors to make replacements

I suggest providing a replacement app/function/whatever BEFORE removing Swarm completely.
Bitaxe Viewer Chrome extension good but is not the same.

skot:
This band-aid is getting ripped off⚡️

benjamin-wilson:
I don't see why we should remove it rather than improve it

skot:
> I don't see why we should remove it rather than improve it

My sense is that it adds a lot of complexity to the project (visually for noobs, maintenance overhead, and technical) for a feature that could be much better served by a separate app.

I still think this would make an awesome web app... But maybe that can't work because of CORS?

mrv777:
I'll gladly help work on an alternative.  I was working on something a few weeks ago, but yeah ran into CORS issues when I tried to host it somewhere for people to try as it wants to access local ips.  So was going to look into making it an electron app...or maybe a native app, but that's more work ;)

skot:
> I'll gladly help work on an alternative. I was working on something a few weeks ago, but yeah ran into CORS issues when I tried to host it somewhere for people to try as it wants to access local ips. So was going to look into making it an electron app...or maybe a native app, but that's more work ;) Was thinking about an umbrel app, but think someone else was working on one and it leaves everyone without an umbrel out, so ditched that

Electron app would be amazing. Hit up @Derekblueeyes on OSMU about his Umbrel management app (called Pluto)... maybe there is some overlap?

mutatrum:
One way of getting around the CORS issue is to have a websocket api instead of a REST api. Browsers don't enforce CORS on websocket connections.

PR #413, opened 2024-10-16
Add Recovery Image to CI

eandersson:
This will trigger Recovery images to get built and uploaded for each Github Release and Pre-Release. The only difference between the recovery and factory image is that the recovery image has selftest disabled by default.

e.g.
> esp-miner-recovery-201-v2.3.0b6.bin


WantClue (approved):
Can we add 205 to it please ?

PR #415, opened 2024-10-17
fix: If pool not in quicklinks, then don't link

mrv777:
If the pool url isn't in the quicklinks, don't link it as we can't be sure of the correct URL to get to the web UI for that pool


skot:
It seems like unknown pools are still getting linked for me;
image
in this case the link is to `http://192.168.1.234/185.xxx.xxx.51`

mrv777:
I'll close this one as I don't think its valid any more

PR #422, opened 2024-10-22
Change TPS546 undervoltage and shutdown voltage thresholds to better values.

aaron3481:
Change the UV_WARN limit cuz 5.8V is not make sense for Ultra and Supra and Gamma which input voltage is 5V

### DIFF
--- a/main/TPS546.h
+++ b/main/TPS546.h
@@ -18,8 +18,8 @@
 
 /* vin voltage */
 #define TPS546_INIT_VIN_ON  4.8  /* V */
-#define TPS546_INIT_VIN_OFF 4.5  /* V */
-#define TPS546_INIT_VIN_UV_WARN_LIMIT 5.8 /* V */
+#define TPS546_INIT_VIN_OFF 4.0  /* V */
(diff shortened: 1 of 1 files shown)


skot:
for some reason my TPS546 doesn't like this setting!
```
I (903) TPS546.c: Setting VIN_OFF: 4.00
I (913) TPS546.c: Setting VIN_UV_WARN_LIMIT: 4.20
E (913) i2c.master: I2C transaction unexpected nack detected
E (923) i2c.master: s_i2c_synchronous_transaction(872): I2C transaction failed
E (933) i2c.master: i2c_master_transmit(1074): I2C transaction failed
```

skot:
for some reason I get the I2C NACK whenever `TPS546_INIT_VIN_UV_WARN_LIMIT` is set below 5.0V

skot:
We have confirmed with Texas Instruments that the TPS546D24A VIN_UV_WARN register is broken and cannot be set to values low enough for use on 5V Bitaxes. We'll have to set to 0V and ignore.

PR #423, opened 2024-10-22
Move back to solostats

eandersson:
Move back from the pure json view to the new actual human readable view from @mrv777 .


skot:
LGTM! ckstats is looking mighty fine these days...

PR #424, opened 2024-10-22
Fix api not handling errors properly

eandersson:
I found this when implementing the firmware protection that when sending `httpd_resp_send_err` if you return `ESP_FAIL` it will trigger a TCP Reset, hiding the actual error message from the user.


WantClue (approved):
LGTM.

Any test responses before merging ?

eandersson:
> LGTM.
> 
> Any test responses before merging ?

I think this is fine to merge. Rebased to fix some merge conflicts.

WantClue (approved):

PR #426, opened 2024-10-24
Fix ESP-NOW warning by limiting ap max_connections

mutatrum:
This removes the warning on boot:

_warning Affected by the ESP-NOW encrypt num, set the max connection num to ##_

The ESP32 can only handle 17 encrypted keys. There are 7 in use for ESPNOW (CONFIG_ESP_WIFI_ESPNOW_MAX_ENCRYPT_NUM), which leave 10.

Some context here: https://github.com/espressif/esp-idf/issues/11697

### DIFF
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -49,7 +49,7 @@
 /* FreeRTOS event group to signal when we are connected*/
 static EventGroupHandle_t s_wifi_event_group;
 
-static const char * TAG = "wifi station";
+static const char * TAG = "wifi_station";
 
 static int s_retry_num = 0;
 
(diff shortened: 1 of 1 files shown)


components/connect/connect.c
```diff
@@ -102,7 +102,7 @@ esp_netif_t * wifi_init_softap(void)
     strncpy((char *) wifi_ap_config.ap.ssid, ssid_with_mac, sizeof(wifi_ap_config.ap.ssid));
     wifi_ap_config.ap.ssid_len = strlen(ssid_with_mac);
     wifi_ap_config.ap.channel = 1;
-    wifi_ap_config.ap.max_connection = 30;
+    wifi_ap_config.ap.max_connection = 10;
```
eandersson, line 105: Something much lower should be fine right e.g. 3?

mutatrum: I don't know exactly what counts as a connection, is each connecting client a connection? Or devices polling networks, so no idea how quickly one would run into this limit, so I put it just low enough to get rid of the warning.

skot:
LGTM!

PR #427, opened 2024-10-24
Improve error handling for TPS546 i2c communications

mutatrum:
My compiler complained about unused variables in the variables used in the read functions. When I tried to fix it, I noticed on a lot of places there is no error handling, so I tried to improve it a bit. My assumption is that the i2c communications shouldn't fail halfway, and if it fails, something is wrong. With this change it'll abort operations.

I'm not 100% sure if this is correct and I don't have a BitAxe with a TPS546, so this is a bit of a stab in the dark. It is more in line with how the error handling for other i2c components are done.

Write functions have the same issue, where the `esp_err_t` return value is not checked, even with multiple sequential writes. I assume this needs to have the same treatment?

### DIFF
--- a/main/TPS546.c
+++ b/main/TPS546.c
@@ -38,11 +38,13 @@ static i2c_master_dev_handle_t tps546_dev_handle;
 /**
  * @brief SMBus read byte
  * @param command The command to read
- * @param data Pointer to store the read data
+ * @return The data read
  */
-static esp_err_t smb_read_byte(uint8_t command, uint8_t *data)
+static uint8_t smb_read_byte(uint8_t command)
 {
-    return i2c_bitaxe_register_read(tps546_dev_handle, command, data, 1);
+    uint8_t data;
+    ESP_ERROR_CHECK(i2c_bitaxe_register_read(tps546_dev_handle, command, &data, 1));
+    return data;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


mutatrum:
These were the compiler errors. I can set them to warning, but I think having them break the compiler is a good thing.

```
/home/mutatrum/code-workspace/esp-miner/main/TPS546.c: In function 'TPS546_get_frequency':
/home/mutatrum/code-workspace/esp-miner/main/TPS546.c:576:12: error: 'value' may be used uninitialized [-Werror=maybe-uninitialized]
  576 |     freq = slinear11_2_int(value);
      |            ^~~~~~~~~~~~~~~~~~~~~~
/home/mutatrum/code-workspace/esp-miner/main/TPS546.c:572:14: note: 'value' was declared here
  572 |     uint16_t value;
      |              ^~~~~
```

This might be because of me using ESP-IDF 5.3.1, with a slightly newer compiler. Not sure about that.

skot:
after working on #524 I see what you mean about inconsistent error handling. We should really be pushing esp_err_t all the way up and printing helpful output to the log. We just had a user the other day with i2c problems, but the log didn;t say which device was causing it.

mutatrum:
Superseded by #698.

PR #428, opened 2024-10-24
Use the same espressif/idf release for github and vscode

eandersson:


mutatrum:
I see ESP-IDF 5.1 also in `.devcontainer/devcontainer.json`. Not sure how to bump that version.

eandersson:
> I see ESP-IDF 5.1 also in [`.devcontainer/devcontainer.json`](https://github.com/skot/ESP-Miner/blob/35855358ada2af5710302cfdc06f4eb05fb1080f/.devcontainer/devcontainer.json#L24). Not sure how to bump that version.

That line has likely been broken for a while, but most likely not impacting anything, as the ESP version was never hardcoded, but that line for Python is tied to a very specific version of ESP.

WantClue (approved):

PR #429, opened 2024-10-25
Overhaul Axeos theme

benjamin-wilson:
Overhaul AxeOs theme to more closely align to the brand identity guidelines. This is a first iteration. 

image


WantClue:
Looks awesome

WhiteyCookie:
How about an option for different colors? I liked the original blueish color

skot:
OMG this looks amazing. nice work @benjamin-wilson !

I think we should resolve this;
image
Maybe just show the RPM number next to the percentage, without it's own gauge?

skot:
Can we do a tick mark on the gauge to show what the requested voltage is? Something like this?
image

skot:
we could do this with the input voltage too.. showing the nominal in the middle, and then the actual measured voltage with the bar 
image

benjamin-wilson:
At this time there is no easy way to add tick marks to the progress bar. 
https://primeng.org/progressbar

I tried doing it with just css and some divs but it looks a bit busy

PR #430, opened 2024-10-28
Redo swarm

benjamin-wilson:
Redo swarm to use network scanning and local storage rather than storing the set of IP's on device.


benjamin-wilson:
After testing is confirmed working, i'll merge it.

skot:
This is pretty slick, it found my Bitaxe quiver as soon as I loaded the Swarm tab.

Do the Bitaxe stats not auto-update?

benjamin-wilson:
Not at the moment but it could be added. This is just to init the architecture change

PR #438, opened 2024-11-01
replaced default address with a prompt

eyelight:
tldr - This PR changes the initial defaults for both fields to something that indicates to the user that they should replace the value with their btc address.

Rationale - I recently updated from a fw version which did not contain the "Fallback" config fields, and was surprised to see an unfamiliar field with a bitcoin address (not mine) prefilled. After investigating what was going on, I realize that this is the same address in the initial default, which of course I changed when initially setting up the unit. However, being an update, I had forgotten about that initial setup, so, the surprise set off alarm bells. It prompted me to make this change to prevent this experience for updaters, while also remaining clear for initial setup.

### DIFF
--- a/main/Kconfig.projbuild
+++ b/main/Kconfig.projbuild
@@ -44,7 +44,7 @@ menu "Stratum Configuration"
 
     config STRATUM_USER
         string "Stratum username"
-        default "1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa.bitaxe"
+        default "replace-this-with-your-btc-address.bitaxe"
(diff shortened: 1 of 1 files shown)


skot:
This works for me. I think the only time this will come up is for users upgrading from an old version pre-2.3.0. They should already have a primary address configured which will replace these defaults... this means an update won;t stop them from mining.

Defaults for new factory units can still have these defaults set in the config.csv

skot:
note: to test this you need to delete the file `sdkconfig` if present in the project root dir

PR #448, opened 2024-11-03
fix: Temp suffix & 5v marker

mrv777:
### Description:

- Show current suffix for temp points on the chart and add the suffix to the axes.
- Make the average hashrate a little easier to see
- Add a 5.0v marker on the input voltage progress bar

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -99,10 +99,15 @@
                     <div class="col-12">
 
                         <h6>Input Voltage <span style="float: right;">{{info.voltage}} V</span></h6>
-                        <p-progressBar [value]="(info.voltage / 5.5) * 100" [style]="{ height: '6px' }" >
-                            <ng-template pTemplate="content" let-value>
-                            </ng-template>
-                        </p-progressBar>
+                        <div class="relative">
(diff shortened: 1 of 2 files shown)


skot:
I got the following warning from node.js when compiling... I don't think we had this before?

```
(node:69549) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
(Use `node --trace-deprecation ...` to show where the warning was created)
```

skot:
looking good! 
image

mrv777:
> I got the following warning from node.js when compiling... I don't think we had this before?
> 
> ```
> (node:69549) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
> (Use `node --trace-deprecation ...` to show where the warning was created)
> ```

That seems to be coming from the script: only-gzip.js

skot:
> > I got the following warning from node.js when compiling... I don't think we had this before?
> > ```
> > (node:69549) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
> > (Use `node --trace-deprecation ...` to show where the warning was created)
> > ```
> 
> That seems to be coming from the script: only-gzip.js

Ah, ok. I see it happens on master too. I'll figure out where that came from.

PR #449, opened 2024-11-04
small change to only-gzip.js to remove warning

skot:
we were getting the following warning;

```
(node:69549) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
(Use `node --trace-deprecation ...` to show where the warning was created)
```

So I made the change it suggests, and the warning went away.

Seems to be working fine?


benjamin-wilson (approved):
🤷‍♂️ lgtm

PR #450, opened 2024-11-04
Rejected Shares Percentage Information added

WhiteyCookie:
resolved conflicts for new UI. Should be fine now


skot:
hey @WhiteyCookie sorry it took me so long to review this!

I like having the rejected percentage in the UI a lot. A couple requests;
image
1. I think we could do with less precision. probably just one decimal place?
2. Is there some way to cleanly get this all on one line? How about;
- 8 rejected (6.8%)

WhiteyCookie:
1. Less precision 1 decimal place could be tricky. For example, in my case I mine on satoshiradio and really having a ridiculous low amount of rejected shares (~0.05%). So I would at least do 2 decimals.
2. Sure not prob ;)

WantClue (approved):

PR #452, opened 2024-11-04
fix: Move back primary to human readable ckpool

mrv777:
Same as https://github.com/skot/ESP-Miner/pull/423 but for the primary too

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.ts
+++ b/main/http_server/axe-os/src/app/components/home/home.component.ts
@@ -219,7 +219,7 @@ export class HomeComponent {
           return `https://solo.d-central.tech/#/app/${address}`;
         } else if (/solo[46]?.ckpool.org/.test(info.stratumURL)) {
           const address = info.stratumUser.split('.')[0]
-          return `https://solo.ckpool.org/users/${address}`;
+          return `https://solostats.ckpool.org/users/${address}`;
         } else {
           return undefined;
         }


eandersson:
Can we just refactor this to something along these lines? Not tested.

```
function generateQuickLink(stratumURL, stratumUser) {
  const address = stratumUser.split('.')[0];
  
  if (stratumURL.includes('public-pool.io')) {
    return `https://web.public-pool.io/#/app/${address}`;
  } else if (stratumURL.includes('ocean.xyz')) {
    return `https://ocean.xyz/stats/${address}`;
  } else if (stratumURL.includes('solo.d-central.tech')) {
    return `https://solo.d-central.tech/#/app/${address}`;
  } else if (/solo[46]?.ckpool.org/.test(stratumURL)) {
    return `https://solo.ckpool.org/users/${address}`;
  } else {
    return undefined;
  }
}

this.quickLink$ = this.info$.pipe(
  map(info => generateQuickLink(info.stratumURL, info.stratumUser))
);

this.fallbackQuickLink$ = this.info$.pipe(
  map(info => generateQuickLink(info.fallbackStratumURL, info.fallbackStratumUser))
);
```

mrv777:
@eandersson Okay, refactored.  Also, I added http at the beginning if not found in the quicklinks as I found I get a combo url otherwise, and got rid of the or since we will always have a quicklink value now

eandersson (approved):
Thanks! Looks good to me

eandersson:
Tested locally as well and works.

eandersson:
Merging this in as this compliments my previous fix.

PR #453, opened 2024-11-04
fix: Change log font and add colors

mrv777:
## Description
This PR uses a mono font on log entries and parses the ANSI color code at the beginning of messages using a regex.  It also change the log array to objects so the class can be in one property and the text in another and adds the css color codes for each class

## Screenshot
image


STSMiner1:
Nice !

eandersson (approved):

WantClue (approved):

PR #458, opened 2024-11-05
fix: Better mobile styles for swarm

mrv777:
### Description
- Adjusted swarm table styles for better display on mobile
- Limited temp and power decimals
- Adjusted layout padding on mobile

### DIFF
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.html
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.html
@@ -19,6 +19,9 @@
         <button pButton (click)="scanNetwork()" [disabled]="scanning">{{scanning ? 'Scanning...' : 'Automatic Scan'}}</button>
         <button pButton severity="secondary" (click)="refreshList()" [disabled]="scanning">Refresh List ({{refreshIntervalTime}})</button>
     </div>
+    <div class="text-sm md:text-base">
+        Total Hash Rate: <span class="text-primary">{{totalHashRate * 1000000000 | hashSuffix}}</span>
+    </div>
 </div>
 
 <div class="table-container">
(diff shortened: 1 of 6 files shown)


mrv777:
Also added
- Total hashrate of all devices in swarm
- An orange color on temps over 65C to make the danger stand out

mrv777:
Added a bit more
- Disable the refresh button while refreshing
- Better error handling to prevent an error showing [object Object]
- Zero out a device that failed to update since the data is no longer valid and makes it stand out more

mrv777:
@WantClue Any chance this will make it in 2.3.1? or are we at feature freeze for now and will be next version?

main/http_server/axe-os/src/app/components/swarm/swarm.component.html
```diff
@@ -40,8 +50,8 @@
                 <td>{{axe.hashRate * 1000000000 | hashSuffix}}</td>
                 <td>{{axe.uptimeSeconds | dateAgo}}</td>
                 <td>{{axe.sharesAccepted | number: '1.0-0'}}</td>
-                <td>{{axe.power | number: '1.2-2'}} <small>W</small> </td>
-                <td>{{axe.temp}}°<small>C</small></td>
+                <td>{{axe.power | number: '1.1-1'}} <small>W</small> </td>
+                <td [ngClass]="{'text-orange-500': axe.temp > 65}">{{axe.temp | number: '1.0-1'}}°<small>C</small></td>
```
eandersson, line 54: Nit: 66 and above feels a bit too aggressive for me, maybe 68 instead?
image

WantClue:
> Added a bit more
> 
> * Disable the refresh button while refreshing
> * Better error handling to prevent an error showing [object Object]
> * Zero out a device that failed to update since the data is no longer valid and makes it stand out more

Considering it right now. I think this is needed I will test and if it works as expected it gets implemented in 2.3.1

PR #463, opened 2024-11-09
Fix serial race conditions on startup

eandersson:
This was a change suggested by Mecanix


mutatrum:
This is related regarding the TPS546 code: #427.

components/asic/serial.c
adammwest, line 1: if (wait) {
        if (uart_wait_tx_done(UART_NUM_1, 1000 / portTICK_PERIOD_MS) != ESP_OK) {
            return ESP_FAIL;
        }
    }


could be 

if (wait) {
       esp_err_t tx_err =  uart_wait_tx_done(UART_NUM_1, 1000 / portTICK_PERIOD_MS) 

        if (tx_err!= ESP_OK) {
            return tx_err;
        }
    }

eandersson: I went with this for now
```
    if (wait) {
        // When using wait we could end up returning a ESP_ERR_TIMEOUT error as well.
        return uart_wait_tx_done(UART_NUM_1, 1000 / portTICK_PERIOD_MS);
    }
```

WantClue (approved):

PR #468, opened 2024-11-10
Enable PSRAM

mutatrum:
This change in `sdkconfig.defaults` enables  PSRAM for the ESP32-S3-WROOM-1 N16R8 variant. This adds 8MB of ram to FreeRTOS.

Assumption is that all manufactured BitAxes are using this variant. If there are BitAxes with the 4MB variant, or with no PSRAM, it will probably crash.

The most defensive option is to make sure this code will work on all variants. The other option is to push it out and see what breaks.


mutatrum:
See https://github.com/skot/ESP-Miner/pull/434#discussion_r1835692625 for why some settings have changed or removed.

eandersson:
I know that we are setting something similar for the multichip design https://github.com/WantClue/esp-miner-multichip/blob/main/sdkconfig.defaults#L12

WantClue:
please resolve conflict

mutatrum:
Done

WantClue (approved):

eandersson (approved):

PR #469, opened 2024-11-10
feat: easter egg

mrv777:
## Description

- Just a fun easter egg


benjamin-wilson:
We've got limited space in the www partition - I'd be wary of anything unnecessary

benjamin-wilson:
Perhaps a string that prints to the browser console when you open dev tools?

mrv777:
Sure, I can do that instead

mrv777:
If we want to save space:
https://github.com/skot/ESP-Miner/pull/470

skot:
I think we should keep this

eandersson:
LGTM 🗡️

PR #471, opened 2024-11-11
Set proper size in hex2bin call

### LINKED ISSUE #302: Possible Memory Corruption
The `hex2bin` len parameter takes the destination length that is `HASH_SIZE` and not the source length.

https://github.com/skot/ESP-Miner/blob/80e72cc87443b22ba5920f775a97da69621b5975/components/stratum/stratum_api.c#L238

The last hex string converted overwrites 32 bytes of memory not allocated

mutatrum:
Fixes #302

It actually doesn't corrupt memory, as hex2bin stops on null termination of the string provided by cJSON. But it's still better to have the correct len value passed to hex2bin.


tdb3 (approved):
Good catch.
Reviewed the change and ran a quick local test feeding hex2bin a 32-byte C string containing a hash, with `bin_len` set to `HASH_SIZE`.  Printed each byte written to `bin` and saw the correct hash.

eandersson (approved):

Issue #264, opened 2024-07-30
Handle share reject reasons


mutatrum:
It seems there is some confusion on the rejected shares. One way of making the user more informed would be to gather statistics on the reject reasons. F.e. number of stale shares, etc.

Unfortunately, error response is outside of the stratum api spec AFAIK, so there's no standard for a reject reason field. But with only limited pool implementations, a best effort implementation could capture more cases.

I've seen two formats:

ckpool:
```
{"reject-reason":"Stale","result":false,"error":null,"id":5333}
{"reject-reason":"Above target","result":false,"error":null,"id":8}
```

public pool:
```
{"id":120,"result":null,"error":[21,"Job not found",""]}
{"id":3423,"result":null,"error":[23,"Difficulty too low",""]}
```

Maybe there are other implementations, but these are the pools I've tested with so far.

These are also the different errors I encountered, so a small map with reason and counter would be sufficient.

Ondalf:
https://github.com/tpfuemp/yiimp/blob/dev/stratum/client_core.cpp#L73
Yiimp pools replies as well with
```
{"id":integer,"result":false,"error":[integer,"string",null]}
error field being [error code, "error text", null]
```
So main difference is "result":false ... Don't ask why :D
But when presenting or collecting these errors, there was no documents for stratum-v1 to handle errors, which are caused by asicboost, so pools are reusing existing numbers but with different strings or adding more.
Theoretically, collecting strings would do alone as well just fine.
```
> {"method": "mining.submit", "params": ["donate", "cc", "00000000", "66b3e54a", "7f2f70a0"], "id":4}
< {"id":4,"result":false,"error":[26,"Low difficulty share",null]}
and valid share
> {"method": "mining.submit", "params": ["donate", "cd", "00000000", "66b3e588", "27b36bb0"], "id":4}
< {"id":4,"result":true,"error":null}
```

PR #472, opened 2024-11-11
Parse stratum api reject reason

mutatrum:
Initial implementation of parsing reject reasons as discussed in #264. It can handle two formats, the error array and the reject-reason string. If other flavours are found, it should be fairly trivial to extend the parsing.

This only handles the parsing aspect, feeding back to the front-end will be a separate issue.

I extended an existing and added a unit test, but I don't know how to run them.

### DIFF
--- a/components/stratum/stratum_api.c
+++ b/components/stratum/stratum_api.c
@@ -146,6 +146,7 @@ void STRATUM_V1_parse(StratumApiV1Message * message, const char * stratum_json)
         // parse results
         cJSON * result_json = cJSON_GetObjectItem(json, "result");
         cJSON * error_json = cJSON_GetObjectItem(json, "error");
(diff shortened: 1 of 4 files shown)


components/stratum/stratum_api.c
```diff
@@ -171,6 +181,11 @@ void STRATUM_V1_parse(StratumApiV1Message * message, const char * stratum_json)
                 message->response_success = true;
             } else {
                 message->response_success = false;
+                if (!cJSON_IsNull(reject_reason_json)) {
```
eandersson, line 184: Nit: Technically I think you don't need the IsNull, as both check for a null value and IsString will return false if the type is `cJSON_NULL`.

mutatrum:
image

eandersson (approved):

WantClue (approved):

PR #477, opened 2024-11-13
adjust bm1370 mhz settings to 6.25 multiplier

WantClue:
takes into account the 6.25 multiplier of the step algo for frequency PLL


benjamin-wilson:
The default frequency is 596 on the antminers

zcht:
While I'm looking at it, the 550 MHz step is still missing. The jump from 525 to 575 is extremely large at a gamma.

PR #479, opened 2024-11-15
Mecanix found that we are not updating the difficulty mask properly

eandersson:


eandersson:
From @adammwest 
`There is a consideration here , you don't actually want to set the difficulty mask to the pool diff, for instance OCEAN, VIABTC uses >= 65536. ESP-Miner has many assumptions about winning nonce's being generated frequently by the chip which for some pools would be no longer true, and as a result the statistics are higher variance this is an issue if you get a nonce every x minutes, all data that is collected on a bitaxe would be less valuable as the hash rate sampling frequency reduces. Also there are limits to the maximum ticket mask being set per chip and each pool can set there own initial diff some pools use ~1M (2**20) also some use not power of 2 difficulties you need to convert it to the lower power of 2 first. I am not sure what the ideal behavior is here maybe counting duplicates, maybe counting stale ratio of nonce's with old job ids, measuring the queue length. Which may help identifying the cause. The chip diff should be as low as possible and not dependent on the pool.`

adammwest:
for formatting!
* Pools have huge varying initial difficulties from 1k-> 1M
* Asic hashrate statistics are dependant on high sampling frequency of nonces, which is only the case when chip difficulty is low
* Pool difficulty changes should not change hashrate sampling frequency
* Anything editing the ticket mask must be a power of 2 limited to the correct bounds

PR #480, opened 2024-11-15
Reduce invalid job found log from error to warning

eandersson:
While this message can be useful, most commonly it is seen after receiving a mining.notify, or briefly getting disconnected from the pool. changing it from a ERROR to WARNING log message to not scare users.

This also makes the implementation more consistent, as in some places and chips we already logged as ERROR and in other places as INFO.

### DIFF
--- a/components/asic/bm1366.c
+++ b/components/asic/bm1366.c
@@ -500,7 +500,7 @@ task_result * BM1366_proccess_work(void * pvParameters)
     GlobalState * GLOBAL_STATE = (GlobalState *) pvParameters;
 
     if (GLOBAL_STATE->valid_jobs[job_id] == 0) {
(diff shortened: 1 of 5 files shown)


skot:
a rejected share seems slightly more important than INFO... maybe WARN?

eandersson:
> a rejected share seems slightly more important than INFO... maybe WARN?

Yea - maybe warning. Most important is that we are consistent across all chip types.

Having that said I have only been able to reproduce these type of errors during race conditions with cleanQueue, and cleanQueue is logged as an error already.

skot (changes requested):
Plz change to ESP_LOGW, otherwise LGTM

eandersson:
> Plz change to ESP_LOGW, otherwise LGTM

Updated

WantClue (approved):

eandersson:
Rebased

PR #483, opened 2024-11-16
Improve BM1370 die temp reading accuracy and stability.

skot:
The correct setting for the BM1370 "Analog Mux" register 0x54 has changed from 0x03 to 0x02. This makes temperature reading much more stable. 

Setting the EMC2101 external temp sensor to manual beta compensation (0.11) and changing the ideality from 1.008 to 1.0319 makes the die temperature much more accurate.


main/self_test/self_test.c
```diff
@@ -18,8 +18,12 @@
 #define POWER_CONSUMPTION_TARGET_GAMMA 11       //watts
 #define POWER_CONSUMPTION_MARGIN 3              //+/- watts
 
+#define TEMP_CAL_LIMIT 1200
```
benjamin-wilson, line 21: I think this file needs to be revered from all the changes

main/EMC2101.c
```diff
@@ -86,8 +98,10 @@ float EMC2101_get_external_temp(void)
 
     // Greater than 200C is probably an erroneous reading...
     if (result > 200){
-        return EMC2101_get_internal_temp();
+        ESP_LOGE(TAG, "EMC2101 Invalid result: %04X", reading);
```
benjamin-wilson, line 101: This needs to be removed and replaced by proper sign handling:

```
float EMC2101_get_external_temp(void)
{
    uint8_t temp_msb, temp_lsb;
    uint16_t reading;

    ESP_ERROR_CHECK(i2c_bitaxe_register_read(emc2101_dev_handle, EMC2101_EXTERNAL_TEMP_MSB, &temp_msb, 1));
    ESP_ERROR_CHECK(i2c_bitaxe_register_read(emc2101_dev_handle, EMC2101_EXTERNAL_TEMP_LSB, &temp_lsb, 1));
    
    // Combine MSB and LSB, and then right shift to get 11 bits
    reading = (temp_msb << 8) | temp_lsb;
    reading >>= 5;  // Now, `reading` contains an 11-bit signed value

    // Cast `reading` to a signed 16-bit integer
    int16_t signed_reading = (int16_t)reading;

    // If the 11th bit (sign bit in 11-bit data) is set, extend the sign
    if (signed_reading & 0x0400) {
        signed_reading |= 0xF800;  // Set upper bits to extend the sign
    }

    if (signed_reading == EMC2101_TEMP_FAULT_OPEN_CIRCUIT) {
        ESP_LOGE(TAG, "EMC2101 TEMP_FAULT_OPEN_CIRCUIT: %04X", signed_reading);
    }
    if (signed_reading == EMC2101_TEMP_FAULT_SHORT) {
        ESP_LOGE(TAG, "EMC2101 TEMP_FAULT_SHORT: %04X", signed_reading);
    }

    // Convert the signed reading to temperature in Celsius
    float result = (float)signed_reading / 8.0;

    return result;
}
```

main/EMC2101.c
```diff
@@ -29,6 +29,18 @@ esp_err_t EMC2101_init(bool invertPolarity) {
         ESP_ERROR_CHECK(i2c_bitaxe_register_write_byte(emc2101_dev_handle, EMC2101_FAN_CONFIG, 0b00100011));
     }
 
+    //set Ideality Factor
```
benjamin-wilson, line 32: This should only be for BM1370 and not all chips

main/EMC2101.c
skot: right shifting a signed int should properly sign extend

benjamin-wilson: For whatever reason it does not but I confirmed the above code does.

benjamin-wilson:
Borked this up - see new pull
https://github.com/skot/ESP-Miner/pull/484

PR #491, opened 2024-11-18
Swarm styles, refresh on load, more combined stats, more info in table

mrv777:
## Description

Some more tweaks for the swarm page
- Decrease the white space around the manual swarm add
- When the page is loaded it should refresh the data right away
- Allow the user to set the refresh interval
- Increase the temp threshold for the temp warning color
- Have total hash on its own line and adjust spacing there


mrv777:
Added:
- Total power on swarm page
- Best diff from all devices on swarm page
- Made these stats cards for better styling

mrv777:
Added voltage regulator temp to swarm table too, if available

skot (approved):

skot:
this LGTM! 
image

Issue #440, opened 2024-11-01
AxeOS: Make it easier to restart after saving. especially on mobile


skot:
On Mobile AxeOS right now, to save any changes to NVS you have to;
- click save
- open the hamburger menu
- click restart

Could we just make the save button into save & restart? Or maybe add a restart button right next to every save button?

STSMiner1:
Add "Restart" under the "Save" option to solve this.

mrv777:
Also, could maybe make the restart button here disabled until the save button was pushed?  That way people don't accidentally click restart when they meant to click save first

mutatrum:
Fixed by #493

PR #493, opened 2024-11-18
fix: Add restart to settings & better disabled state

mrv777:
## Description

- Added the restart button to the settings page as requested here: #440 
- Also changed the disabled state so that save is only enabled if there are changes and restart only enabled if changes saved to prevent accidentally clicking that first

### DIFF
--- a/main/http_server/axe-os/src/app/components/edit/edit.component.html
+++ b/main/http_server/axe-os/src/app/components/edit/edit.component.html
@@ -198,9 +198,13 @@
         </div>
 
         <div class="mt-2">
-            <button pButton [disabled]="form.invalid" (click)="updateSystem()"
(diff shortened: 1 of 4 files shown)


skot:
The restart button is enabled after I slick save, but it doesn't actually restart the Bitaxe when I click on it.
image

skot:
I'd like to add this functionality to the network tab too 🙏

mrv777:
Sorry about that, simple fix.  Should work now.  I'll add it to the network tab later too

mrv777:
PR updated @skot if you want to test when you get a chance.  I tested the settings page, but haven't had a chance yet to upload the changes to a device again to test the network page changes

skot:
both settings and network work great! I think this is a huge usability improvement, very nice @mrv777

skot (approved):

PR #495, opened 2024-11-19
Format, Add a warm up time, increase supra power target

benjamin-wilson:


WantClue (approved):
LGTM

skot:
@benjamin-wilson can you resolve these conflicts? It looks bad, but I think it's just formatting

PR #497, opened 2024-11-19
Refresh list on F5 (page reload)

eandersson:


eandersson:
Already solved in https://github.com/skot/ESP-Miner/pull/491/files

PR #502, opened 2024-11-20
add 205 config and remove self test flag 

WantClue:
remove self test flag as we already have it remove for all other models

add 205 config


benjamin-wilson:
The units should always test on a factory flash

benjamin-wilson:
Please allow discussion for large changes

skot:
@eandersson is this going to break how the CI scripts build factory images?

eandersson:
> @eandersson is this going to break how the CI scripts build factory images?

Yea - we use this for factory, but we could use a different approach for that if needed, but gets tricky if there is a firmware that does not support the test flag.

skot:
Ok, we definitely don't want to remove selftest support from the esp-miner

eandersson:
> Ok, we definitely don't want to remove selftest support from the esp-miner

Let me know if you want me to implement it in the CI, instead of reverting this PR. I would need to know though if there are any models that needs to be excluded from the self test.

Issue #518, opened 2024-11-26
Browsing for update file disallowed after unsuccessful initial attempt


tdb3:
Settings page, `Update Firmware` and `Update Website` areas.

If the user mistakenly loads an incorrectly named update file (e.g. `dummy.bin` instead of `www.bin` or `esp-miner.bin`), the user receives an error toast (e.g. `Incorrect file, looking for www.bin`). Afterward, clicking the `Browse` button repeats the same error. The user is not permitted to browse for a new file (to remedy the mistake). This was observed on both Firefox and Chrome.

The current workaround is to refresh the `Settings` page, but the desired behavior would be permitting the user to browse for a new file without having to refresh the page.

![image](https://github.com/user-attachments/assets/22a8e906-849c-4baf-a0b5-443d3dc1fc82)

tdb3:
#521 fixed this issue nicely

PR #521, opened 2024-11-27
fix: update feedback & reset file input

mrv777:
## Description

- This should fix the errors so that file upload events are properly passed and the function doesn't error out, resulting in accurate feedback and a refresh on www upload success
- The also resets the file inputs on upload which should resolve issue #518


skot (approved):

tdb3:
Post merge test results:
- Attempting to use a dummy.bin file for either firmware or website resulted in an error toast (expected)
- 2nd attempt to browse now works well (able to select file)
- Updated firmware with `esp-miner.bin`. Observed the percentage completion (nice!). Received the success toast.
- Updated website with `www.bin`. Observed the percentage completion. Received the success toast and the page refreshed

Great work!

Issue #511, opened 2024-11-23
Master Branch - PR's / Updates that have been pushed / merged around 11/21/2022 to the Master branch are causing an issue with 205 board.


STSMiner1:
Release firmware v2.4.0 - 205 works fine.

Firmware built 11/21/22 onwards with recent updates / PR's that have been merged to the master branch are causing an issue with the 205, issue is not seen on Supra (401) or Gamma (601).

Logs enclosed (startup) for both firmware builds from the 205 board.

Releasefirmware-v2.4.0.txt

Firmware built 11/21/22 with latest updates
PreRelease-11.21.2022.txt  (v2.4.0-7-g4100402-dirty)

Screenshot of issue also enclosed.

[Releasefirmware-v2.4.0.txt](https://github.com/user-attachments/files/17878146/Releasefirmware-v2.4.0.txt)

[PreRelease-11.21.2022.txt](https://github.com/user-attachments/files/17878147/PreRelease-11.21.2022.txt)

![205-masterbranch-11-21-2022-2](https://github.com/user-attachments/assets/7a98cdcf-0b6b-4031-a0ce-b02b5dfa0a6c)

STSMiner1:
Typo's in report (year), sorry about that.

STSMiner1:
Pulled the repo down fresh and re-built the firmware to rule that out, flashed both files to the Ultra 205, no change.

![205 v2 4 0-12-g29a543d-dirty](https://github.com/user-attachments/assets/81d44a03-4c94-4a2f-aa29-cd284e15c26c)

eandersson:
What happens if you change the frequency?

STSMiner1:
Nothing changes with how it performs, voltage readings and temps show wrong along with the temp.

ASIC Frequency updates on the Dashboard as does the Measured ASIC Voltage when changed in the Settings page.

eandersson:
Can you try to comment out this line to see if it makes a difference? might be some weird race condition
https://github.com/skot/ESP-Miner/blob/master/components/asic/serial.c#L51

mutatrum:
Can you zoom in on the actual commit by doing a `git bisect`?

skot:
Narrowing it down: problem does not happen in 17873069a2a5237f4e8910e3fd500018d9ea25c9 and happens in 41004029901a0127b626c6070495f9e0af7046d0

skot:
[...]

PR #530, opened 2024-11-29
Fix serial issues on 204 and 205 boards (bm1366)

eandersson:
This moves the ux wait to the asic level. Fixing #511 .


WantClue (approved):
LGTM

skot:
I _think_ this is fixed in #532

PR #532, opened 2024-11-30
Fix hashing on BM1366

skot:
The register setting to change the baudrate on the ASIC and the the ESP32 baudrate change were not happening at the same time. I moved the ASIC baudrate register change to happen after the init function is completed (like we do on BM1368 and BM1370)


skot:
Seems to be (still) working on 401 and 601

WantClue (approved):
Working, this reminds me of reworking the ramp up function...

PR #543, opened 2024-12-02
Keep old pool config synchronized until reboot

### LINKED ISSUE #537: Inconsistent pool config without restart
Url/Port from global state remain old after saving but user/pass are updated from nvs config.

The config should either work without reboot or the old data should be retained?

terratec:
Resolves #537


eandersson (approved):
LGTM (not tested).

WantClue (approved):
Looks good

PR #558, opened 2024-12-06
Add support for light mode to AxeOS

bitdaw:
I just got my Bitaxe yesterday and I noticed there doesn't seem to be any option for light mode. I added support for it with this PR based on the preference set in the user's OS (Windows, iOS etc.)

If I have missed some rules for contributing to the project then sorry for that, like I say I only just found out about this project. 

Happy to make any suggested changes based on any feedback given :)


bitdaw:
Looked deeper into the code and I see there is some functionality in the layout service for theme switching but it doesn't seem to be used, maybe there were some plans for this in the future? But it looks like my PR doesn't fit with that structure. I'm happy to work on a solution that fits with the layout service. @benjamin-wilson maybe you can give some info on that? I see your name in this service a lot 😅

WantClue:
Would be nice to have a toggle functionallity, or even add a whole palette of theme's instead of just looking for windows,mac,etc. system preferences

PR #560, opened 2024-12-06
fix: Swarm hashRate check, shorter uptime, & pause refresh on scan

mrv777:
## Description

- On a swarm scan, only add devices that have the hashRate property to prevent devices like the Philips Hue hub from trying to get added that has the same endpoint
- Allow shorter date ago pipe and use it on the swarm page so the columns don't get too long
- Prevent swarm refreshes during scans


WantClue (approved):
LGTM

eandersson (approved):
Tested on one Supra. LGTM.

PR #576, opened 2024-12-12
fix: Show warning if freq is low or not set

mrv777:
## Description

- Show warning if frequency is null or low as mentioned here: https://github.com/skot/ESP-Miner/issues/584


alltheseas:
@mrv777 I see you added a null/low frequency warning - thank you.

One thing to consider from the perspective of a non-technical Bitaxe newbie - the newbie does not know what a low value is. Consider adding some context (for instance, as a tooltip to frequency field)- e.g. "Range of X - Y is typical value. Values below Z are considered low. " 

I can create a new issue for this if helpful.

WantClue (approved):

PR #578, opened 2024-12-12
fix: Show custom values in dropdown if set

mrv777:


WantClue (approved):

WantClue (changes requested):
add suggested code from dc and don't miss on the non-arrow binding for the checkDevTools

WantClue (approved):

eandersson (approved):

PR #586, opened 2024-12-17
No addresses in config files

eyelight:
This PR 
- propagates the changes from #438 by updating the config cvs files with the same prompt
- adds the new 'stratum fallback' related keys to the example config file, bringing the example config into conformity with the rest
- removes a btc address from axeos graceful fallback return in `SystemService::getInfo` method, which could cause surprise/confusion if displayed

If merged, I believe the repo will be purged of all btc addresses.


main/http_server/axe-os/src/app/services/system.service.ts
```diff
@@ -47,8 +47,8 @@ export class SystemService {
           stratumPort: 21496,
           fallbackStratumURL: "test.public-pool.io",
           fallbackStratumPort: 21497,
-          stratumUser: "bc1q99n3pu025yyu0jlywpmwzalyhm36tg5u37w20d.bitaxe-U1",
```
eandersson, line 50: This is mock data for testing and development. I think we should keep a valid btc address here.

eyelight: Yes, it seemed like it was mock data. I guess the question is, given bitaxe's growth in popularity and the corresponding expectation that most new eyes on the project will trend toward less experienced, does the convenience of hardcoding a valid address outweigh the downstream public relations effort needed to squelch misinformed accusations of foul play on twitter, etc. I thought it would be cool to make the repo so far above reproach that there isn't even a single valid address present at all, which makes a good proactive talking point for skot when discussing the project publicly, etc.

WantClue:
For now we will keep them in there for easy testing and debug purposes

PR #600, opened 2024-12-27
UI themes

### LINKED ISSUE #443: Theme - Add Theme options for those that are color blind etc
Theme 
https://github.com/skot/ESP-Miner/pull/429

Add Theme options for those that are color blind or want different colors for text / graph.

WantClue:
Adding Color Themes Dark - Light
Adding Multiple Accent color styles

Closes #443


ChimpRocket:
@WantClue I love the new themes.  1 colour is missing though... BitCoin orange :-) . Then it would match 1 of my Bitaxe stands and Halo colour

PR #602, opened 2025-01-02
Screen carousel interaction

WantClue:
Gives the user to enable/disable the screen carousel


mutatrum:
I actually thought this was a toggle to disable the scroll animation and instantly switch screens.

An alternative option for this would be to do this with a 'double click', e.g. two single clicks in short succession to pause or unpause the carousel. That would save another configuration option. It does overload the button quite a bit.

mutatrum:
With #1022 we upgraded to LVGL 9.3, which adds double- and triple click handlers.

WantClue:
Unnecessary

PR #603, opened 2025-01-03
feature: add expected hashrate to api

### LINKED ISSUE #508: feature: Add expected hashrate to API
It would be nice to have the expected hashrate shown in the API directly to make it a bit easier for external tools to parse it.  AFAIU there is some calculation being done involving cores, frequency, etc, to find this value.

I am happy to PR this if someone can start me in the right direction as to where to source this.

jpcomps:
Fixes #508 

Adds expected HR in GH/s based on chip count, freq and small_core_count. There is a divide which could potentially have a small performance impact, but should be minimal/negligible as is only calculated when the endpoint is called.

### DIFF
--- a/main/http_server/http_server.c
+++ b/main/http_server/http_server.c
@@ -542,6 +542,8 @@ static esp_err_t GET_system_info(httpd_req_t * req)
     char * stratumUser = nvs_config_get_string(NVS_CONFIG_STRATUM_USER, CONFIG_STRATUM_USER);
     char * fallbackStratumUser = nvs_config_get_string(NVS_CONFIG_FALLBACK_STRATUM_USER, CONFIG_FALLBACK_STRATUM_USER);
     char * board_version = nvs_config_get_string(NVS_CONFIG_BOARD_VERSION, "unknown");
+    uint16_t freq = nvs_config_get_u16(NVS_CONFIG_ASIC_FREQ, CONFIG_ASIC_FREQUENCY);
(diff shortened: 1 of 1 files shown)


main/http_server/http_server.c
```diff
@@ -418,6 +420,8 @@ static esp_err_t GET_system_info(httpd_req_t * req)
             small_core_count = -1;
             break;
     }
+    float expected_hashrate = freq * ((small_core_count * asic_count) / 1000.0);
+    cJSON_AddNumberToObject(root, "expectedHashrate", expected_hashrate);
```
b-rowan, line 424: Any opposition to changing this to expectedHashrateMhs to avoid the / 1000?  I think unit should be included in the naming anyway, either that or in raw hashes/s?

remcoros: please keep it a number (not a string), so we can use it in prometheus without hacks.

WantClue:
@jpcomps resolve conclict please

jpcomps:
@WantClue should be resolved and rebased again the dev-latest branch.

WantClue:
for 2.8.0 this is planned, be aware there will also be some more API changes with 2.8.0 so this will need some work after we're done with the init changes

jpcomps:
No problem let me know and will update according to the changes

WantClue:
@jpcomps can you check on this PR please? We had some issues lately with the dev-latest branch which might caused trouble with your PR

jpcomps:
@WantClue should be good. Thanks for the headsup

WantClue (approved):

PR #606, opened 2025-01-04
Build images, flash with webflasher locally

dustinb:
Doing as a draft as I'm not sure if this is a good idea :).  My thoughts are if someone wants to hack on axeos or the firmware but not deal with esp idf and python dependencies but have docker this will build images and spin up the webflasher locally.

1. Clone ESP-miner
2. Do your stuff
3. sh build-image.sh
4. Flash your dev factory images with web flasher


skot:
This should great to me -- could I persuade you to also include setup instructions?

build-docker/Dockerfile-factory
```diff
@@ -0,0 +1,15 @@
+FROM espressif/idf:v5.3.1
```
eandersson, line 1: We probably want to use arguments for dependency versioning. Also, we are currently using idf v5.3.2 and moving to 5.4.0 soon.

dustinb:
Moved to web flasher

PR #617, opened 2025-01-05
Press reset then hold boot for self test

benjamin-wilson:
Requested from reseller

### DIFF
--- a/main/main.c
+++ b/main/main.c
@@ -56,8 +56,10 @@ void app_main(void)
         return;
     }
 
+    // Optionally hold the boot button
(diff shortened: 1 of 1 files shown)


skot:
This sounds like it is going to interfere with getting into the ESP32 bootloader.

benjamin-wilson:
It does not if you press reset then hold boot

skot:
Isn't that how you get into the ESP32 Bootloader?

benjamin-wilson:
I think the boot button needs to be held down while reset is pushed and not after.

skot:
BOOT just needs to be held at power up. (you can do it when plugging in power, or resetting with the button)

Issue #523, opened 2024-11-27
Grandma doesn't know what SSID means


skot:
When first plugging in a new Bitaxe, Grandma gets this screen:
<img width="871" alt="image" src="https://github.com/user-attachments/assets/170505ee-6b01-40a9-8d99-ba79b88aa30c">

Problem is, Grandma doesn't know what a SSID is, or what she's supposed to do here. I think we need more helpful wording here to guide users to connect to the setup wifi..

Rushikeshmagdum0379:
Can we get a video over to this from start to end. To solve this issue.

STSMiner1:
https://github.com/skot/ESP-Miner/issues/447#issuecomment-2453547568

mrv777:
I think using `Wi-Fi` instead of `SSID` might make more sense to most people

PR #623, opened 2025-01-06
Grandma knows what Wifi means!

b-rowan:
Use Wifi instead of SSID on screen, and add short `(join to setup)` message instead of `Configuration SSID`. This should be easier to understand, but does add 2 characters, which will fill the screen completely.

Fixes: #523

### DIFF
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -224,27 +224,41 @@ void wifi_init(const char * wifi_ssid, const char * wifi_pass, const char * host
     ESP_LOGI(TAG, "ESP_WIFI Access Point On");
     wifi_init_softap();
 
-    /* Initialize STA */
-    ESP_LOGI(TAG, "ESP_WIFI_MODE_STA");
-    esp_netif_t * esp_netif_sta = wifi_init_sta(wifi_ssid, wifi_pass);
 
-    /* Start WiFi */
-    ESP_ERROR_CHECK(esp_wifi_start());
+    /* Skip connection if SSID is null */
+    if (strlen(wifi_ssid) == 0) {
+        ESP_LOGI(TAG, "No WiFi SSID provided, skipping connection");
 
-    /* Disable power savings for best performance */
-    ESP_ERROR_CHECK(esp_wifi_set_ps(WIFI_PS_NONE));
+        /* Start WiFi */
+        ESP_ERROR_CHECK(esp_wifi_start());
 
-    /* Set Hostname */
-    esp_err_t err = esp_netif_set_hostname(esp_netif_sta, hostname);
-    if (err != ERR_OK) {
-        ESP_LOGW(TAG, "esp_netif_set_hostname failed: %s", esp_err_to_name(err));
+        /* Disable power savings for best performance */
+        ESP_ERROR_CHECK(esp_wifi_set_ps(WIFI_PS_NONE));
+
+        return;
     } else {
-        ESP_LOGI(TAG, "ESP_WIFI setting hostname to: %s", hostname);
+        /* Initialize STA */
+        ESP_LOGI(TAG, "ESP_WIFI_MODE_STA");
+        esp_netif_t * esp_netif_sta = wifi_init_sta(wifi_ssid, wifi_pass);
+
+        /* Start WiFi */
+        ESP_ERROR_CHECK(esp_wifi_start());
+
+        /* Disable power savings for best performance */
+        ESP_ERROR_CHECK(esp_wifi_set_ps(WIFI_PS_NONE));
+
+        /* Set Hostname */
+        esp_err_t err = esp_netif_set_hostname(esp_netif_sta, hostname);
+        if (err != ERR_OK) {
+            ESP_LOGW(TAG, "esp_netif_set_hostname failed: %s", esp_err_to_name(err));
+        } else {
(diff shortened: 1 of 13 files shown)


skot:
I think we might have already changed this? @mutatrum ?

b-rowan:
> I think we might have already changed this? @mutatrum ?

Figured I would give that issue a shot, just did a CTRL+F for `Configuration SSID` and changed the couple instances that came up.  Could be the wrong ones, my C isn't particularly strong, but for what its worth I did replace 3 instances.

skot:
> You want me to change this in this PR? I have some ideas, maybe `Welcome to your new Bitaxe! Set it up by connecting to the setup Wifi.?

I think this would be great! The one issue we have is determining if the Bitaxe is being setup for the first time, or it just can't find the WiFi it has been configured for..

b-rowan:
> > You want me to change this in this PR? I have some ideas, maybe `Welcome to your new Bitaxe! Set it up by connecting to the setup Wifi.?
> 
> I think this would be great! The one issue we have is determining if the Bitaxe is being setup for the first time, or it just can't find the WiFi it has been configured for..

My thought here is have Wifi SSID initially set to `null`, then check that.  Any time it is set by a user, can show a "failed to connect" message, otherwise it can show a setup message.  Can also have it so that if an empty string is input into the SSID it can reset to `null` and show the initial setup message.

Again, my C is very very limited, so I might need some pointers on where to start with this (at least where I would put these checks), and I can do some work to have it put in here.

skot:
> > > You want me to change this in this PR? I have some ideas, maybe `Welcome to your new Bitaxe! Set it up by connecting to the setup Wifi.?
> > 
> > 
> > I think this would be great! The one issue we have is determining if the Bitaxe is being setup for the first time, or it just can't find the WiFi it has been configured for..
> 
> My thought here is have Wifi SSID initially set to `null`, then check that. Any time it is set by a user, can show a "failed to connect" message, otherwise it can show a setup message. Can also have it so that if an empty string is input into the SSID it can reset to `null` and show the initial setup message.
> 
> Again, my C is very very limited, so I might need some pointers on where to start with this (at least where I would put these checks), and I can do some work to have it put in here.

I think that would work as long as it is the null value, and not the "null" string (someone is definitely going to name their wifi that). Are you on the OSMU Discord? (discord.gg/osmu) the firmware channel is a good place to discuss this / get help.

components/connect/connect.c
```diff
@@ -224,27 +224,41 @@ void wifi_init(const char * wifi_ssid, const char * wifi_pass, const char * host
...
 
-    /* Initialize STA */
-    ESP_LOGI(TAG, "ESP_WIFI_MODE_STA");
-    esp_netif_t * esp_netif_sta = wifi_init_sta(wifi_ssid, wifi_pass);
 
-    /* Start WiFi */
-    ESP_ERROR_CHECK(esp_wifi_start());
+    /* Skip connection if SSID is null */
+    if (strlen(wifi_ssid) == 0) {
+        ESP_LOGI(TAG, "No WiFi SSID provided, skipping connection");
```
mutatrum, line 230: Nitpick, also further down: the official spelling is `Wi-Fi`.

components/connect/connect.c
```diff
@@ -224,27 +224,41 @@ void wifi_init(const char * wifi_ssid, const char * wifi_pass, const char * host
...
+
+        /* Disable power savings for best performance */
+        ESP_ERROR_CHECK(esp_wifi_set_ps(WIFI_PS_NONE));
+
+        /* Set Hostname */
+        esp_err_t err = esp_netif_set_hostname(esp_netif_sta, hostname);
+        if (err != ERR_OK) {
+            ESP_LOGW(TAG, "esp_netif_set_hostname failed: %s", esp_err_to_name(err));
+        } else {
+            ESP_LOGI(TAG, "ESP_WIFI setting hostname to: %s", hostname);
```
mutatrum, line 255: The TAG is already `wifi_station`, which is not in line with other files. Maybe this file should be wifi.c, and the logging TAG `wifi`?

mutatrum (approved):
Code LGTM, added 2 nitpick comments. Didn't test as I don't have access to my Bitaxe at the moment.

components/connect/connect.c
mutatrum: Clarifications: most if not all other files have a TAG of the source file without extension (e.g. `display.c` has tag `display`, etc.)

b-rowan: No opposition to this, not entirely sure if this is within PR scope but I have no issue changing it.

b-rowan:
I still have some stuff I want to add here, mainly I would like to change the message shown on the display depending on whether the WiFi is not set up, or if it is failing to connect.

b-rowan:
Ready for review 👍

components/connect/connect.c
b-rowan: Quick CTRL+F through the project shows a bunch of instances of both.  I have changed all instances in `connect.c` for this PR, but maybe a separate PR is in order to standardize this?  Want to make sure it doesn't break anything...

WantClue (approved):
LGTM

STSMiner1:
This needs more info on the screen for "Grandma" !

image

eandersson (approved):

PR #624, opened 2025-01-07
Init AxeOS AP mode

dustinb:
Referring to these issues 

https://github.com/skot/ESP-Miner/issues/568
https://github.com/skot/ESP-Miner/issues/385

A way to do ap mode client side.  Only showing the network component when in ap mode

image

Thinking the AxeOS logo should be there and maybe some instructions like "Your Bitaxe is in Wi-fi setup mode. Enter your Wi-fi network name and password and restart for the full AxeOS experience".   A "Save & Restart" button might be nice too.

### DIFF
--- a/main/http_server/axe-os/src/app/app-routing.module.ts
+++ b/main/http_server/axe-os/src/app/app-routing.module.ts
@@ -13,8 +13,14 @@ import { ApModeGuard } from './guards/ap-mode.guard';
 
 const routes: Routes = [
   {
-    path: 'ap',
-    component: NetworkComponent
+      path: 'ap',
+      component: AppLayoutComponent,
+      children: [
+        {
+          path: '',
+          component: NetworkComponent
+        }
+      ]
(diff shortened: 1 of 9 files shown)


STSMiner1:
Tested - works - loads the Network config page (first time setup) when connected. 

Notes 
1/ Save and Restart button needs attn, not easy to see that these are clickable options.
2/ In some cases the end user will need to use the default IP address to access AxeOS (192.168.4.1) - defaults to the Network page. 
3/ If the Bitaxe can not find WiFi SSID (ie, you changed your router for a new one) it will default to "AP Mode" so you can set it up again.

image
image

dustinb:
There must have been change to how styles are loaded.  I've updated the PR to use app layout but without the side bar so the logo is also shown and button styles work.   It works but maybe better to use an additional layout?


IMG_3663

WantClue (approved):
lgtm

PR #626, opened 2025-01-07
Is SPIRAM Available?

eandersson:


WantClue:
I think with that approach we can use this for multiple checks.

WantClue (approved):
LGTM

PR #634, opened 2025-01-09
Enable Unit Test CI

eandersson:
Enable automatic unit testing as part of the CI process. This runs an emulated version using QEMU which means that not all tests are possible to run e.g. no access to the ASIC, mbedtls runs in software mode.


test-ci/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+#
+set(TEST_COMPONENTS "bm1397 stratum" CACHE STRING "List of components to test")
+
+include($ENV{IDF_PATH}/tools/cmake/project.cmake)
+
+set(IDF_TARGET "esp32s3")
+
+idf_build_set_property(COMPILE_DEFINITIONS "-DCONFIG_ASIC_FREQUENCY=100" APPEND)
+
+project(unit_test_stratum)
```
tdb3, line 21: This seems to be really close to `test/CMakeLists.txt`. To prevent forgetting to include future changes to `test/CMakeLists.txt` to `test-ci/CMakeLists.txt` maybe we can use CMake's `include`?

For example, if someone adds a new component to `TEST_COMPONENTS` in test/CMakeLists.txt but forgets to add it to test-ci/CMakeListst.txt, then the qemu focused unit test build might run into issues or may omit some tests?

tdb3 (approved):
Great work! Adding unit tests to CI makes a lot of sense, even if we can't run all unit tests due to host limitations.

test-ci/CMakeLists.txt
eandersson: I added a link for now.

johnny9:
Great idea to integrate QEMU into the github action. Very pragmatic solution to getting unit tests finally running in CI

eandersson:
Rebased

WantClue (approved):
LGTM

PR #641, opened 2025-01-10
Add instructions for development

pRizz:
This is useful for new folks wanting to develop and build their own custom bitaxe software.


skot:
this is great, thanks for this @pRizz ! LGTM

skot:
> This is useful for new folks wanting to develop and build their own custom bitaxe software.

I think there might be a step missing for installing the proper version of node.js?

pRizz:
Ah good point! I added a line for nodejs. I did not specify what version since I am not sure which is the oldest version that works. I have tested that v22 works.

skot:
> Ah good point! I added a line for nodejs. I did not specify what version since I am not sure which is the oldest version that works. I have tested that v22 works.

thanks!

Issue #633, opened 2025-01-09
[FEATURE REQUEST] Add "avg hash rate" and "avg Efficiency"


ovik88:
Hey, would it be possible to add Avg values for hash rate and efficiency?
<img width="308" alt="image" src="https://github.com/user-attachments/assets/d778feec-6530-4e76-8ed7-a5b0e4df8138" />

skot:
We could do this now with just a running average over the current session (ie how long the browser window has been open)

There are future plans to store some historical performance data onboard, which would make this much more useful.

mrv777:
If we add avg hash to the card, I'm wondering if it should be removed from the chart to clean the chart up.  Maybe people like seeing the history of the avg hash there, but not sure if its needed

skot:
> If we add avg hash to the card, I'm wondering if it should be removed from the chart to clean the chart up. Maybe people like seeing the history of the avg hash there, but not sure if its needed

I think that makes sense.. I find myself wanting to know the average hashrate, and sometimes it's tough to read off the chart.

PR #643, opened 2025-01-11
fix: Move avg hash to card & add eff avg

mrv777:
## Description

- Move the average hashrate from the chart to the card
- Add average efficiency 
  - Needed to add array to record power since it can change and is needed for this
- Removed unused variable & don't let temperatureData go over 720 data points too
image

Issue reference: #633


mrv777:
Only thing I don't like is how long the text gets with expected on that same line.  I didn't want to do a new line though as there is that imbalance in my eyes unless we had more data to the last 2 cards too so they all have 4 lines

mrv777:
I think this may be better
image

mutatrum:
Maybe add expected on the bottom line, next to average, right aligned in the card?

mrv777:
> Maybe add expected on the bottom line, next to average, right aligned in the card?

Have to figure out if that looks okay on smaller screens too

mutatrum:
If it gets smaller, all values of a card should be on their own line, I would think?

mrv777:
Maybe this is better?
image

mrv777:
or this
image

eandersson (approved):

WantClue (approved):

PR #646, opened 2025-01-11
fix: Show IP of bitaxe you are restarting on swarm

mrv777:
## Description

- When using the restart button on swarm, show the IP address of the successful or failed restart


skot:
maybe we should give this treatment to all the toasts releated to things happening on a "remote" bitaxe? I'm thinking about the ones when you're in the edit window in swarm.

STSMiner1:
Tested - All good (sub-menu) for each device.

WantClue (approved):

PR #647, opened 2025-01-11
fix: Color swarm ips by model or count

mrv777:
## Description

- Color the IPs in swarm by the ASIC modal or if the count > 1 to easier ID what each is


mrv777:
image

STSMiner1:
I like this, maybe this can be added in the next release ?

WantClue (approved):

PR #653, opened 2025-01-15
fix: selftest button functionality

benjamin-wilson:


skot:
this still isn't the functionality I originally made;

- Tests fail, you can long-press BOOT to switch off selftest mode and restart, or you can press reset to run the self tests again.
- Tests pass, you can just press reset to switch off selftest mode and restart

skot:
I think the NVS write on BOOT long press in reset_self_test() isn't actually finishing before the esp_restart() happens

skot:
Seems like `nvs_config_set_u16(NVS_CONFIG_SELF_TEST, 0);` wasn't working when being called from the lvgl button callback function context. I added a semaphore to do this and it seems to be working now.

PR #654, opened 2025-01-16
fix: add mock data and cache to theme service

mrv777:
## Description

- The current theme service doesn't work well in localhost testing since it can't pull the info from the device, so this adds mock data
- getThemeSettings is called a few times on load so adding a cache should just do one API call instead of the 5+ we currently do.  Just did 1 second to start to keep the data fresh while preventing all these same calls in the same second

### DIFF
--- a/main/http_server/axe-os/src/app/services/theme.service.ts
+++ b/main/http_server/axe-os/src/app/services/theme.service.ts
@@ -3,7 +3,7 @@ import { HttpClient } from '@angular/common/http';
 import { Observable } from 'rxjs';
 import { environment } from '../../environments/environment';
 import { of } from 'rxjs';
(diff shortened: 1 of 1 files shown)


WantClue (approved):
LGTM, but do we actually need to load in mock data?

mrv777:
> LGTM, but do we actually need to load in mock data?

I could be doing it wrong, but without the mock data when I try to just run the UI locally for testing no theme gets loaded so a lot more is black then there should be and doesn't look right

eandersson:
> > LGTM, but do we actually need to load in mock data?
> 
> I could be doing it wrong, but without the mock data when I try to just run the UI locally for testing no theme gets loaded so a lot more is black then there should be and doesn't look right

I haven't tried it in a while with themes, but I believe we do need it for local testing.

PR #656, opened 2025-01-16
Optimize compiler for performance

eandersson:
Enables some compile level optimization focusing on performance and fixing some compiler errors triggered by the optimizations.

e.g.
> /app/main/TPS546.c:576:12: error: 'value' may be used uninitialized [-Werror=maybe-uninitialized]


mutatrum:
Maybe take a look at #427. It's an outdated PR but it does change some things where the re-used values are refactored out. Might be a cleaner solution.

PR #659, opened 2025-01-16
 allow-cors-in-ap-mode

benjamin-wilson:


skot:
this fixes the captive portal domain filtering issue on iOS

PR #665, opened 2025-01-18
Don't allow 0.0.0.0 in CORS

eandersson:
This is mostly just weird behavior, as in theory we are allowing `0.0.0.0`, as we used `0` to detect a CORS failure making it fallback on the request ip address. This reorganizes the code a bit so that if CORS is set to `http://0.0.0.0` we evaluate it properly as `0`.


WantClue (commented):
LGTM

eandersson:
Rebased

PR #667, opened 2025-01-18
api formatting consistency

### LINKED ISSUE #621: bug: API formatting inconsistency

0xf0xx0:
`/api/system/info` keys are all camelCased now  

fixes #621


WantClue (changes requested):
too much other unrelated changes in formatting,
also hashrate has been changed to all lower case, why?

0xf0xx0:
> too much other unrelated changes in formatting,
they were also not consistently formatted, and if we're gonna fix the formatting of one key we might as well fix the rest, instead of waiting for issues to trickle in uwu  
i can reduce it to just `overheat_Mode` -> `overheatMode`.

> also hashrate has been changed to all lower case, why?
following microsoft conventions, closed compound words don't have each word capitalized: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalizing-compound-words-and-common-terms

0xf0xx0:
yeah, i've been busy and haven't gotten around to updating the webui. i'll add it soon uwu

dustinb:
Consider that AxeOS is not the only consumer of this API.   There are things like benchmark tools, metrics collectors and home automation integrations.   Most of these fields probably don't matter but hashRate would be one that other tools definitely use.

This would be a breaking change for an API and usually done by creating a new version like /api/system/v2/info. Or sending and allowing a patch with both spellings.    Then deprecate the v1 api or the old spellings in a future release.   Could cause problems if someone updates their firmware but doesn't update www.bin

0xf0xx0:
there's been a little bit of discussion on the discord, but the focus has been on more important prs.

0xf0xx0:
ok, so it seems the main nit is to add versioning. should we start at /v2/, leaving the current api as an implicit /v1/?

skot:
> i think starting a v2 seems reasonable. it'd be helpful to start with an openapi doc as well to get us all on the same page as well as facilitating support for other tooling that knows how to work with openapi documents.

A solid API doc would be _amazing_

0xf0xx0:
alright, here's all the formatting changes that should be made to keep it all camelcased, as well as nudging the endpoints down to `/api/v2/`. also, convention is for endpoints to be all-lowercase. is there a way to have a case-insensitive handler? a quick skim of the [httpd_uri](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/protocols/esp_http_server.html#_CPPv49httpd_uri) docs doesn't show anything...

WantClue (approved):

0xf0xx0:
id like to also update https://github.com/bitaxeorg/ESP-Miner/blob/5b5b46255616b0efec9d7ce3990035b92e1ef4a7/main/http_server/openapi.yaml#L489 but this seems like itd involve a frontend refactor, as its used directly here https://github.com/bitaxeorg/ESP-Miner/blob/5b5b46255616b0efec9d7ce3990035b92e1ef4a7/main/http_server/axe-os/src/app/services/system.service.ts#L130-L134
i can create a small patch if thatd be wanted until the frontend can be refactored to use `asicModel`?

0xf0xx0:
depends on #945

0xf0xx0:
initial suggestions from 945 implemented, except metrics, need to look into that  
other than that its testable  
U: frontend will be done in 672 once the v2 api is finalized

0xf0xx0:
just keepin it up to date, we should look into an api freeze after #1271 and #1202 and get it cleaned up.

changes since last rebase:
- the `hashrateRegister` and `errorRegister` keys are removed, theyre just duplicates of `hashrateMonitor.hashrate` and `hashrateMonitor.errorCount`
- `hashrateMonitor.asics[].total` -> `hashrateMonitor.asics[].hashrate`
- `hashrateMonitor.hashrate` -> `hashrateMonitor.totalHashrate`
- `hashrateMonitor.errorCount` -> `hashrateMonitor.totalErrorCount`
- `scriptsig` -> `scriptSig`

0xf0xx0:
nudged `uptimeSeconds`, `ipv4`, `ipv6`, and the free mem keys to `/system/board`

0xf0xx0:
prs merged, lets tackle the api :3

main/http_server/axe-os/api/system/asic_settings.c
```diff
@@ -35,14 +38,42 @@ esp_err_t GET_system_asic(httpd_req_t *req)
...
+    cJSON_AddNumberToObject(root, "smallCoreCount", GLOBAL_STATE->DEVICE_CONFIG.family.asic.small_core_count);
+    cJSON_AddStringToObject(root, "swarmColor", GLOBAL_STATE->DEVICE_CONFIG.family.swarm_color);
+    cJSON_AddStringToObject(root, "deviceModel", GLOBAL_STATE->DEVICE_CONFIG.family.name);
+    cJSON_AddStringToObject(root, "boardVersion", GLOBAL_STATE->DEVICE_CONFIG.board_version);
+    cJSON_AddBoolToObject(root, "isPSRAMAvailable", GLOBAL_STATE->psram_is_available);
+    cJSON_AddNumberToObject(root, "maxPower", GLOBAL_STATE->DEVICE_CONFIG.family.max_power);
+    cJSON_AddNumberToObject(root, "nominalVoltage", GLOBAL_STATE->DEVICE_CONFIG.family.nominal_voltage);
+    cJSON_AddBoolToObject(root, "apEnabled", GLOBAL_STATE->SYSTEM_MODULE.ap_enabled);
+    cJSON_AddStringToObject(root, "macAddr", formattedMac);
+    cJSON_AddStringToObject(root, "runningPartition", esp_ota_get_running_partition()->label);
```
WantClue: The asic endpoint should be specific to asic parts. I don't think a switch to a /board endpoint is useful here in this kind of combination. A better approach would be an /asic /info /board /system

0xf0xx0:
closing, will open a new pr after further api discussion

PR #669, opened 2025-01-20
Experimenting with mDNS and CORS

eandersson:
Not sure how CORS would be implemented here, but this seems to work at least. :D

### DIFF
--- a/main/http_server/http_server.c
+++ b/main/http_server/http_server.c
@@ -83,18 +83,10 @@ static esp_err_t ip_in_private_range(uint32_t address) {
     return ESP_FAIL;
 }
 
-static uint32_t extract_origin_ip_addr(httpd_req_t *req)
+static uint32_t extract_origin_ip_addr(char *origin)
 {
-    char origin[128];
     char ip_str[16];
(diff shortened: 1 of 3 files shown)


eandersson:
> I can't get changes to work with this when using "hostname", not sure if that was the intent here with mDNS ?!?

Do you have a router with mDNS enabled?

STSMiner1:
Nope, my stupid Netgear R8000 router will not resolved hostnames on devices that are not PC's.

Tested this with "hosts" file on PC as well, changes will not save when using "hostname", only works when the IP address of the device is used.

eandersson:
> Nope, my stupid Netgear R8000 router will not resolved hostnames on devices that are not PC's.
> 
> Tested this with "hosts" file on PC as well, changes will not save when using "hostname", only works when the IP address of the device is used.

Yea - hosts files cannot work, as the mining device itself would have no way of knowing the hostname(s) you configured on Windows.

eandersson:
I'll close this PR as while it works fine, it requires a more advanced setup for it to be "safe" / secure and that is likely going to be more confusing to regular users than anything else.

PR #675, opened 2025-01-26
Menu rework, adding menu Design

WantClue:
- Rework on the Design part and put it into a new Menu point
- adding Color Orange


eandersson (approved):
Not tested, but code LGTM

PR #679, opened 2025-01-29
SSID WiFi lookup

WantClue:
add feature to search for nearby ssid's and select them

### DIFF
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -15,6 +15,9 @@
 #include "connect.h"
 #include "main.h"
 
+// Maximum number of access points to scan
+#define MAX_AP_COUNT 20
+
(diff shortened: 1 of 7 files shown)


eandersson (approved):
Code looks fine, but haven't tested it yet.

WantClue:
It's causing a reboot if the bitaxe has no wifi connection and is trying to search for wifi,
also need to eliminate empty or hidden ssid's from the search

kisenberg:
Discovery workds fine, but it also discovers SSID with 32 bytes, but 32 SSID are not connectable by the espminer, only a maximum of 31 byte SSIDs. So 32 bytes SSID should not be selectable in the GUI.

PR #692, opened 2025-02-04
fix(ui): display placeholder for unavailable ASIC temp

steven-s-martins:
Instead of showing `-1` as the ASIC temperature value when it's not yet available, this change displays "--" as a placeholder. This aims to improve the user experience by avoiding a confusing negative temperature during initialization.

Resolves skot/ESP-Miner#680

Before:
before

After:
after

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -146,7 +146,14 @@
                 <div class="grid">
                     <div class="col-12">
 
-                        <h6>ASIC Temperature <span style="float: right;">{{info.temp}}&deg;C</span></h6>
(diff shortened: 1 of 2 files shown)


mutatrum:
The display shows `n/a` before the first reading. Would it make sense to align the two?

skot:
> The display shows `n/a` before the first reading. Would it make sense to align the two?
> 
> 
> 
> https://github.com/skot/ESP-Miner/blob/cbb02389baa4ca0b74c7f958f9a73518038cbe31/main/screen.c#L204

Yes, let's align 'em. I kinda think "--" is better than "n/a" across the board.

steven-s-martins:
Thank you both for the feedback! I have gone ahead and updated the screen to now use "--" as the placeholder.

PR #693, opened 2025-02-04
Improved pool fallback code

eandersson:
This PR improves the fallback mechanism to better handle pool failures by introducing stricter connection management and health checks.

- Added a TCP receive timeout that automatically closes the connection if no data is received for 10 minutes, preventing stale connections.
- Updated the fallback health check to use a minimal stratum handshake, ensuring a valid and healthy connection to the pool.
- Moved connection retry further down to the application layer so that we only reset the counter if we have established a successful stratum handshake with the pool.
- Refactored send_uid to allow multiple stratum connections.
- Moved send_uid for the main connection to global context.

TODO in the future:
- Break heartbeat function into multiple functions.
- Further improve stratum implementation to better (cleaner) allow for multiple connections.

The tl;dr is that the new heartbeat check does something along these lines to validate that the pool is back up.
```
int sock = socket(...)
int err = connect(sock, ...)

int send_uid = 1;
STRATUM_V1_subscribe(sock, send_uid++, ...);
STRATUM_V1_authenticate(sock, send_uid++, ...);
...
int bytes_received = recv(sock, recv_buffer, BUFFER_SIZE - 1, 0);
...
if (strstr(recv_buffer, "mining.notify") != NULL) {
...
sock.close()
```
And the previous heartbeat check would only check if a connection could be established.
```
int sock = socket(...)
int err = connect(sock, ...)
...
sock.close()
```

### DIFF
--- a/components/stratum/stratum_api.c
+++ b/components/stratum/stratum_api.c
@@ -19,19 +19,8 @@ static const char * TAG = "stratum_api";
 static char * json_rpc_buffer = NULL;
 static size_t json_rpc_buffer_size = 0;
 
-// A message ID that must be unique per request that expects a response.
-// For requests not expecting a response (called notifications), this is null.
-static int send_uid = 1;
-
 static void debug_stratum_tx(const char *);
 int _parse_stratum_subscribe_result_message(const char * result_json_str, char ** extranonce, int * extranonce2_len);
-
-void STRATUM_V1_reset_uid()
-{
-    ESP_LOGI(TAG, "Resetting stratum uid");
-
-    send_uid = 1;
-}
 
 void STRATUM_V1_initialize_buffer()
 {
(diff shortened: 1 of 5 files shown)


main/tasks/stratum_task.c
```diff
@@ -125,16 +148,34 @@ void stratum_primary_heartbeat(void * pvParameters)
...
             continue;
         }
+
+        if (setsockopt(sock, SOL_SOCKET, SO_RCVTIMEO , &tcp_timeout, sizeof(tcp_timeout)) != 0) {
+            ESP_LOGE(TAG, "Fail to setsockopt SO_RCVTIMEO ");
+        }
+
+        int send_uid = 1;
+        STRATUM_V1_subscribe(sock, send_uid++, GLOBAL_STATE->asic_model_str);
+        STRATUM_V1_authenticate(sock, send_uid++, GLOBAL_STATE->SYSTEM_MODULE.pool_user, GLOBAL_STATE->SYSTEM_MODULE.pool_pass);
```
Georges760, line 158: don't you want to save the `send_uuid` value into the global after incrementing it twice ?

components/stratum/stratum_api.c
```diff
@@ -319,31 +308,31 @@ int _parse_stratum_subscribe_result_message(const char * result_json_str, char *
     return 0;
 }
 
-int STRATUM_V1_subscribe(int socket, char * model)
+int STRATUM_V1_subscribe(int socket, int send_uid, char * model)
```
Georges760, line 311: was it necessary to manage the send_uid outside ?

Georges760 (approved):

main/tasks/stratum_task.c
eandersson: This is for a separate connection that is just used to determine if the primary is back online. That is why this send_uid isn't shared with the global.

components/stratum/stratum_api.c
eandersson: Yea - this is so that we can create a new temporary connection that won't impacting the existing connection's send_uid count.

main/tasks/stratum_task.c
eandersson: In a future implementation we will probably have two connections open at all times, that is what most vendors do and just switch connection instead of opening and closing connections.

PR #698, opened 2025-02-11
GammaTurbo support and HW abstraction

skot:
This PR adds some much needed hardware abstraction layers. To support new hardware "all" you need to do is implement the functions in

- Thermal functions like fan and temp sensing in `/main/thermal/thermal.c`
- Power monitoring and voltage regulation in `/main/power/power.c` and `/main/power/vcore.c`
- ASIC support in `/asic/asic.c`
and create an appropriate `config.cvs`

I have used this to add support for the new [bitaxeGammaTurbo](https://github.com/bitaxeorg/bitaxeGT) hardware


mutatrum:
Reading through the code, it looks like gammaturbo shouldn't be a device model. The two differences (tps config and fan controller) might be cleaner to be handled by model number instead of device name, as almost all things are the same between gamma and gammaturbo.

skot:
> Reading through the code a bit more, it looks like gammaturbo shouldn't be a device model. The two differences (tps config and fan controller) might be cleaner to be handled by model number instead of device name, as almost all things are the same between gamma and gammaturbo.

Hmm I think that makes sense. By that same token the hex wouldn't be a separate device from the single chip model it shares the ASIC with.

mutatrum:
I would opt for a model name, but have no logic based on that. So every board number has a config (that's already the case), and in that config is defined:
 * Chip type (BM13xx)
 * Chip count (1, 2, 6)
 * Name (Supra, GammaTurbo, SupraHex, etc).

That way there's no conditional logic on the name, as that's only a human readable thing. All firmware logic can be on board number alone. And on further though, it doesn't even need a name in the config, the name can be created based on chip type and count. (1 core BM1397 -> Max, 2 core BM1370 -> GammaTurbo, 6 core BM1368 -> SupraHex, etc). That way at least the naming is uniform, and will stay uniform. Downside is that it's in code, so maybe name in config is simpler.

In a later stage, we can migrate the other hardware options (voltage regulator type and configurations, fan controller type, temp sensor, etc) also to config values, and then all conditional logic makes sense. E.g. only operate a certain component if the board actually has it.

Drawback is it needs a backwards compatible upgrade path, to put those values in for existing devices.

TLDR: Chip type + count should be leading, not the name, IMO.

PR #703, opened 2025-02-13
Don't show -1 temperature on screen

mutatrum:
Follow-up of #692


WantClue (approved):
good one

Hylinus:
In Swarm mode, when devices are restarted, temperature still displays "-1"

PR #714, opened 2025-02-17
feat: unlock OC

WantClue:
this adds in the console function: "unlockSettings()" which can be used in the browser to unlock the settings and set custom values for frequency and voltage

it also removes the old devToolCheck which seems no longer to work on all browser


skot:
how necessary is it to remove the devToolCheck functionality? That still works well on my browsers..

Issue #565, opened 2024-12-08
null values in API crash the device


w3irdrobot:
Note: Issues are not for customer support, configuration or discussion. For those topics please consult with your HW vendor or the OSMU Discord at: https://osmu.bitaxe.org

**Describe the bug**

When updating settings through the API at `PATCH /api/system`, if any values are explicitly `null`, it seems to crash the entire device.

**To Reproduce**
1. Have a Bitaxe.
2. Dance because you own a Bitaxe, and that's awesome.
3. Get device IP
4. Run the command `curl -v -X PATCH -d '{"wifiPass":null}' http://<DEVICE_IP>/api/system`
5. Notice that the device restarts itself (fans spin down, Uptime in web app resets)
6. Cry

**Expected behavior**

I would expect it to be ignored or an error be thrown to let me know I screwed up the request body.

**Screenshots & Photos**
If applicable, add AxeOS screenshots and/or photos of your Bitaxe to help explain your problem.

**Hardware (please complete the following information):**
 - Bitaxe HW version: Gamma 601
 - Bitaxe HW vendor: Solo Satoshi
 - ESP-Miner FW version: v2.4.0 
 - Hash Frequency: 525
 - Voltage: 1150
 - Pool URL, Port, User: Braiins.....

**Additional context**

I checked all the settings and it appeared these ones being set to `null` are when the issues turn up. The others seem to just treat it as `false`.

```
hostname
ssid
wifiPass
stratumURL
stratumUser
stratumPassword
fallbackStratumURL
fallbackStratumUser
fallbackStratumPassword
```

w3irdrobot:
anyone actually working on this yet? if not, i can work on it.

mutatrum:
Fixed by #748

PR #728, opened 2025-02-21
Trim spaces from SSID

w3irdrobot:
I noticed a while ago that the string inputs are not trimmed of spaces. This can cause, in my use-case, issues with the SSID not matching when typing the SSID in manually from a phone when the word Bitcoin is auto-completed.  This PR simply trims the SSID input on the API and client side. I only did this for the SSID for now to simplify the initial PR, since this concept could, and perhaps _should_, be expanded to the other string inputs. This expansion would probably also help with #565.

### DIFF
--- a/main/http_server/axe-os/src/app/components/network-edit/network.edit.component.ts
+++ b/main/http_server/axe-os/src/app/components/network-edit/network.edit.component.ts
@@ -62,6 +62,11 @@ export class NetworkEditComponent implements OnInit {
       delete form.wifiPass;
     }
 
+    // Trim SSID to remove any leading/trailing whitespace
+    if (form.ssid) {
+      form.ssid = form.ssid.trim();
+    }
+
     this.systemService.updateSystem(this.uri, form)
       .pipe(this.loadingService.lockUIUntilComplete())
       .subscribe({


skot:
Thanks for looking into this -- I have been wanting this one for a while!

Curious, is it necessary to trim strings in C, or would just on the frontend be enough?

WantClue (approved):

WantClue:
LGTM
A good start and should be expanded

skot:
> @skot my usual motto/advice/ethos/whatever is that validation and sanitation **have** to happen on the backend API level. it's optional but suggested to also at least validate, if not also sanitize, values on the frontend as well.
> 
> the frontend can be bypassed by just using the API without the frontend. that's why it's most important there. having it also occur in the browser is just good UX so things that happen on the backend are consistent with what the user sees when they submit the form.

I agree that sanitizing inputs on the backend is a good idea. I think in this case though, we're just cleaning up a UI/Autocorrect quirk. If a user submits a SSID with a trailing space over the API, we should prolly assume they meant it, and allow it.

WantClue:
Frontend trimming should be the only goal here. Still it might be confusing to the user.

WantClue (approved):

PR #729, opened 2025-02-21
fix: use url params to unlock overclock

mrv777:


WantClue (approved):

MaSe-Time:
Newbie question but how do you overclock now? On Chrome 'inspect' no longer works

PR #736, opened 2025-02-24
add openapi spec for existing api

w3irdrobot:
adds an [openapi v3.1](https://swagger.io/specification/) document describing the current api. we can use this document to iterate on what a new version of the api might look like. there's alot of existing tooling that exist for this format that the community can utilize to make working with our apis easier. 

the idea for this came from https://github.com/skot/ESP-Miner/pull/667#issuecomment-2676441320

to view, docker can be used. 

```shell
docker run -p 8080:8080 \
    -e SWAGGER_JSON=/workspace/swagger.json  \
    -v $PWD/main/http_server/openapi.yaml:/workspace/swagger.json \
    docker.swagger.io/swaggerapi/swagger-ui
```

open up `localhost:8080` in a browser to view the docs using the swagger ui tool. 

the contents of the file can also just be copy and pasted into [the online editor](https://editor.swagger.io/) as well.

in the future, we can consider options around hosting this, maybe making it available on the bitaxe itself?

### DIFF
--- a/main/http_server/openapi.yaml
+++ b/main/http_server/openapi.yaml
@@ -1,0 +1,508 @@
+openapi: 3.1.0
+info:
+  title: ESP-Miner API
+  description: API specification for ESP-Miner HTTP server
+  version: 1.0.0
+  license:
+    name: GNU General Public License version 3
+    url: https://opensource.org/license/gpl-3-0
+
+servers:
+  - url: http://{device_ip}
+    description: ESP-Miner device
+    variables:
+      device_ip:
+        default: "192.168.1.1"
+        description: IP address of the ESP-Miner device
+
+components:
+  schemas:
+    WifiNetwork:
+      type: object
+      required:
+        - ssid
+        - rssi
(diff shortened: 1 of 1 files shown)


0xf0xx0:
being on the bitaxe would be nice. forgejo hosts theirs at `$host/api/swagger`.

main/http_server/openapi.yaml
```diff
@@ -0,0 +1,508 @@
...
+          description: Hardware board version
+        coreVoltage:
+          type: number
+          description: Configured ASIC core voltage
+        coreVoltageActual:
+          type: number
+          description: Actual measured ASIC core voltage
+        current:
+          type: number
+          description: Current draw in amps
```
0xf0xx0, line 149: now that i can see the draw on my axe, this seems to be milli-amps.

w3irdrobot: good call. updated.

WantClue (approved):

PR #737, opened 2025-02-24
Fix polling loop deatch on getInfo error

mutatrum:
The polling loop of `getInfo` on the dashboard and the logs screen dies if the call returns an error, or is unavailable.

I tried fixing it, but not sure if it works as intended. One solution is mapping error objects to a DEFAULT_SYSTEM_INFO, and add an `error` field. That's the current implementation. It seems to work, however, the `console.log` does not show up. The dashboard does continue polling, so at least that's that. The error part should probably be shown in a toast or something, but I'm out of my depth here.

The alternative is the `filter` call, this should exit the loop completely. However, this would not give any feedback back to the user.

The deleted lines at home.component.ts#L58-59 are unrelated, this just fixes a console error. We no longer have 3 datasets.


mutatrum:
I have trouble testing this as my Bitaxe is getting a new IP address on every reboot and I don't have access to my router to give it a fixed IP address.

mutatrum:
image

WantClue:
please resolve conflicts so this can be merged @mutatrum

WantClue:
@mutatrum if you got time would be nice if you could resolve the conflicts

mutatrum:
Removed the 2.7.0 milestone, I don't have time to resolve conflicts for a good week or so. That, and I'm not sure if I'm happy with the AxeOS notification, this needs more testing.

PR #739, opened 2025-02-24
Expose Wifi RSSI to API endpoint

eandersson:
This commit adds Wi-Fi RSSI to the API endpoint, keeping the implementation minimal to allow the front end to handle advanced logic (e.g., a user-friendly indicator for signal strength).

If the device is not connected to Wi-Fi, the endpoint returns `-90` (representing "poor connectivity"). This approach was chosen because API access inherently requires a working Wi-Fi connection, making this an acceptable fallback.

### DIFF
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -58,6 +58,20 @@ static bool is_scanning = false;
 static uint16_t ap_number = 0;
 static wifi_ap_record_t ap_info[MAX_AP_COUNT];
 
+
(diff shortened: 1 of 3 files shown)


mutatrum:
Can this also be put into global_context, so it can be displayed on the screen?

eandersson:
> Can this also be put into global_context, so it can be displayed on the screen?

Yea - but I think it makes more sense to just re-use the function I implemented, since there is no good loop to update the variable today. The WiFi status update is event based.

WantClue (approved):

PR #741, opened 2025-02-28
Prompt for restart only when needed

### LINKED ISSUE #581: AxeOS 2.4.1 incorrectly states restart is needed to change hostname
_what happens_
1) change hostname

restartfalsealarm

_suggestion_

dustinb:
Closes #581 

Only prompt for restart when it's needed.  The only question is on hostname. Technically restart isn't needed but it won't take until a dhcp lease renewal.  Also think the buttons look better lined up.

image

---

This is the list I've identified that requires restart (Y)

# Network

Hostname: ?
SSID: Y
Wifi-Pass: Y

# Pool Configuation

All pool settings require restart

# Settings

Frequency: N
Core Voltage: N
Flip Screen: Y
Invert Fan Polarity: Y
AutoMatic Fan Control: N
Fan Speed: N
Overheat Mode: N

### DIFF
--- a/main/http_server/axe-os/src/app/components/edit/edit.component.html
+++ b/main/http_server/axe-os/src/app/components/edit/edit.component.html
@@ -81,13 +81,12 @@
         <div class="mt-2">
             <button pButton [disabled]="!form.dirty || form.invalid" (click)="updateSystem()"
                 class="btn btn-primary mr-2">Save</button>
-            <b style="line-height: 34px;">You must restart this device after saving for changes to take effect.</b>
+            <button pButton [disabled]="!savedChanges" *ngIf="requiresRestart()" (click)="restart()">Restart</button>
         </div>
 
         <div class="mt-2">
-            <button pButton [disabled]="!savedChanges" (click)="restart()">Restart</button>
+            <b style="line-height: 34px;" *ngIf="requiresRestart()">You must restart this device after saving for changes to take effect.</b>
         </div>
-
 
     </form>
 </ng-container>
--- a/main/http_server/axe-os/src/app/components/edit/edit.component.ts
+++ b/main/http_server/axe-os/src/app/components/edit/edit.component.ts
@@ -206,6 +206,14 @@ export class EditComponent implements OnInit, OnDestroy {
     this.updateSystem();
   }
 
+  public requiresRestart(): boolean {
+    if (!this.form) return false;
+    const flipscreen = this.form.get('flipscreen');
+    const invertscreen = this.form.get('invertscreen');
+    const invertfanpolarity = this.form.get('invertfanpolarity');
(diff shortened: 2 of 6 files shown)


WantClue:
@dustinb please resolve conflict

dustinb:
> @dustinb please resolve conflict

@WantClue conflicts have been resolved

WantClue:
We're currently debating if this is a good approach in terms of maintainability. ATM AxeOS is convoluted with hard to maintain code and we want to switch as much as possible in an easier format. Said so this PR is really useful but might take some more time to think about it and potential upcoming changes in how we handle the webui in general

mutatrum:
Some more context, I wrote this on Discord:

> What is a bit tricky with this change, and also with f.e. the openapi spec one, there's no guardrails for keeping this logic up to date. These changes create technical debt, and make future changes more difficult and make the barrier to entry for new developer higher. I have no solution on how this can be mitigated, other than some convoluted automatic testing system. 
> Having a list somewhere in the code that dictates which fields require a restart are bound to get out of sync.

One possible way to futureproof this code somehwat better is to add the `restartRequired` to the fb.group parameters:
```
this.form = this.fb.group({
      flipscreen: this.fb.control(info.flipscreen == 1, { restartRequired: true }),
      invertscreen: this.fb.control(info.invertscreen == 1, { restartRequired: true }),
      coreVoltage: this.fb.control(info.coreVoltage, { restartRequired: false, validators: [Validators.required] }),
      frequency: this.fb.control(info.frequency, { restartRequired: true, validators: [Validators.required] }),
      autofanspeed: this.fb.control(info.autofanspeed == 1, { restartRequired: false, validators: [Validators.required] }),
      invertfanpolarity: this.fb.control(info.invertfanpolarity == 1, { restartRequired: true, validators: [Validators.required] }),
      fanspeed: this.fb.control(info.fanspeed, { restartRequired: false, validators: [Validators.required] }),
      overheat_mode: this.fb.control(info.overheat_mode, { restartRequired: false, validators: [Validators.required] })
    });
```
On form submit, if a form element doesn't have this property, it can throw an error. This way, for any developer in the far future, it would be abundantly clear the `restartRequired` property needs to be added and the whole mechanism stays in sync. Something like:
```
    let restartNeeded = false;

    for (const [key, control] of Object.entries(this.form.controls)) {
      if (control.restartRequired === undefined) {
        throw new Error(`Field "${key}" is missing the restartRequired property. All fields must specify if a restart is needed.`);
      }

      if (control.restartRequired && control.dirty) {
        restartNeeded = true;
      }
    }
```
It might feel like overkill, but having these kinds of guardrails keeps the growth of technical debt somewhat contained.

dustinb:
This looks like a good what to do this and make it more explicit.  Could also default to restart required if a new field is added and restartRequired isn't specified.

PR #743, opened 2025-03-02
Remove unit test workaround for qemu

eandersson:
This workaround is no longer required as it was fixed in an upstream release of QEMU.


WantClue (approved):
LGTM

PR #746, opened 2025-03-06
Show share reject reasons

### LINKED ISSUE #264: Handle share reject reasons

mutatrum:
Example from ckpool:

image

Example from public-pool:

image

Reject reasons are pool software specific, so this is different per pool. See the linked issue for some messages.

Fixes #264

### DIFF
--- a/components/stratum/stratum_api.c
+++ b/components/stratum/stratum_api.c
@@ -137,12 +137,15 @@ void STRATUM_V1_parse(StratumApiV1Message * message, const char * stratum_json)
         cJSON * error_json = cJSON_GetObjectItem(json, "error");
         cJSON * reject_reason_json = cJSON_GetObjectItem(json, "reject-reason");
 
-        //if the result is null, then it's a fail
+        // if the result is null, then it's a fail
         if (result_json == NULL) {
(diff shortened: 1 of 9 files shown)


WantClue (approved):

WantClue:
@eandersson please double check

eandersson (approved):

PR #747, opened 2025-03-06
feat: frequency transition

WantClue:
moves the frequency transition into it's own component. 
cleans up code on 1366,1368 and 1370
generic use of all newer chip models


components/asic/include/frequency_transition.h
```diff
@@ -0,0 +1,20 @@
...
+ * @brief Transition the ASIC frequency to a target value
+ * 
+ * This function gradually adjusts the ASIC frequency to reach the target value,
+ * stepping up or down in increments to ensure stability.
+ * 
+ * @param target_frequency The target frequency in MHz
+ * @param asic_type The type of ASIC chip (1366, 1368, 1370)
+ * @return bool True if the transition was successful, false otherwise
+ */
+bool do_frequency_transition(float target_frequency, int asic_type);
```
mutatrum, line 18: This is a strange signature, and makes the code convoluted by passing around the asic type for method call logic, if you want to make it more modular, I would:
 * Change the BM1370_send_hash_frequency method signature to match the others. The id parameter is not used and the max_diff can be a constant in bm1370.c or h.
 * Pass the different BM####_send_hash_frequency methods as a callback.
This way these new files don't know about the asic type.

components/asic/frequency_transition.c
```diff
@@ -0,0 +1,105 @@
...
+#include "bm1368.h"
+#include "bm1370.h"
+#include "esp_log.h"
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+#include <math.h>
+
+const char *FREQUENCY_TRANSITION_TAG = "frequency_transition";
+
+// Track current frequency for each ASIC type
```
mutatrum, line 12: Why multiple floats? Are there multi-chip boards out there that have a mixture of ASIC types?

mutatrum (changes requested):
IMO the code can be simplified quite a bit.

components/asic/frequency_transition.c
Georges760: Mixed ASIC type chain is and will most probably never happen, There are a lot of reason for that (most of them are HW related)

components/asic/frequency_transition.c
```diff
@@ -0,0 +1,105 @@
...
+                break;
+            case 1368:
+                BM1368_send_hash_frequency(current);
+                break;
+            case 1370:
+                BM1370_send_hash_frequency(-1, current, 0.001);
+                break;
+        }
+        
+        vTaskDelay(100 / portTICK_PERIOD_MS);
```
eandersson, line 63: Not sure if it works on all devices, but I changed the `vTaskDelay` to a `uart_wait_tx_done(UART_NUM_1, 100 / portTICK_PERIOD_MS);` on mine and it works great.

components/asic/include/frequency_transition.h
WantClue: I agree good point, I was trying to be as little invasive as possible.

components/asic/frequency_transition.c
WantClue: ye I will remvoe that

WantClue:
@mutatrum check again. Suggested changes are in

main/tasks/power_management_task.c
```diff
@@ -143,7 +143,31 @@ void POWER_MANAGEMENT_task(void * pvParameters)
 
         if (asic_frequency != last_asic_frequency) {
             ESP_LOGI(TAG, "New ASIC frequency requested: %uMHz (current: %uMHz)", asic_frequency, last_asic_frequency);
-            if (do_frequency_transition((float)asic_frequency)) {
+            bool success = false;
+            
+            // Call the appropriate set_frequency function based on the ASIC model
+            switch (GLOBAL_STATE->asic_model) {
```
skot, line 149: I don't want to have any asic_model specific code in the power_management_task. it all needs to go somewhere inside power, thermal or asic.

The goal is to make as few spots as possible in the code to add support for new hardware.

WantClue: shall we divide asic voltage and asic frequency? like voltage into power.c and frequency into asic.c ?

skot: that seems like a good way to do it.

PR #748, opened 2025-03-07
Improve null handling in API settings PATCH

### LINKED ISSUE #565: null values in API crash the device

w3irdrobot:
- Adds a function for getting a string value from a JSON payload attribute and returns a boolean representing if it's a non-null value
- Updates value retrieval from the JSON settings object in the settings PATCH endpoint to correctly handle JSON `null` values as well as values that don't exist in the JSON payload

I am not a C developer. So there may be a more C-like way of doing this, but this does work. :smile: 

Closes #565

### DIFF
--- a/main/http_server/http_server.c
+++ b/main/http_server/http_server.c
@@ -43,10 +43,10 @@ static const char * CORS_TAG = "CORS";
 static esp_err_t GET_wifi_scan(httpd_req_t *req)
 {
     httpd_resp_set_type(req, "application/json");
-    
+
     // Give some time for the connected flag to take effect
     vTaskDelay(100 / portTICK_PERIOD_MS);
-    
+
     wifi_ap_record_simple_t ap_records[20];
     uint16_t ap_count = 0;
 
@@ -374,6 +374,11 @@ static esp_err_t handle_options_request(httpd_req_t * req)
     return ESP_OK;
 }
 
+static bool is_valid_json_string(cJSON *root, cJSON **item, const char *key) {
+    *item = cJSON_GetObjectItem(root, key);
+    return (*item) != NULL && (*item)->type != cJSON_NULL && (*item)->valuestring != NULL;
+}
+
 static esp_err_t PATCH_update_settings(httpd_req_t * req)
 {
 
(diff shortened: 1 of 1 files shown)


skot:
I just checked and cJSON_GetObjectItem() doesn't malloc, so this looks good from a memory leak PoV

main/http_server/http_server.c
```diff
@@ -374,6 +374,11 @@ static esp_err_t handle_options_request(httpd_req_t * req)
     return ESP_OK;
 }
 
+static bool is_valid_json_string(cJSON *root, cJSON **item, const char *key) {
```
eandersson, line 377: We could use `cJSON_IsString` as well here

w3irdrobot: damn that actually looks like it does what we need. good call! I'll update my PR to use that instead.

WantClue (approved):

mutatrum:
If `NULL` is not allowed, it should return a HTTP 400, and make sure the whole request has been ignored.

WantClue:
> we _can_ handle the request that way. it wouldn't necessarily be invalid to just ignore them like we are now on a patch. returning a 400 would also be an API breaking change which I believe we've been avoiding to do until we "redesign" the API.

yes api rework is one of the next bigger things to do. with that this PR looks good and will be merged

PR #717, opened 2025-02-18
Refactor Stratum code for Seamless Failover

eandersson:
This refactor improves Stratum to support multiple simultaneous connections, enabling faster and more reliable failover. It also lays the foundation for more advanced connection strategies, monitoring, and metrics. In addition, the code supports more than two pools (primary and fallback) .

Each connection now maintains its own Stratum queue, with the active connection pushing its messages to a shared ASIC queue. During a failover event, the system seamlessly transitions to pulling messages from the failover connection's queue, minimizing disruption. Additionally, all ongoing work is completed before failing back to the primary pool, reducing wasted effort - especially critical in environments with frequent failovers due to pool or network instability.

PR #755, opened 2025-03-10
Fix difficulty adjustment

mutatrum:
I think this fixes the difficulty adjustment race condition, at least when testing on public-pool and ckpool.

The main point is the removal of 
```
current_connection->stratum_difficulty = next_bm_job->pool_diff;
```
in `asic_task.c`. The other point is adding the same mutex around `MINING_SET_DIFFICULTY` in `stratum_task.c`.

Public-pool is easy, as they request a queue clear whenever the difficulty changes. Ckpool doesn't, however it does complain for a while (about 1 minute or so), this is the same as it was before.

---

Addition: I now see the discussion on #717, so this is moot I guess, unless parts of this code will be salvaged without the secondary_pool being opened.

### DIFF
--- a/main/tasks/asic_task.c
+++ b/main/tasks/asic_task.c
@@ -50,7 +50,6 @@ void ASIC_task(void *pvParameters)
         if (current_connection->stratum_difficulty != next_bm_job->pool_diff)
         {
             ESP_LOGI(TAG, "New pool difficulty %lu", next_bm_job->pool_diff);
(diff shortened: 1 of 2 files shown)


mutatrum:
Closing this now, maybe @eandersson if you want to salvage parts of your code _I think_ this fixes it. I think it's worth having the separation in the code, but maybe with different reconnect logic. There's something to say to f.e. briefly open up the secondary pool connection, to test if it's good or not, and re-open it again if a failover is needed.

PR #769, opened 2025-03-13
ensure timeout in self test

adammwest:
Fixes https://github.com/skot/ESP-Miner/issues/767
* timeout is 5s for hashtest
* uses 85% of expected hashrate vs constants (exception being supra which is 80%)

### DIFF
--- a/main/self_test/self_test.c
+++ b/main/self_test/self_test.c
@@ -464,7 +464,9 @@ void self_test(void * pvParameters)
 
     ESP_LOGI(TAG, "Hashrate: %f", hash_rate);
 
+    float hashrate_test_percentage_target = 0.85;
     float expected_hashrate_mhs = (float)GLOBAL_STATE->POWER_MANAGEMENT_MODULE.frequency_value;
+
     switch (GLOBAL_STATE->device_model) {
         case DEVICE_MAX:
             expected_hashrate_mhs *= BM1397_CORE_COUNT * 4;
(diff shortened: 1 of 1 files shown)


main/self_test/self_test.c
```diff
@@ -448,45 +442,50 @@ void self_test(void * pvParameters)
...
-                tests_done(GLOBAL_STATE, TESTS_FAILED);
-            }
+            expected_hashrate_mhs *= BM1368_CORE_COUNT * 8;
             break;
         case DEVICE_GAMMA:
-            if(hash_rate < HASHRATE_TARGET_GAMMA){
-                display_msg("HASHRATE:FAIL", GLOBAL_STATE);
-                tests_done(GLOBAL_STATE, TESTS_FAILED);
-            }
+            expected_hashrate_mhs *= BM1370_CORE_COUNT * 16;
```
skot, line 479: BM1370_CORE_COUNT * 16 = 128 * 16 = 2048. if this is MH/s then that's too low. If it's GH/s then that's too high.

skot (changes requested):

main/self_test/self_test.c
adammwest: It's 
128*16* freq * 0.75 

Expected starts at 0.75*freq
 For gamma maybe it should be explicit per device case rather than being a multiplier?

skot: I'm not sure where the 900 GH/s for Gamma came from that @benjamin-wilson had in there before, but I think we should stick with that.

WantClue (approved):

skot:
this worked well on my 601. LGTM

PR #779, opened 2025-03-17
fix: changed 'Invert Fan Polarity' text 

### LINKED ISSUE #776: Change "Invert Fan Polarity" text
Seems like people are confusing this with a fan spin direction setting. Let's change the text to "Invert PWM Duty Cycle"

JasonB1833:
- changed above text in the settings menu to "Invert Fan Duty Cycle" for added clarity
closes #776

### DIFF
--- a/main/http_server/axe-os/src/app/components/edit/edit.component.html
+++ b/main/http_server/axe-os/src/app/components/edit/edit.component.html
@@ -123,7 +123,7 @@
             <div class="field-checkbox">
                 <p-checkbox name="invertfanpolarity" formControlName="invertfanpolarity" inputId="invertfanpolarity"
                     [binary]="true"></p-checkbox>
-                <label for="invertfanpolarity">Invert Fan Polarity</label>
+                <label for="invertfanpolarity">Invert Fan Duty Cycle</label>
             </div>
         </div>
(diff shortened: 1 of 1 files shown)


mutatrum:
`Invert Fan PWM Polarity`. But we might need to specify which is which, e.g. is checkbox enabled Polarity High or Polarity Low.

JasonB1833:
> Another suggestion would be `Invert Fan PWM Polarity`. But we might need to specify which is which, e.g. is checkbox enabled Polarity High or Polarity Low.

I'm toying around with the idea of putting a couple radio buttons in, marked high/low respectively with a label above marked "Fan Polarity:" as it might suit the need better. Although I'm going to need a bit to implement it, I've got the html written up I just need to make sure the TS form is working as expected

skot:
Let's not get carried away. This is a really esoteric feature -- I'm not sure we even need it at all.

WantClue:
@JasonB1833 please reopen again and choose the branch dev-latest not master
I was not looking closely

JasonB1833:
@WantClue I'm going to be away from my work station for another day most likely, but I will re-open  with the correct branch as soon as I can.

PR #780, opened 2025-03-18
Create TPS546 VCORE alerts

skot:
The TPS546D24A has several built-in fault modes. Some of them will shutdown the voltage regulator. The goal with this PR is to;
- Show an alert in AxeOS when one of these faults has occured.
- Provide a way to clear the fault and start the TPS546 on Bitaxe reset.

This should address issue #588


mutatrum:
#777 can be leveraged for this as well.

aaron3481:
that save a lot

PR #787, opened 2025-03-22
Persist Swarm sort across updates

dustinb:
Trying to make the least amount of changes since we're in beta test.  But I did update the swarm page so we get fake Bitaxe when running axeos in a local environment.  Needed some test Bitaxe for testing the sort on refresh.

Otherwise it's just persisting the sort field and direction anytime swarm refreshes or scans for new Bitaxe


WantClue (approved):

mutatrum:
Superseded by #928. Thank you for the contribution, nonetheless!

PR #793, opened 2025-03-24
Max power gauges

WantClue:
adds variable gauges for the Power and Input Voltage progress bars, this is in preparation for the GT


main/http_server/axe-os/src/app/components/home/home.component.html
```diff
@@ -130,15 +130,27 @@ <h6>Power <span style="float: right;">{{info.power}} W</span></h6>
...
                                 <ng-template pTemplate="content" let-value></ng-template>
                             </p-progressBar>
-                            <!-- Add marker at 5.0V (90.91% of 5.5V) -->
-                            <div class="absolute bg-white" style="left: 90.5%; top: 0; width: 3px; height: 6px;">
-                                <small class="absolute text-white" style="top: -1.25rem; transform: translateX(-50%)">5.0</small>
+
+                            <!-- Conditional marker based on max voltage -->
+                            <div class="absolute bg-white"
+                              [style.left]="(5 / info.maxVoltage) * 100 + '%'"
+                              *ngIf="info.maxVoltage === 5.5"
```
mutatrum, line 140: :grimacing:

It would be cleaner if the nominal voltage (e.g. 5 and 12) would be in the API endpoint, defined somewhere where other hardware (freq/voltage/etc.) is defined. The fact the UI adds 0.5 is a UI think I guess, this is not a hardware spec. That way these conditionals based on floats can be removed and the code can be simplified.

TLDR: The UI shouldn't differentiate between 5 or 12 in code, that should be all based on the API info.

skot: This is a really good point. Ideally the UI could support any nominal voltage and power with only a change to power.c

WantClue: > 😬
> 
> It would be cleaner if the nominal voltage (e.g. 5 and 12) would be in the API endpoint, defined somewhere where other hardware (freq/voltage/etc.) is defined. The fact the UI adds 0.5 is a UI think I guess, this is not a hardware spec. That way these conditionals based on floats can be removed and the code can be simplified.
> 
> TLDR: The UI shouldn't differentiate between 5 or 12 in code, that should be all based on the API info.

It's mainly in there for the UI of course but the hardware itself is capable of a bit more V then the standard operation Vin.

I see the point of changing it based on the API info

WantClue: And these conditional based markers are there for the UI and I think they should stay

skot: I think the markers should stay, but the values (ie 5.5, etc) should not be in the UI code

skot: how about the API just send nominal values;
- bitaxeMax: 5V, 15W
- bitaxeUltra: 5V, 15W
- bitaxeSupra: 5V, 18W
- bitaxeGamma: 5V, 20W
- bitaxeGT: 12V, 40W
(these numbers could use some tuning)

and then the gauges range from a fixed percentage over and under nominal?

WantClue: I will adjust the max power for the progress bar and eliminate the hard coded markers.

main/power/power.c
```diff
@@ -11,6 +11,21 @@
 #define GAMMA_POWER_OFFSET 5 //Watts
 #define GAMMATURBO_POWER_OFFSET 5 //Watts
 
+// max power settings
+#define MAX_MAX_POWER 20 //watts
```
mutatrum, line 15: lol

mutatrum (approved):

PR #799, opened 2025-03-28
Pid controller

WantClue:
adds PID controller for FAN speed


benjamin-wilson:
Should test with a variety of setups to determine if pid parameters should be set via nvs

PR #800, opened 2025-03-29
Pid controller

WantClue:
adds a PID controller

### DIFF
--- a/main/tasks/power_management_task.c
+++ b/main/tasks/power_management_task.c
@@ -11,6 +11,7 @@
 #include "TPS546.h"
 #include "vcore.h"
 #include "thermal.h"
+#include "PID.h"
 #include "power.h"
 #include "asic.h"
 
@@ -28,45 +29,28 @@
 
 static const char * TAG = "power_management";
 
-// static float _fbound(float value, float lower_bound, float upper_bound)
-// {
-//     if (value < lower_bound)
-//         return lower_bound;
-//     if (value > upper_bound)
-//         return upper_bound;
+double pid_input = 0.0;
+double pid_output = 0.0;
+double pid_setPoint = 55.0;
+double pid_p = 2.0;
+double pid_i = 0.1;
+double pid_d = 5.0;
 
-//     return value;
-// }
-
-// Set the fan speed between 20% min and 100% max based on chip temperature as input.
-// The fan speed increases from 20% to 100% proportionally to the temperature increase from 50 and THROTTLE_TEMP
-static double automatic_fan_speed(float chip_temp, GlobalState * GLOBAL_STATE)
-{
-    double result = 0.0;
-    double min_temp = 45.0;
-    double min_fan_speed = 35.0;
-
-    if (chip_temp < min_temp) {
-        result = min_fan_speed;
-    } else if (chip_temp >= THROTTLE_TEMP) {
-        result = 100;
-    } else {
-        double temp_range = THROTTLE_TEMP - min_temp;
-        double fan_range = 100 - min_fan_speed;
-        result = ((chip_temp - min_temp) / temp_range) * fan_range + min_fan_speed;
-    }
-    PowerManagementModule * power_management = &GLOBAL_STATE->POWER_MANAGEMENT_MODULE;
-    power_management->fan_perc = result;
-    Thermal_set_fan_percent(GLOBAL_STATE->device_model, result/100.0);
-
-	return result;
-}
+PIDController pid;
 
 void POWER_MANAGEMENT_task(void * pvParameters)
 {
     ESP_LOGI(TAG, "Starting");
 
     GlobalState * GLOBAL_STATE = (GlobalState *) pvParameters;
+    
+    // Initialize PID controller
+    pid_init(&pid, &pid_input, &pid_output, &pid_setPoint, pid_p, pid_i, pid_d, P_ON_E, DIRECT);
+    pid_set_sample_time(&pid, POLL_RATE - 1);
+    pid_set_output_limits(&pid, 35, 100);
+    pid_set_mode(&pid, AUTOMATIC);
+    pid_set_controller_direction(&pid, REVERSE);
+    pid_initialize(&pid);
 
     PowerManagementModule * power_management = &GLOBAL_STATE->POWER_MANAGEMENT_MODULE;
     SystemModule * sys_module = &GLOBAL_STATE->SYSTEM_MODULE;
@@ -115,7 +99,11 @@ void POWER_MANAGEMENT_task(void * pvParameters)
 
         if (nvs_config_get_u16(NVS_CONFIG_AUTO_FAN_SPEED, 1) == 1) {
 
-            power_management->fan_perc = (float)automatic_fan_speed(power_management->chip_temp_avg, GLOBAL_STATE);
+            pid_input = power_management->chip_temp_avg;
+            pid_compute(&pid);
+            power_management->fan_perc = (uint16_t) pid_output;
+            Thermal_set_fan_percent(GLOBAL_STATE->device_model, pid_output / 100.0);
+            ESP_LOGI(TAG, "Temp: %.1f°C, SetPoint: %.1f°C, Output: %.1f%%", pid_input, pid_setPoint, pid_output);
 
         } else {
             float fs = (float) nvs_config_get_u16(NVS_CONFIG_FAN_SPEED, 100);
--- a/main/CMakeLists.txt
+++ b/main/CMakeLists.txt
@@ -24,6 +24,7 @@ SRCS
     "./thermal/EMC2103.c"
     "./thermal/TMP1075.c"
     "./thermal/thermal.c"
+    "./thermal/PID.c"
     "./power/TPS546.c"
     "./power/DS4432U.c"
     "./power/INA260.c"
--- a/main/thermal/PID.c
+++ b/main/thermal/PID.c
@@ -1,0 +1,134 @@
+#include "PID.h"
+#include <math.h>
+
+void PID_init(
+    PIDController* pid, 
+    float kp, 
+    float ki, 
+    float kd, 
+    float setpoint, 
+    float output_min, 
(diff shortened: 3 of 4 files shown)


mutatrum:
IMO it would be safer to also keep the existing automatic fan control, and add a dropdown for which fan control algo to use, until we have enough testing data. There are so many different fan setups, heatsinks and environments out there, it would be a miracle if this works immediately on all devices.

main/tasks/power_management_task.c
```diff
@@ -28,45 +29,28 @@
...
 
 void POWER_MANAGEMENT_task(void * pvParameters)
 {
     ESP_LOGI(TAG, "Starting");
 
     GlobalState * GLOBAL_STATE = (GlobalState *) pvParameters;
+    
+    // Initialize PID controller
+    pid_init(&pid, &pid_input, &pid_output, &pid_setPoint, pid_p, pid_i, pid_d, P_ON_E, DIRECT);
+    pid_set_sample_time(&pid, POLL_RATE - 1);
```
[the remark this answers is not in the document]
shufps: The PID control loop runs every $sample_time. But sampling rate of the task is POLL_RATE, so this makes sure the PID control loop is run everytime the temperature is sampled. But the sampling rate should be about the real value and not just `1` because it's used for some scaling internally of the PID library

NilByte:
Please consider making the target temperature configurable instead of fixed 60°C.

If you underclock you may wish to set a target temperature of 45°C and if you overclock a target temperature 70°C may be reasonable compared to the THROTTLE_TEMP 75°C.

mutatrum:
I'm running this on both my Gamma and Supra, the Gamma stays at 35% and 55°, the Supra at 100% and 65°. So not really a proper test, maybe later today when it gets warmer here I get more feedback.

WantClue:
> Please consider making the target temperature configurable instead of fixed 60°C.
> 
> If you underclock you may wish to set a target temperature of 45°C and if you overclock a target temperature 70°C may be reasonable compared to the THROTTLE_TEMP 75°C.

I will add the configureable setpoint

WantClue:
> IMO it would be safer to also keep the existing automatic fan control, and add a dropdown for which fan control algo to use, until we have enough testing data. There are so many different fan setups, heatsinks and environments out there, it would be a miracle if this works immediately on all devices.

The plan is to test this extensivley and integrate this into 2.6.3 if this is save to use

MaSe-Time:
This is exciting to see my idea take shape 😬.

Running it on my three today so should see it adjust as the temperatures rise. For me I'd like to see the minimum fan speed be reduced to 20% though as when ambient is cool as per my image, 35% isn't low enough. 

Screenshot_20250405_062605_Chrome

main/http_server/http_server.c
```diff
@@ -609,6 +612,7 @@ static esp_err_t GET_system_info(httpd_req_t * req)
     cJSON_AddNumberToObject(root, "autofanspeed", nvs_config_get_u16(NVS_CONFIG_AUTO_FAN_SPEED, 1));
 
     cJSON_AddNumberToObject(root, "fanspeed", GLOBAL_STATE->POWER_MANAGEMENT_MODULE.fan_perc);
+    cJSON_AddNumberToObject(root, "temptarget", nvs_config_get_u16(NVS_CONFIG_TEMP_TARGET, 60));
```
0xf0xx0, line 615: this should be added to main/http_server/openapi.yaml

MaSe-Time:
My three ran perfectly yesterday ran upto 100% fan speed as ambient rose and back down to 35% through the evening.

I'd very much like the lower end fan threshold to be around 20% though. 👍

mutatrum:
Is there a reason low end can't be 0%?

MaSe-Time:
> Is there a reason low end can't be 0%?

Someone mentioned somewhere that not all fans modulate down to that level so if 15% was called for example the fan would actually stop but I don't see that as much of an issue as if the fan did stop the temp would rise and trigger it to start again at what ever level it works at.

I'd prefer 0%  to be honest as when I've been manually adjusting mine I sometimes need 9%.

mutatrum:
The fan on the supra I have only starts spinning at ~10%. On restart it takes a bit to get the temperature right, it overshoots about 5C until it comes back down to the set point:

image

Maybe a more advanced way would be to do a calibration run on startup, see what the RPM curve is over the fan percentage, and then steer the PID based on desired RPM instead of fan percentage?

main/tasks/power_management_task.c
```diff
@@ -115,7 +100,11 @@ void POWER_MANAGEMENT_task(void * pvParameters)
 
         if (nvs_config_get_u16(NVS_CONFIG_AUTO_FAN_SPEED, 1) == 1) {
 
-            power_management->fan_perc = (float)automatic_fan_speed(power_management->chip_temp_avg, GLOBAL_STATE);
+            pid_input = power_management->chip_temp_avg;
```
mutatrum, line 103: At boot the `chip_temp_avg` is  -1, I think the pid_compute should be skipped then.

main/tasks/power_management_task.c
```diff
@@ -115,7 +100,11 @@ void POWER_MANAGEMENT_task(void * pvParameters)
...
+            pid_compute(&pid);
+            power_management->fan_perc = (uint16_t) pid_output;
+            Thermal_set_fan_percent(GLOBAL_STATE->device_model, pid_output / 100.0);
+            ESP_LOGI(TAG, "Temp: %.1f°C, SetPoint: %.1f°C, Output: %.1f%%", pid_input, pid_setPoint, pid_output);
```
mutatrum, line 107: Also log `power_management->fan_rpm`. This will be useful when troubleshooting weird fan setups.

mutatrum:
I rad a test with my Supra with crappy fan and heatsink, from a cold start, with minimum at 0%:

chart (1)
(left axis is both temps and percentage, right axis is fan rpm)

It takes about 100 seconds to reach peak temp of 65, and another 50 seconds or so to get it back down. This is in a warm environment, ambient temp was 27C. Even with a deadzone of 10%, it picked up the fan rpm quick enough to prevent too far of an overshoot.

On the other hand, the Gamma I have on stock can be happy below 10%, it doesn't have a deadzone, it's the opposite: at 0% the fan still runs around 3000 rpm.

So I would say the PID controller is absolutely capable of handling both scenario's from 0%, both run better with this and a lot quieter and stable compared to the old fan control.

MaSe-Time:
Not sure if me posting is useful but my three have worked well down to 20% so far. Not quite cool enough for then to call for 15%. If they weren't clocked so high they for sure would need a hell of a lot less fan. 

Screenshot_20250407_052507_Chrome

zcht:
It seems to work well.

However, I wouldn't set a target of 60°C, but rather 50°C. The reason for this is simple: if the fan runs just a little faster and can lower the temperature, the ASIC automatically consumes less power. Less power consumption = automatically lower temperatures. So you would optimize the process in both directions and on top of that, you would also get significantly quieter operation. 

From 65°C I would activate some kind of hard panic mode that sets the fan to 100% until the temperature drops back to 60°C. 

In my tests, the ASIC at 50°C uses almost 2 watts less, but as soon as you are at 50°C and use less, the temperature and fan speed automatically drop. 

With these processes, we would have covered all eventualities and optimized the process perfectly.

MaSe-Time:
In my testing yesterday with target set at 60 the temps overshot that because my ambient went upto 30+c. They were between 62 and 65 degrees with 100% fan speed. I belive if it hits 70c it shuts down so in my mind it kinda works how it should?

NilByte:
On Gamma 601 I noticed that restart is required in order to apply changed target temperature setting.

Any reason why changed target  temperature couldn't be applied on the fly when saving settings as it works for changed voltage and frequency?

main/http_server/axe-os/src/app/services/system.service.ts
```diff
@@ -63,6 +63,7 @@ export class SystemService {
           invertfanpolarity: 1,
           autofanspeed: 1,
           fanspeed: 100,
+          temptarget: 66,
```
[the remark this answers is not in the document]
WantClue: 66 for the lulz I will set the default to 66 as well

main/tasks/power_management_task.c
WantClue: done

WantClue:
> On Gamma 601 I noticed that restart is required in order to apply changed target temperature setting.
> 
> Any reason why changed target temperature couldn't be applied on the fly when saving settings as it works for changed voltage and frequency?

Working on that ATM

main/tasks/power_management_task.c
```diff
@@ -28,45 +29,28 @@
...
     ESP_LOGI(TAG, "Starting");
 
     GlobalState * GLOBAL_STATE = (GlobalState *) pvParameters;
+    
+    // Initialize PID controller
+    pid_init(&pid, &pid_input, &pid_output, &pid_setPoint, pid_p, pid_i, pid_d, P_ON_E, DIRECT);
+    pid_set_sample_time(&pid, POLL_RATE - 1);
+    pid_set_output_limits(&pid, 35, 100);
+    pid_set_mode(&pid, AUTOMATIC);
+    pid_set_controller_direction(&pid, REVERSE);
```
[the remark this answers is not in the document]
WantClue: - AUTOMATIC mode:

The PID controller continuously calculates the output based on the current input, setpoint, and tuning parameters. It actively adjusts the output to minimize the error.

- MANUAL mode:

The PID controller stops automatic adjustments. The output is either fixed or controlled externally. This is useful for tuning, testing, or when you want to override the PID temporarily.

- DIRECT:

An increase in input (e.g., temperature) causes an increase in output (e.g., fan speed).

Use this when the output should move in the same direction as the error.

- REVERSE:

An increase in input causes a decrease in output.

Use this when the output should move in the opposite direction of the error.

It is possible to add a dynamically mode change to it maybe in the future, I'm peeking into auto tuning functionallity. Therefore I would leave it in here for now

PR #805, opened 2025-03-29
Add difficulty tooltip to show full number

timrosenbach:
This PR adds a tooltip for the achieved difficulties. On hover non suffixed numbers are shown.
grafik

To achieve this two new values are added to the /info endpoint:

- `bestNonceDiff` : number
- `bestSessionNonceDiff` : number


0xf0xx0:
why not replace `bestDiff` and `bestSessionDiff`, and use a small function in the webui to generate the suffixed strings?

timrosenbach:
> why not replace `bestDiff` and `bestSessionDiff`, and use a small function in the webui to generate the suffixed strings?

Kinda did that in the first commit but was not sure if I should change an existing api without deprecation first. If thats okay for you guys I'll change `bestDiff` and `bestSessionDiff` to numbers and add a pipe to format it.

I'll also take a look at the swarm screen then.

mutatrum:
Replaced by #1399

Issue #802, opened 2025-03-29
crash when searching for AP list in setup mode


skot:
- dev-latest 4fc925a
- Bitaxe 601

```
I (119170) wifi_station: Could not connect to 'myssid' [rssi -128]: reason 201
I (119170) wifi_station: Retrying Wi-Fi connection...
I (119210) http_server: Redirecting to root
I (119350) http_server: File sending complete
I (119910) http_server: File sending complete
I (119980) http_server: Redirecting to root
I (119980) http_server: File sending complete
I (120080) http_server: File sending complete
I (120230) example_dns_redirect_server: Received 31 bytes from 192.168.4.2 | DNS reply with len: 47
I (120230) example_dns_redirect_server: Waiting for data
IP address: 1.4.168.192
I (120230) example_dns_redirect_server: Received 31 bytes from 192.168.4.2 | DNS reply with len: 47
I (120240) example_dns_redirect_server: Waiting for data
I (120810) wifi:<ba-add>idx:3 (ifx:1, b2:73:34:dc:71:39), tid:1, ssn:2, winSize:64
E (122080) httpd: Failed to post esp_http_server event: ESP_ERR_TIMEOUT
E (124080) httpd: Failed to post esp_http_server event: ESP_ERR_TIMEOUT
I (124180) wifi_station: Could not connect to 'myssid' [rssi -128]: reason 201
I (124180) wifi_station: Retrying Wi-Fi connection...
E (129170) esp_netif_lwip: ip lost timer: failed to post lost ip event (107)
I (129190) wifi_station: Could not connect to 'myssid' [rssi -128]: reason 201
I (129190) wifi_station: Retrying Wi-Fi connection...
I (129490) http_server: File sending complete
I (129920) http_server: File sending complete
I (129920) CORS: Device in AP mode. Allowing CORS.
W (129930) wifi:Haven't to connect to a suitable AP now!
E (131940) httpd: Failed to post esp_http_server event: ESP_ERR_TIMEOUT
E (133940) httpd: Failed to post esp_http_server event: ESP_ERR_TIMEOUT
I (134190) wifi_station: Could not connect to 'myssid' [rssi -128]: reason 201
I (134190) wifi_station: Retrying Wi-Fi connection...
I (134240) CORS: Device in AP mode. Allowing CORS.
W (134240) wifi:Haven't to connect to a suitable AP now!
I (134410) http_server: File sending complete
I (137010) http_server: File sending complete
[...]

PR #815, opened 2025-04-01
Improve AP mode stability by stopping unnecessary Wi-Fi reconnections

eandersson:
When the device is in AP mode, repeatedly attempting to connect to an incorrect or non-existent Wi-Fi network causes instability. This patch halts the reconnection process if a client is already connected to the AP, ensuring a stable connection.

### DIFF
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -58,20 +58,6 @@ static bool is_scanning = false;
 static uint16_t ap_number = 0;
 static wifi_ap_record_t ap_info[MAX_AP_COUNT];
 
-bool is_client_connected_to_ap() {
-    wifi_sta_list_t sta_list;
-    esp_err_t err = esp_wifi_ap_get_sta_list(&sta_list);
-
-    if (err != ESP_OK) {
-        ESP_LOGE(TAG, "Failed to get station list (error: %d)", err);
-        return false;
-    }
-
-    ESP_LOGD(TAG, "Found %d clients connected", sta_list.num);
-
-    return sta_list.num > 0;
-}
-
 esp_err_t get_wifi_current_rssi(int8_t *rssi)
 {
     wifi_ap_record_t current_ap_info;
@@ -146,6 +132,7 @@ esp_err_t wifi_scan(wifi_ap_record_simple_t *ap_records, uint16_t *ap_count)
 }
 
 static int s_retry_num = 0;
+static int clients_connected_to_ap = 0;
 
 static char * _ip_addr_str;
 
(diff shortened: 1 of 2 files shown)


WantClue (approved):

mutatrum:
Good change, makes it a lot more stable. Code can be simplified by checking on event_id `WIFI_EVENT_AP_STACONNECTED` for user connect to STA and keep state in a static boolean. The opposite is `WIFI_EVENT_AP_STADISCONNECTED`, so the retrying can continue if user disconnects (for whatever reason).

mutatrum (changes requested):
Good change, makes it a lot more stable. Code can be simplified by checking on event_id WIFI_EVENT_AP_STACONNECTED for user connect to STA and keep state in a static boolean. The opposite is WIFI_EVENT_AP_STADISCONNECTED, so the retrying can continue if user disconnects (for whatever reason).

STSMiner1:
V2.6.1 - First time setup - Wi-Fi typo error results in the Bitaxe AP dropping thus stopping you from fixing the typo error when connecting to it's AP. 

I found an issue that if you make a typo error with the first time setup with the Wi-Fi SSID and you've clicked "Save" and "Restart", you will not be able to fix the issue when reconnecting to the Bitaxe AP, the connection drops and connection between your PC / cell phone is lost with the Bitaxe.

The same applies if your changing out your Wi-Fi SSID for another one and you make a typo mistake here.
_Note, you will not be able to change your Wi-Fi SSID via the AP of the Bitaxe if your running v2.6.1_

Image below shows that the PC is not connected to the AP of the Bitaxe, the connection was dropped by the Bitaxe and the browser timed out because of that.

The only way I found to resolve this issue was to flash the Bitaxe to it's factory state with the Web Flasher or with the BitaxeTool with the default configs (no entry for Wi-Fi and password) and start again with the setup (first time setup).

How did I stumble on this issue ?

I was trying to reproduce the issue described here -->  https://github.com/bitaxeorg/ESP-Miner/issues/802

I was not able to reproduce this issue (802), thus leading to the issue I've described above, this PR (#815) resolves this issue I found.


image

eandersson:
> Good change, makes it a lot more stable. Code can be simplified by checking on event_id WIFI_EVENT_AP_STACONNECTED for user connect to STA and keep state in a static boolean. The opposite is WIFI_EVENT_AP_STADISCONNECTED, so the retrying can continue if user disconnects (for whatever reason).

I'm not convinced this simplifies things, though it’s probably more robust. Also, not sure if we even care about retrying here, since we've already failed multiple times while waiting for the user to connect to the AP. This is really just fixing an edge case that very few users would care about - for example, after having typo'd the wifi password, or maybe the password changed.

Feel free to open a PR though if you feel strongly about it.

mutatrum (approved):
LGTM. Agree that continuing after user disconnect is not neccesary. At that point user had successfully connected to the setup AP, so it should be fine.

PR #809, opened 2025-03-31
Move to ESP IDF 5.4.1

eandersson:
Lots of bug fixes in the latest stable release.
https://github.com/espressif/esp-idf/releases/tag/v5.4.1

PR #819, opened 2025-04-02
Handle all Wi-Fi error messages

mutatrum:
This refactor is a bit of an oil spill. It started with the way different fields were passed into and out of connect.c, which was a mixed bag. I wanted to clean this up and encountered several bits that needed some TLC. It's clear that GlobalState pulls in a lot of other header files, which tie everything together in a way that needs further detanglement. This also raises the question why the Wi-Fi code is in a component, and maybe the same question for the other components as well.

 * Add all Wi-Fi error codes
 * Improve errors on display
 * Set Wi-Fi fields in `GlobalState`
 * Remove `EventGroupHandle` for Wi-Fi
 * Remove callback functions from `main.c`
 * Generate `ap_ssid` only once
 * Use `NVSDevice_get_wifi_creds` in `main.c` instead of duplicated code


mutatrum:
The reason for this change was several people asking about Wi-Fi error codes with only an error code. This should improve the feedback.

main/main.h
```diff
@@ -3,8 +3,6 @@
 
 #include "connect.h"
 
-void MINER_set_wifi_status(wifi_status_t status, int retry_count, int reason);
-void MINER_set_ap_status(bool enabled);
 void self_test(void * pvParameters);
```
mutatrum, line 6: self_test should probably be somewhere else as well. Save that for another say.

mutatrum:
This can be integrated as well, incorrect password is a bit of a mixed bag of error messages with this change.

https://github.com/bitaxeorg/ESP-Miner/pull/809#issuecomment-2767178357

WantClue (approved):
LGTM, the screen stuff got in there by accident I assume ? 😅

mutatrum:
> LGTM, the screen stuff got in there by accident I assume ? 😅

Taking out the screen stuff, should be a separate PR indeed.

mutatrum:
Done, see #827.

WantClue (approved):

PR #832, opened 2025-04-09
Generate angular api services and model from openapi spec file

timrosenbach:
This PR add generation for api services and models based on the openapi specification file. Removes existing services and refactors existing components to use new generated api services. For swarm functionality a wrapper service is added that makes the IP configurable for the generated services.


0xf0xx0:
could you rebase this pr? this would be useful while we're cleaning up the api and frontend :3

PR #839, opened 2025-04-12
feat(ci): Add Angular unit test workflow and fix test suite

d4r1as:
- Create dedicated GitHub Actions workflow for Angular tests
- Add `npm` command to run headless Chrome tests in CI environment
- Add test coverage reporting and artifact upload
- Fix app component tests to match actual template structure
- Fix test configurations across multiple components:
  - Edit, Home, Logs components
  - Network, NetworkEdit components
  - Settings, Swarm components
  - System, WebSocket, GithubUpdate services
- Add missing HttpClientTestingModule and CUSTOM_ELEMENTS_SCHEMA


.github/workflows/angular-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    - name: Run tests
+      run: npm run test-ci
+
+    - name: Upload coverage reports
+      uses: actions/upload-artifact@v4
+      if: always()
+      with:
+        name: coverage-report
+        path: ./main/http_server/axe-os/coverage/cobertura-coverage.xml
```
eandersson, line 38: I think we should be able to upload an HTML (aka human readable) report here in addition to the XML one right?

eandersson: Also, do we want to set some sort of minimal coverage required?

PR #840, opened 2025-04-12
Alternative OC GUI

dskvr:
I am now using this on my supras/gammas, figured I would share; no idea if this meets your standards/expectations. glhf.

##  Features
- Risk Acknowledgement
- User configureable, named presets
- +/- 1/10/25/50/100 (constrained to frequency ramp)
- Built-in (non-removable) default preset.
- Frequeency/Voltage value color range
- Updated tests
- Additive; original interface left in tact.
- Hotkeys (arrow up/arrow down)

**Test by appending `?oc` to the settings url.**

Screen Shot 2025-04-12 at 12 32 21 PM
Screen Shot 2025-04-12 at 12 32 25 PM

##  TODO
- [x] **Optimize CSS to meet budget.**
- [x] Adjust min/max for chipsets
- [ ] Add more (removable) default presets?
- [x] Add more tests
- [x] Fix/remove "Disable Overclock Mode"
- [ ] Mobile Compatabilty  
- [ ] Squash commits

## Why? 

The dropdowns bother me because they are in 25-50 step increments, I had already modified the OC values a few weeks ago, built and deployed them on my bitaxes. However, I did some testing and found (likely not the first to find this) that each device/chip behaves slightly differently and small increments/decrements (such as +/-1) have a significant impact on stability (and sometimes performance). 

### Aside
I went down this path originally because one of my supras was unstable at default clock speeds (i posted something about this on reddit ~6 months ago). I found that by undervolting just a little bit and then bumping the clock speed just a little bit, the troubled supra ran like a dream at close-to-stock factory settings. This could be due to faulty hardware or bad assembly, unsure, all I know is that a minor software change fixed my issue.


skot:
Due to the way the ASIC hash frequency registers work, we can actually set every frequency on Bitmain chips. There should be a message in the log if a specific frequency can’t be set. Maybe we should expose that error to AxeOS?

dskvr:
That would be useful

mutatrum:
> Due to the way the ASIC hash frequency registers work, we can actually set every frequency on Bitmain chips. There should be a message in the log if a specific frequency can’t be set. Maybe we should expose that error to AxeOS?

Can the frequency ramp be exposed to the front-end?

skot:
> Due to the way the ASIC hash frequency registers work, we can't actually set every frequency on Bitmain chips. There should be a message in the log if a specific frequency can’t be set. Maybe we should expose that error to AxeOS?

Sorry typo; "we **can't** actually set every frequency"

dskvr:
> My point is that when you're doing +/- 1 MHz tuning you might not be changing anything.

Good to know, are the acceptable values a static list or can they be calculated?

dskvr:
Added calculations and UX considerations. Added one more class, under budget. 

https://github.com/user-attachments/assets/a7019bd0-ecc8-4499-a325-42d0357eb9ec

dskvr:
I improved the logic, but something is still off. Also broke something in the process, form is never marked as dirty now. 

Will revisit later.

dskvr:
Everything should work now, however, there is a caveat

Since floats are not stored in NVS, I attempt to derive the correct value on init, however, there could be situations where the wrong value is derived; resulting in a frustrating UX loop where you save, come back, and onload the value has been altered to one you did not set (and on the backend, it is set correctly, but if you do not notice and save again, it will save the altered value). This is an anti-pattern, but don't know of any other solution.

I haven't been able to find such a case, but I know it exists.

WantClue:
> > > My point is that when you're doing +/- 1 MHz tuning you might not be changing anything.
> > 
> > 
> > Good to know, are the accepted values a static list or can they be calculated?
> 
> they are calculated by the send_hash_frequency() function for that respective ASIC. For example the BM1370 (gamma) function:
> 
> https://github.com/bitaxeorg/ESP-Miner/blob/412b33fea966921b8c99868bd8758724f4d87116/components/asic/bm1370.c#L148-L203

basically every frequency that is dividable by 6.25 can be applied, we could adjust this. I was thinking about playing around with this

dskvr:
I have made larger changes to the settings GUI as a whole in [this branch](https://github.com/dskvr/ESP-Miner/tree/feature/oc-alt-gui). I kept this separate to avoid too many changes at once.  👍 or 👎 compared to GUI already shared above.

Screen Shot 2025-04-19 at 9 44 43 AM

skot:
> I have made larger changes to the settings GUI as a whole in [this branch](https://github.com/dskvr/ESP-Miner/tree/feature/oc-alt-gui). I kept this separate to avoid too many changes at once. 👍 or 👎 compared to GUI already shared above.
> 
> Screen Shot 2025-04-19 at 9 44 43 AM

I guess I have mixed emotions about this layout. This seems like a lot of GUI real estate dedicated to this task. Does it look to complicated? Overwhelming for the average user?

I like this functionality a lot. Part of me would rather see this (and more) in a separate app

dskvr:
> I guess I have mixed emotions about this layout. This seems like a lot of GUI real estate dedicated to this task. Does it look to complicated? Overwhelming for the average user?

I hear ya. A counter argument is that the average user wouldn't add `?oc` to the URL. 

IMO OC should come stock, however, other features such as histograms and more advanced functionality should absolutely be in a separate app. 

Also having an accessible, smooth, modern OC experience inspires things [like this](https://njump.me/nevent1qywhwumn8ghj7mr4de3kscn00qh8xctwv3mkjcmg9enxzund9uq32amnwvaz7tmjv4kxz7fwv3sk6atn9e5k7tcprdmhxue69uhhw6r9v96zu6rpwpc8jarpwejhym3wvdhj7qgjwaehxw309ahx7um5wf6k2tnrdakj7qgwwaehxw309ahx7uewd3hkctcpzamhxue69uhhyetvv9ujumn0wd68ytnzv9hxgtcpzemhxue69uhkzat5dqhxummnw3erztnrdakj7qf4waehxw309uerxctnv3nxzumxxferxdp3vahxyunjdp4xsam8vaskgenxvashxennv9jxvctnv9nxztnc09az7qg6waehxw309askgatvwshrzwrsd36hxtnnda3kjctv9uq32amnwvaz7tmpv4nkjueww468smewdahx2tcpzamhxue69uhkzarvv9ejumn0wd68ytnvv9hxgtcpzpmhxue69uhkztnwdaejumr0dshszrthwden5te0vfexytnfduhsz9rhwden5te0v9kxwmeww468smewdahx2tcpr3mhxue69uhkyatrddjhgtnrdaexzcmvv5h8xmmrd9skctcpzdmhxue69uhkyunz9e5k7tnjv4kxz7f0qy2hwumn8ghj7cnvv9ehgu3wvcmh5tnc09az7qg5waehxw309uerxt338qezuvfj8qhrxve0qyd8wumn8ghj7cn0wd68ytnvv43hgatjd9n8jtnwv46z7qfqwaehxw309a3xjarrda5kumtp0p5k6ctvd9ehguewdahxc6twv5hsqg9kc0j3ml8yxdyutfaqquwpgd2s5zrq3amsuyz39e759qvuukfvm509c64t)

benjamin-wilson:
Nack
Some interesting ideas and you've definitely put in a lot of work into this, but I'm not sure it would be beneficial to expand the complexity of this area for the average user. There is probably a better way to do what we're currently doing but expanding it to a whole feature might be overkill.

There's also the issue of hardware specific code which we're currently trying to remove.

PR #845, opened 2025-04-13
ft: Add support for InfluxDB

d4r1as:


WantClue:
I think this would be awesome if this could be added into hash-command https://github.com/bitaxeorg/hash-command

d4r1as:
> I think this would be awesome if this could be added into hash-command https://github.com/bitaxeorg/hash-command

@WantClue 

I can look into it in the next couple of days

Issue #350, opened 2024-09-25
Consolidation of serial comms issues


adammwest:
This is a super issue relating to all the chip related data. 
I will try and keep this updated and not stale

**Related issues**
https://github.com/skot/ESP-Miner/issues/69 [PR merged, DONE]
https://github.com/skot/ESP-Miner/issues/24 [PR made]
https://github.com/skot/ESP-Miner/issues/286
https://github.com/skot/ESP-Miner/issues/395 [PR merged, DONE]
https://github.com/skot/ESP-Miner/issues/248  [PR made]

**Items on List**
Key
[ ] not started
[X] done
[=] started

List
[X] the warning if consecutive RX have no messages or they are failing to convert to nonces
[=] the counting of TX (work) send fails
[  ] the counting of successful conversion of RX_bytes->nonces and RX_Bytes==0
[=] make RX byte processing handle offset problems
[X] handle crc from nonces
[X] determine fullscan for chips

**(2) Metric TX send % (no issue)**
 bad TX under fast CPU conditions
https://github.com/skot/ESP-Miner/pull/462

https://github.com/skot/ESP-Miner/blob/master/components/asic/bm1366.c#L623
https://github.com/skot/ESP-Miner/blob/master/components/asic/bm1368.c#L357
https://github.com/skot/ESP-Miner/blob/master/components/asic/bm1370.c#L435

for all chips the return value of send_work is ignored, this can be counted, in a metric like send TX fails/total TX  send
counting the raw values rather than the division is preferable.

TX_send % = end TX fails/total TX  send
the metric proves data is sent to the chip

if the fullscan_ms is not less than half of the time when dups appear, a TX send fail will result in duplicates, it has explanatory power.

**(3) Metric asic message conversions to nonces (no issue)**
conversion_rate = chip_diff_nonces /(asic_rx_bytes/11)

this metric calculated how many bytes recived from the chip and compares it to how many nonces have been recovered
it can be thought of how good the RX handling is.

**(4) For RX byte handling https://github.com/skot/ESP-Miner/issues/24**
a fix for this problem may also resolve https://github.com/skot/ESP-Miner/issues/286

there is already a pr https://github.com/skot/ESP-Miner/pull/48/commits
it would need to be extended for BM68/70  and tested

if misalignment by 1 byte many cases where nonce could exist this is not handled

0xYY represents any value
0xYY 0x55 0xaa 11 bytes fails but could be nonce
[...]

PR #851, opened 2025-04-16
Serial Buffer Alignment for non zero preamble offset

adammwest:
**Links**
Fixes https://github.com/bitaxeorg/ESP-Miner/issues/24
Refs https://github.com/bitaxeorg/ESP-Miner/issues/350

**Problem Definition**

What happens currently
consider this toy example 
our expected message is in the format  55 AA XX XX XX XX CC with length 7

the following uart buffer
UART BUF = 00 55 AA XX XX XX XX CC
**Currently**
1. we request size 7
2. we get 
UART BUF = CC
buffer = 00 55 AA XX XX XX XX
4. we fail the preamble test
5. clear the UART BUF

result a lost nonce

**this change does**
UART BUF = 00 55 AA XX XX XX XX CC
1. request size 7
2. we get 
UART BUF = CC
buffer = 00 55 AA XX XX XX XX
4. find offset of 55 aa, in this case its 1
5. offset the buffer by 1 
6. UART BUF = CC
buffer = 55 AA XX XX XX XX

7. request the offset in this case 1
UART BUF = 
buffer = 55 AA XX XX XX XX CC
8. complete tests

in this case we have overcome a serial offset error
in the normal case where everything works 
we dont have an increase of buffer reads
in the case where we have an offset we increase the reads by 1

**Status** Untested

### DIFF
--- a/components/asic/common.c
+++ b/components/asic/common.c
@@ -88,6 +88,36 @@ int count_asic_chips(uint16_t asic_count, uint16_t chip_id, int chip_id_response
     return chip_counter;
 }
 
+void shift_buffer_left(uint8_t *buffer, size_t len) {
+    if (len < 1) return;
+    memmove(buffer, buffer + 1, len - 1);
+}
+
+int find_preamble_offset(uint8_t *buffer, int buffer_size) {
+    int preamble_offset  = -1;
+    for (int i = 0; i < buffer_size; i++) {
+        uint16_t received_preamble = (buffer[i] << 8) | buffer[i+1];
+        if (received_preamble == PREAMBLE)  {
+            preamble_offset = i;
+            break;
+        }
+    }
+
+    return preamble_offset; // -1 not found, >=0 location
+}
+
+esp_err_t serial_alignment(uint8_t * buffer, int buffer_size, int preamble_offset) {
+    shift_buffer_left(buffer,preamble_offset);
+    uint8_t reserve_buf[11] = {0};
+    int reserve_received = SERIAL_rx(reserve_buf, preamble_offset, 10);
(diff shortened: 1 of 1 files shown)


mutatrum:
Commits 08e7ca1 and 412b33f should be taken out.

mutatrum:
I've never seen this happen, is this something specific for some chips?

Georges760 (approved):

Georges760:
> I've never seen this happen, is this something specific for some chips?

that"s very common for uart related communication, to have framing out of sync because a glitch that add a ghost byte in the buffer. It is a good practice to have any sort of resync anyway...

mutatrum:
> > I've never seen this happen, is this something specific for some chips?
> 
> that"s very common for uart related communication, to have framing out of sync because a glitch that add a ghost byte in the buffer. It is a good practice to have any sort of resync anyway...

Ok, that makes sense. But I doubt this will fix it, then. There is a strict length check, so if the feed is misaligned, won't it read a different amount of bytes?

adammwest:
> > > I've never seen this happen, is this something specific for some chips?
> > 
> > 
> > that"s very common for uart related communication, to have framing out of sync because a glitch that add a ghost byte in the buffer. It is a good practice to have any sort of resync anyway...
> 
> Ok, that makes sense. But I doubt this will fix it, then. There is a strict length check, so if the feed is misaligned, won't it read a different amount of bytes? Maybe I'm just misunderstanding the situation, btw.

As I understand if you ask for size 11 and there is noise at the beginning you still get 11 but with the noise, this means the meaningful byte at the end will be in the uart buffer but not in the local buffer.
Then the message if a nonce is lost because the noise is parsed.

WantClue:
@adammwest can you check on the build failures ?

PR #854, opened 2025-04-18
Initial page load causes multiple inconsistent GET responses

### LINKED ISSUE #794: Initial page load causes multiple inconsistent GET responses
From the developer tools in Firefox I can see that the server sends back the same `/api/theme` response multiple times (7) on an initial page load. The responses are inconsistent, which for instance causes invalid theme settings being sent back to the client. Under some circumstances, that even leads to settings being _changed_ without the user knowing.

(The responses with 875 Bytes are `colorScheme` "dark", the ones with 876 Bytes are "light")

Image

AxisRay:
#794


WantClue (commented):
Doing some testing later today

WantClue (approved):

PR #861, opened 2025-04-22
Simplify share rejection explanations

GIGIG4:
Simply share rejection explanations into a json file that contains all reasons and remove pool specific implementations.

@mutatrum is that what you head in mind?


WantClue (approved):
Way better in maintainability

PR #867, opened 2025-04-25
set GAMMATURBO_POWER_OFFSET to 10W

skot:
10W offset matches up with my power meter. 

I kinda suspect the actual TPS546 power reading is too low, and so this offset is too big, but that's a problem for another day.

### DIFF
--- a/main/power/power.c
+++ b/main/power/power.c
@@ -10,7 +10,7 @@
 
 #define SUPRA_POWER_OFFSET 5 //Watts
 #define GAMMA_POWER_OFFSET 5 //Watts
-#define GAMMATURBO_POWER_OFFSET 5 //Watts
+#define GAMMATURBO_POWER_OFFSET 10 //Watts
 
 // max power settings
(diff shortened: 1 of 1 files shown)


main/power/power.c
```diff
@@ -10,7 +10,7 @@
 
 #define SUPRA_POWER_OFFSET 5 //Watts
 #define GAMMA_POWER_OFFSET 5 //Watts
-#define GAMMATURBO_POWER_OFFSET 5 //Watts
+#define GAMMATURBO_POWER_OFFSET 10 //Watts
```
WantClue, line 13: maybe an average of 8W as an offest? Or try to use some differnt power supplies to find a good average?

WantClue (approved):

main/power/power.c
skot: I am going for power usage at the Bitaxe, ie not taking PSU efficiency into account at all. We don't have any control over what PSU people use.

PR #875, opened 2025-04-29
emc2101-fix and emc2103

### LINKED ISSUE #870: EMC2101 I2C errors should not call abort()
Several of the I2C functions in EMC2101.c are wrapped in ESP_ERROR_CHECK() which calls esp-idf abort() on failure. Theis bascially just restarts the ESP32, potentially in a loop. These I2C commands should prolly be wrapped in ESP_RETURN_ON_ERROR() with a nice error message for the log.

https://github.com/bitaxeorg/ESP-Miner/blob/e31043b8b4330063a9886cdc8931515ce053aaa1/main/thermal/EMC2101.c#L16-L28

WantClue:
changes the emc2101 error checks into `esp_return_on_error`
closes #870

### DIFF
--- a/main/thermal/EMC2101.c
+++ b/main/thermal/EMC2101.c
@@ -1,5 +1,6 @@
 #include <stdio.h>
 #include "esp_log.h"
+#include "esp_check.h"
 
 #include "i2c_bitaxe.h"
 #include "EMC2101.h"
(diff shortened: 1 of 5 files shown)


skot:
Nice, thanks for taking this on @WantClue. this is exactly what I was thinking in #870 . just a couple points;
- Can you do the same thing in EMC2103.c 🙏
- I think these errors should be cascaded up into thermal.c so that those functions also return an error.

WantClue:
check will do this

WantClue:
@skot added

skot (approved):

skot:
LGTM. Tested on 601 and 800xx

PR #880, opened 2025-05-02
Add type checks for API settings

terratec:
Check proper type for the API settings. Especially important for "fanspeed".
A JSON command `{"fanspeed": "50"}` would be converted to 0 fan speed.

Maybe the string notation is misleading in the wiki:
`curl -X PATCH http://yourbitaxe/api/system \
     -H "Content-Type: application/json" \
     -d '{"fanspeed": "desired_speed_value"}'`

### DIFF
--- a/main/http_server/http_server.c
+++ b/main/http_server/http_server.c
@@ -379,6 +379,52 @@ static esp_err_t handle_options_request(httpd_req_t * req)
     return ESP_OK;
 }
 
+bool check_json_type(int expected_type, const cJSON * const item)
+{
+    if (NULL == item) {
+        return false;
+    }
+
+    bool result = false;
+
+    if ((expected_type & cJSON_String) != 0) {
+        result = cJSON_IsString(item);
+    }
+    if (!result && (expected_type & cJSON_Number) != 0) {
+        result = cJSON_IsNumber(item);
+    }
+    if (!result && (expected_type & (cJSON_True | cJSON_False)) != 0) {
+        result = cJSON_IsBool(item);
+    }
+    if (!result && (expected_type & cJSON_True) != 0) {
+        result = cJSON_IsTrue(item);
+    }
+    if (!result && (expected_type & cJSON_False) != 0) {
+        result = cJSON_IsFalse(item);
+    }
+    if (!result && (expected_type & cJSON_Array) != 0) {
+        result = cJSON_IsArray(item);
+    }
+    if (!result && (expected_type & cJSON_Object) != 0) {
+        result = cJSON_IsObject(item);
+    }
+    if (!result && (expected_type & cJSON_Raw) != 0) {
+        result = cJSON_IsRaw(item);
+    }
+    if (!result && (expected_type & cJSON_Invalid) != 0) {
+        result = cJSON_IsInvalid(item);
+    }
+    if (!result && (expected_type & cJSON_NULL) != 0) {
+        result = cJSON_IsNull(item);
+    }
+
+    if (!result) {
+        ESP_LOGW(TAG, "Wrong API input type for setting '%s'", item->string);
+    }
+
+    return result;
+}
+
 static esp_err_t PATCH_update_settings(httpd_req_t * req)
 {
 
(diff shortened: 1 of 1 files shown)


main/http_server/http_server.c
```diff
@@ -385,6 +385,52 @@ static esp_err_t handle_options_request(httpd_req_t * req)
...
+    }
+    if (!result && (expected_type & cJSON_Invalid) != 0) {
+        result = cJSON_IsInvalid(item);
+    }
+    if (!result && (expected_type & cJSON_NULL) != 0) {
+        result = cJSON_IsNull(item);
+    }
+
+    if (!result) {
+        ESP_LOGW(TAG, "Wrong API input type for setting '%s'", item->string);
```
mutatrum: This should return HTTP 400, and ideally the whole request should be ignored. I'm not 100% sure but it looks like it'll accept the values it can, and ignore the ones that are invalid.

terratec: Yes, invalid items are ignored.
Ignoring the whole request would only work if the first item is invalid. Otherwise the whole request would have to be parsed first?

terratec: There are now two differences with the pre-check generic loop:
- "overheat_mode" can also be set to values ​​greater than 0. Is it better to remove it from the loop and check it manually?
- "coreVoltage" / "frequency": The generic pre-check protects against negative values (like #326) but the original implementation also excluded 0?
`item->valueint > 0`
Do we need an optional min_value/max_value check for each value?

mutatrum: It's tricky and I'm not sure where to go with this. If we want to do it properly, we need to check for valid values on each item. Having the SettingsFormat struct makes this more difficult. For example, the bools only allow 0/1, even though they're stored in `U16`. The `display` field only has a specific set of strings that are valid. The `rotation` field of #1022 only should have 0/90/180/270. You also already pointed out voltage/frequency, and `overheat_mode` is a special case indeed, which can only be disabled through the API, not enabled, so `0` is the only valid option here.

Maybe the more complete way would be to first indeed to a pre-check, per field. Does the field exist, is it of the right type and is the parsed value actually valid, and if so, continue. Only after all fields have been checked, the update part is checking for the existence of the field and update the NVS store. That, or have intermediate fields, which might be even cleaner actually, as you don't have to parse it again.

terratec: I think the latest changes address your suggestions.

- check for valid type, valid data type range and value range for each item using the generic approach
- check individual values or strings manually if necessary
- parse only once with cached values
- all errors of the request are logged and not just one

main/http_server/http_server.c
```diff
@@ -411,9 +430,231 @@ static esp_err_t handle_options_request(httpd_req_t * req)
...
+                        }
+                        break;
+                    case NVS_TYPE_I16:
+                        if ((SHRT_MIN > item->valueint) || (SHRT_MAX < item->valueint)) {
+                            ESP_LOGW(TAG, "Value '%d' for '%s' is not compatible with NVS_TYPE_I16", item->valueint, settings[i].name);
+                            result = false;
+                        }
+                        break;
+                    case NVS_TYPE_I32:
+                        break;
```
mutatrum, line 664: Missed this one?

main/http_server/http_server.c
```diff
@@ -411,9 +430,231 @@ static esp_err_t handle_options_request(httpd_req_t * req)
...
+                    result = false;
+                    break;
+            }
+        }
+    }
+
+    // update NVS (if result is okay) and clean up
+    for (int i = 0; i < ARRAY_SIZE(settings); i++) {
+        if (settings[i].updated) {
+            if (result) {
```
mutatrum, line 737: Why not have the for loop inside the if (result)? Can't we skip the whole loop if result == false?

main/http_server/http_server.c
terratec: NVS_TYPE_I32 is a valid case but the type range is the same.
Maybe I should add a check in case the json type `item->valueint` changes? Or at least a comment?

main/http_server/http_server.c
terratec: For now the loop is crucial to free the string buffers (if used).

WantClue:
@terratec please resolve the conflict

WantClue:
@terratec please resolve conflicts again :D sorry

WantClue:
gonna merge this afterwards

Issue #879, opened 2025-05-01
Text wrapping of User address broken


NilByte:
Text wrapping of User address is broken in Dashboard Pool Information.

![Image](https://github.com/user-attachments/assets/996c4c8b-8107-4c79-8333-0090185763fc)

skot:
thanks for the report!

It would be cool to shorten the display of long usernames, because there are likely BTC addresses. Privacy and UI at the same time. Maybe we could show the full username on hover or something.

PR #886, opened 2025-05-03
fix: substring username on dashboard

mrv777:
Fix for issue #879


mrv777:
Went with a substring that is shorten at the beginning and longer at the end with a tooltip, I liked this solution since usually a worker is added at the end and better chance its not cut off if the end has more character

image

duckAxe:
Another (simpler) way to fix this issue would be to use [text-overflow: ellipsis](https://www.w3schools.com/cssref/css3_pr_text-overflow.php)

duckAxe:
> yeah, we could just do that, but think it may be helpful to see some of the end and try to show the worker name

Do you have a screenshot of what your change looks on desktop? Sorry, can't test your commit, don't have BitAxe with AxeOS.

duckAxe:
@mrv777 See my implementation for Nerd*Axe https://github.com/shufps/ESP-Miner-NerdQAxePlus/pull/158

mutatrum:
This is a neat solution. But I would implement it slightly differently, by finding the first `.` character, and if it's there, split it 6 characters before that. This way it will always show the username, which is the most important part and you don't have to guess the cut-off point.

Secondly, I would add a tooltip showing the full field.

duckAxe:
> split it 6 characters before that.

In my implementation for Nerd*Axe I used 20 characters from the end. I don't think we should check for `.` just because it's possible to have a looooong username without a worker. Maybe we shouldn't overshoot, it's just a string output ;)

> Secondly, I would add a tooltip showing the full field.

Of course. I use the browser native `title` attribute for that because I never saw a tooltip on Nerd*Axe UI.

duckAxe:
PR https://github.com/bitaxeorg/ESP-Miner/pull/920

WantClue:
this will be closed in favour of #920

PR #888, opened 2025-05-03
Show highest temperatures since system boot

terratec:


main/http_server/axe-os/src/app/components/home/home.component.html
```diff
@@ -259,8 +259,20 @@ <h5>Pool Information ({{ activePoolLabel }})</h5>
...
+                <table>
+                    <tr>
+                        <td>{{info.uptimeSeconds | dateAgo}}</td>
+                    </tr>
+                    <tr>
+                        <td>Highest ASIC Temperature:</td>
+                        <td>{{info.tempHighest}}&deg;C</td>
+                    </tr>
+                    <tr>
+                        <td>Highest VR Temperature:</td>
```
mutatrum, line 272: Not all devices have VR temperature sensors. This should be conditional.

main/http_server/openapi.yaml
```diff
@@ -129,10 +129,12 @@ components:
         - stratumURL
         - stratumUser
         - temp
+        - tempHighest
```
mutatrum, line 132: Style thing, but I don't really like `Highest`. `tempMax` maybe? Same for the vrTemp one.

main/tasks/power_management_task.c
```diff
@@ -75,8 +77,14 @@ void POWER_MANAGEMENT_task(void * pvParameters)
 
         power_management->fan_rpm = Thermal_get_fan_speed(GLOBAL_STATE->device_model);
         power_management->chip_temp_avg = Thermal_get_chip_temp(GLOBAL_STATE);
+        if (power_management->chip_temp_avg > power_management->chip_temp_highest) {
+            power_management->chip_temp_highest = power_management->chip_temp_avg;
+        }
 
         power_management->vr_temp = Power_get_vreg_temp(GLOBAL_STATE);
+        if (power_management->vr_temp > power_management->vr_temp_highest) {
```
mutatrum, line 85: Same here about conditionality. Not sure how it's done in other parts.

mutatrum (changes requested):

main/http_server/openapi.yaml
terratec: There are already `maxTemp` and `maxPower` which refer to the chip's specified maximum value. Could this be confusing?

mutatrum: Ok, I'll open an issue to unconfuse them. Feels out of scope for this PR.

duckAxe:
@terratec Could you please confirm whether we need this functionality? With the new PID temperature control, you set the target temperature, which is then maintained constantly. Therefore, the temperature will always be constant and never change. I have several devices, and the temperature has remained constant since the restart.

terratec:
> @terratec Could you please confirm whether we need this functionality? With the new PID temperature control, you set the target temperature, which is then maintained constantly. Therefore, the temperature will always be constant and never change. I have several devices, and the temperature has remained constant since the restart.

I think this can be useful for manual fan control setting or for hot environments (100% fan speed, either PID or manual)?

duckaxe:
@terratec I pushed a commit to my local branch to fix the uptime table https://github.com/terratec/ESP-Miner/commit/d1372c9c3aa9e1423b2cdfd49ee2ea5389cd2903

Your change is good, but it's better. `colspan=2` can be removed

**Yours**
Screenshot 2025-06-19 at 09 10 08

**Mine**
Screenshot 2025-06-19 at 09 07 41

duckaxe:
@terratec pls solve merge conflicts. there is also an open requested change.

terratec:
> @terratec pls solve merge conflicts. there is also an open requested change.

Rebased and fixed the visibility for max VR temperature.
This seems to be an orphaned notice? As far as I can see, I have resolved all change requests

duckaxe:
@terratec According to https://github.com/bitaxeorg/ESP-Miner/pull/1077 I adjusted the Uptime card. Can you please apply the [Card.patch](https://github.com/user-attachments/files/20932122/Card.patch). After that both cards should look similar.

Screenshot 2025-06-26 at 21 29 15

duckaxe:
@terratec How do we want to proceed with this PR?

terratec:
> @terratec How do we want to proceed with this PR?

Perhaps it should also be moved to the system tab?

duckaxe:
> > @terratec How do we want to proceed with this PR?
> 
> Perhaps it should also be moved to the system tab?

That's fine with me. @mutatrum any opinions?

mutatrum:
Makes sense, fits in with the uptime statistics.

PR #896, opened 2025-05-05
hotfix fan controller, increase minimum and add ap mode fan speed

WantClue:
Hotfix Fan Controller.

This changes the minimum fan speed for the PID to 25% as well as ensure the fan is running in the AP mode at 70% to prevent heat buildup.


sz4bi:
15% was fine for my 601, without heavy cooling mods, it kept 60 C at 15% fanspeed. Now it's loud even at the lowest PID speed.

PR #899, opened 2025-05-06
Fix in a race condition, stratum_task run ahead of asic_task will causing cleanQueue function accessing a NULL address

aaron3481:
Fix in a race condition, stratum_task run ahead of asic_task will causing cleanQueue function accessing a NULL address and hence halt the system.


main/tasks/asic_task.c
```diff
@@ -28,6 +28,8 @@ void ASIC_task(void *pvParameters)
         GLOBAL_STATE->valid_jobs[i] = 0;
     }
 
+    GLOBAL_STATE->job_queue_initalized = true;
+
     ESP_LOGI(TAG, "ASIC Job Interval: %.2f ms", GLOBAL_STATE->asic_job_frequency_ms);
     SYSTEM_notify_mining_started(GLOBAL_STATE);
```
mutatrum, line 34: No need for a new boolean, you can leverage `duration_start` which is set in `SYSTEM_notify_mining_started`.

mutatrum (changes requested):

mutatrum:
We had to force-push a different history on `dev-latest`, could you rebase your commit on top of the current dev-latest branch? Sorry about this.

PR #910, opened 2025-05-09
Message update with ref to new firmware release.

STSMiner1:
This is to inform the user that both files are required when updating the firmware on their device.

Changed to ....

Clicking this button will connect to GitHub to check for the latest release, both files are required.

image


WantClue (approved):
Simple and hopefully effective

PR #913, opened 2025-05-12
Send mining.suggest_difficulty after authorization

mutatrum:
Fixes finding from https://github.com/bitaxeorg/ESP-Miner/discussions/902

It also refactor `set_difficulty` passing so `mining_notify` only actually received fields and accept a new difficulty in the same way as accepting a new version_mask in `create_jobs_task.c`.


skot:
Thanks for sorting this out. I am very weary of this stratum init sequence. We’ve had so many pool-specific issues here. This needs to be tested across lots of different pools 😭

Issue #722, opened 2025-02-19
Add expand button for the Log window


Bitcoinite:
Even when having a big screen, the Log window is not scaling at all.
We could add a expand button to it to make use of the entire length of the browser window.
This would really help monitoring the live log.

PR #919, opened 2025-05-14
Feature: Expand Logs

duckAxe:
To allow the user to monitor logs in full screen size, an additional button has been added to display the log view in full size.

- [x] Additional button `Expand Logs` added top expand the logs view
- [x] `Close` button will close the expanded view

Screenshot 2025-05-14 at 14 57 05

Closes https://github.com/bitaxeorg/ESP-Miner/issues/722


mutatrum:
Would it make sense to make the expanded logs default, without the button?

And how is this on mobile?

duckAxe:
The user should decide for himself whether he wants to see the logs in full screen. And the best way to do this is with a button.

Mobile is fullscreen too.

mutatrum:
I like the functionality, I just don't think it should be a button like this. It's not an 'action' as such, it's a UI toggle.

Not sure if it's too tricky, but maybe it can be done with a maximize icon on the top right of the console window or something like that?

Silly quick-and-dirty mockup:

image

mutatrum (approved):

mutatrum:
LGTM, love the demo video's on all the high quality PRs, thank you!

PR #925, opened 2025-05-15
Fix field label vertical alignment

duckAxe:
On the settings page, the last field label is not vertically centered because the field has an extra hint line compared to the other two fields. PR fixes the alignment of all field labels.

**Before**

**After**


mutatrum (approved):
Sharp eye!

duckAxe:
Blocked by https://github.com/bitaxeorg/ESP-Miner/issues/941

mutatrum:
Might as well just fix it in #942.

duckAxe:
Closed because no longed needed. Replaced by https://github.com/bitaxeorg/ESP-Miner/issues/941

PR #926, opened 2025-05-16
Move restart warning to Toast

duckAxe:
The hint `You must restart this device after saving for changes to take effect` is currently right next to the save button. That does not look good. It would be much better if this warning only appeared when needed. For example, when saving.

My idea is to display the warning as a [Toast](https://primeng.org/toast) notification. This way, the user would immediately notice the warning when saving.

This affects 3 settings pages: Pool, Network, (Miner) Settings.

**Before**
localhost_4200_

**After**
localhost_4200_ (2) (1) 11


main/http_server/axe-os/src/app/components/edit/edit.component.html
```diff
@@ -59,15 +59,15 @@
             </div>
         </div>
 
-        <div class="col-12 md:col-4">
+        <div class="mt-2">
```
mutatrum, line 62: Be cautions with fixes in files that are otherwise not touched by a PR. This makes reviewing more difficult, as it's less clear what's part of the changes, and what are unrelated changes.

It's fine for this one, as it's limited, but more of a general remark.

mutatrum (approved):
LGTM, makes a lot of sense to move this to toasters. I assume this is also fine on mobile?

There are some other banner type warnings as well, should they all be moved to toasters?

main/http_server/axe-os/src/app/components/edit/edit.component.html
duckAxe: 👍

mutatrum:
Merges, created #944 for possible follow-up.

WantClue:
You introduced a misbehavior with this PR, the "you must restart your device for changes to take effect" is now being shown every single time. Which is incorrect. This is only needed on certain changes and not always @duckAxe would be nice if you can tackle this?

duckAxe:
@WantClue Have you tested my changes? The message only pops up when you save something and only on the pages I mentioned.

WantClue:
> @WantClue Have you tested my changes? The message only pops up when you save something and only on the pages I mentioned. Check also my code changes on .ts files. There is always a condition around the message. And can you pls provide a screenshot or vid.

@duckAxe Yes I've tested them, the provided vid e.g. shows the toast on the freq change as an example. Which does change on the fly and doesn't req a reboot

https://github.com/user-attachments/assets/00b928fa-fe7c-4761-be15-64e35851a46b

duckAxe:
@WantClue Do you mean something else? Are you saying that a restart isn't required for every field change? Do I understand correctly? If so, that wasn't the case before I made the change. The message "You must restart ..." was always visible, making it clear to the user that a restart was necessary for every adjustment.

If we want to show the message only for certain settings, we can do so in a new PR. In order to do so, I need to know which settings require a restart and which do not.

PR #927, opened 2025-05-16
Add option to disable or enable mining via API and UI

GIGIG4:
This PR adds the option to disable and enable the mining without powering of the whole ESP.

Add two endpoints `/api/mining/stop` and `/api/mining/start` as well as a card to the home.component.ts:

grafik
grafik

When mining is resumed, it hovers at roughly 800 GH/s for a minute or two before going up to normal levels again.

Tested on my 601 for a week with multiple starts and stops.

### DIFF
--- a/main/mining_controller.c
+++ b/main/mining_controller.c
@@ -1,0 +1,224 @@
+#include "mining_controller.h"
+#include "esp_log.h"
+#include "esp_system.h"
+#include "freertos/semphr.h"
+#include <stdlib.h>
+
+#include "asic_result_task.h"
+#include "asic_task.h"
+#include "create_jobs_task.h"
+#include "main.h"
+#include "power_management_task.h"
+#include "stratum_task.h"
+
+#include "asic.h"
+#include "power.h"
+#include "serial.h"
+#include "vcore.h"
+#include "work_queue.h"
+
+#include "nvs_config.h"
+
+static const char * TAG = "mining_controller";
+
+esp_err_t start_mining(GlobalState * global_state)
+{
+    ESP_LOGI(TAG, "Starting mining operations...");
+
+    if (global_state->mining_enabled) {
+        ESP_LOGW(TAG, "Mining is already enabled.");
+        return ESP_OK;
+    }
+
+    BaseType_t task_created;
+    task_created = xTaskCreate(POWER_MANAGEMENT_task, "power management", 8192, (void *) global_state, 10,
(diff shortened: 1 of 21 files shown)


duckAxe:
Would it be a good idea to put the stop/start button on the dashboard? I think, no. This button isn't something anyone needs every day. I would put the button in the settings. After all, it's just a setting to turn the miner on or off, so it's not important to put it on the dashboard.

main/mining_controller.c
```diff
@@ -0,0 +1,253 @@
...
+#include "asic.h"
+#include "serial.h"
+#include "vcore.h"
+#include "work_queue.h"
+
+#include "nvs_config.h"
+
+static const char * TAG = "mining_controller";
+
+esp_err_t start_mining(GlobalState * global_state)
```
mutatrum, line 23: A general comment, it looks like this class has too much responsibility. Ideally, it should only toggle the flag, and start tasks, and the tasks themselves should handle all responsibilities. F.e. retrieving frequency and logging about it, that should be done inside POWER_MANAGEMENT_task.

Secondary remark: it looks like some stuff is duplicated from `main.c`. Ideally, the cold boot should use the exact same startup sequence and code as a restart.

main/tasks/stratum_task.c
```diff
@@ -344,6 +416,15 @@ void stratum_task(void * pvParameters)
             } else if (stratum_api_v1_message.method == CLIENT_RECONNECT) {
                 ESP_LOGE(TAG, "Pool requested client reconnect...");
                 stratum_close_connection(GLOBAL_STATE);
+                if (!GLOBAL_STATE->mining_enabled) {
```
mutatrum, line 419: Why is this check and the stop block so many times in `stratum_task.c`? Is it possible to have one stop block per task? I see it logs a different stop reason, but in all cases an error is already logged, so having a specific stop reason doesn't add a lot.

mutatrum (commented):
I really like this functionality, but I think it can be achieved with quite a bit less code. See comments.

mutatrum (changes requested):
See comments

WantClue:
@GIGIG4 please resolve the conflict so we can proceed with reviewing it.

mutatrum:
A lot has changed in the code-base, and this PR is too far removed from the current state, unfortunately. It's still a good feature to have, but it would make more sense to start with a clean slate. Linked issue stays open and is stays connected, so we can use this as reference if needed.

Issue #915, opened 2025-05-12
Sorting on swarm page is not persisted


bredita:
**Describe the bug**
Sorting miners by hostname works, but inly unitil the next automatic refresh when it defaults back to sorting by IP.

**To Reproduce**
1. Go to 'swarm page' (i see my 5 gammas there)
2. Click on 'hostname', notice the sort order
3. wait until next refresh
4. List is now sorted by IP again.

**Expected behavior**
I expect it to maintain whatever sort order I have chosen.

**Hardware (please complete the following information):**
 - Bitaxe HW version: gamma 601
 - ESP-Miner FW version: 2.7.1

bredita:
fixed in v2.8.0b2.
thank you.

PR #928, opened 2025-05-17
Fix swarm sorting

duckAxe:
This PR solves the issue of lost sorting after manual and automatic swarm list refresh. Now the changed code remembers the user's sorting preference and applies them each time the swarm is updated. Even after (swarm) page reload.

Fixes https://github.com/bitaxeorg/ESP-Miner/issues/915

### DIFF
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
@@ -7,6 +7,8 @@ import { LocalStorageService } from 'src/app/local-storage.service';
 import { SystemService } from 'src/app/services/system.service';
 const SWARM_DATA = 'SWARM_DATA'
 const SWARM_REFRESH_TIME = 'SWARM_REFRESH_TIME';
+const SWARM_SORTING = 'SWARM_SORTING'
+
 @Component({
   selector: 'app-swarm',
   templateUrl: './swarm.component.html',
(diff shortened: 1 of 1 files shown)


jddebug:
It would be great if the swarm was numbered or at least if at the top it showed how many active miners are in the swarm. It's nice to be able to order them, but being able to at a glance know that all of your miners are still mining would be handy. Currently, I have to count them one by one and make sure they're all there.

duckAxe:
@jddebug That's a nice idea, but it's not part of this fix / PR. Feel free to create an issue for it.

mutatrum:
Thank you for the PR. #787 also addresses this, we need to see which is the best fit.

duckAxe:
> Thank you for the PR. #787 also addresses this, we need to see which is the best fit.

With all due respect to @dustinb, I prefer my change because it retains the sorting even after a page reload. This way, I won't have to set my preferred sorting when I reopen the browser (tab).

duckAxe:
@mutatrum Why hasn't the PR https://github.com/bitaxeorg/ESP-Miner/pull/787 already been merged? Because the review from @mrv777 is missing? I'm trying to figure out how the approval of pull requests works. Why are they not merged much faster if everything is right?

duckAxe:
@mutatrum Thank you. I get it. 

> You're more than welcome into Discord, if you're not there yet, we would love to discuss it. That would be a better channel than here

I was banned from the osmu server because I posted an external link. Not my choice.

benjamin-wilson:
> [duckAxe](/duckAxe) If you give me your username I can review it. There is a bot that automatically bans new users from posting links are we are often attacked by spam.

WantClue (approved):
LGTM nice one

PR #929, opened 2025-05-17
Remember menu toggle state

duckAxe:
This change saves the current menu toggle status in the local storage so that the status is retained after page reload. If the user has already hidden the menu, he does not need to do so again. Desktop only.

https://github.com/user-attachments/assets/d5a1059f-9e1c-42ba-a480-22b2341071b9


main/http_server/axe-os/src/app/layout/service/app.layout.service.ts
```diff
@@ -156,6 +163,8 @@ export class LayoutService {
         if (this.isDesktop()) {
             this.state.staticMenuDesktopInactive =
                 !this.state.staticMenuDesktopInactive;
+
+            this.localStorageService.setBool('STATIC_MENU_DESKTOP_INACTIVE', this.state.staticMenuDesktopInactive);
```
mutatrum, line 167: Is this using the constant or redefining a string? Same on L232

mutatrum (changes requested):
LGTM, except the constant comment.

WantClue (approved):

mutatrum (approved):

PR #931, opened 2025-05-18
Fix broken text scroll animation for self test

terratec:
This stops serveral calls to `lv_label_set_text` and initializes a few other variables.

### DIFF
--- a/main/screen.c
+++ b/main/screen.c
@@ -35,8 +35,8 @@ static lv_obj_t * screens[MAX_SCREENS];
 static int delays_ms[MAX_SCREENS] = {0, 0, 0, 0, 0, 1000, 3000, 3000, 10000, 10000};
 
 static screen_t current_screen = -1;
-static int current_screen_time_ms;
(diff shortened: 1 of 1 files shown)


main/screen.c
```diff
@@ -35,8 +35,8 @@ static lv_obj_t * screens[MAX_SCREENS];
 static int delays_ms[MAX_SCREENS] = {0, 0, 0, 0, 0, 1000, 3000, 3000, 10000, 10000};
 
 static screen_t current_screen = -1;
-static int current_screen_time_ms;
-static int current_screen_delay_ms;
+static int current_screen_time_ms = 0;
```
mutatrum, line 38: Not sure what the general style is, but IIRC these initialisers are not necessary for static fields.

mutatrum (approved):
LGTM, only one small question

main/screen.c
terratec: Ah yes it is specified

PR #937, opened 2025-05-20
Ssd1309 extra screen

WantClue:
builds ontop of #934 
adds another screen depending on the resolution that's available to display wifi rssi.

To-Do:
- [x] Add Uptime
- [x] Update RSSI constantly
- [ ] Add RSSI strength indicator


main/screen.c
```diff
@@ -403,6 +436,45 @@ static void screen_update_cb(lv_timer_t * timer)
...
+        } else {
+            snprintf(rssi_buf, sizeof(rssi_buf), "RSSI: -- dBm");
+        }
+        if (strcmp(lv_label_get_text(wifi_rssi_value_label), rssi_buf) != 0) {
+            lv_label_set_text(wifi_rssi_value_label, rssi_buf);
+        }
+    }
+
+    if (esp_uptime_label) {
+        char uptime[50];
```
mutatrum, line 453: This whole thing gets called on every `screen_update_cb`, which is a lot. It might make sense to do this in a timer that runs once every second and set a char * field in SYSTEM_MODULE.

main/screen.c
```diff
@@ -403,6 +436,45 @@ static void screen_update_cb(lv_timer_t * timer)
         lv_label_set_text_fmt(chip_temp_label, "Temp: %.1f C", power_management->chip_temp_avg);
     }
 
+    if (wifi_rssi_value_label) {
+        char rssi_buf[25];
+        
+        if (module->wifi_rssi < 0 && module->wifi_rssi >= -127) { // Typical RSSI range
```
mutatrum, line 442: ESP docs say it can actually sometimes be positive, for some reason?

main/screen.c
```diff
@@ -403,6 +436,45 @@ static void screen_update_cb(lv_timer_t * timer)
         lv_label_set_text_fmt(chip_temp_label, "Temp: %.1f C", power_management->chip_temp_avg);
     }
 
+    if (wifi_rssi_value_label) {
```
mutatrum, line 439: Assume label is there, we don't check any of the others.

main/screen.c
```diff
@@ -295,6 +320,14 @@ static void screen_update_cb(lv_timer_t * timer)
             display_on(true);
         }
     }
+    
+    // Update WiFi RSSI periodically
+    if (GLOBAL_STATE->SYSTEM_MODULE.is_connected) {
+        int8_t rssi_value;
+        if (get_wifi_current_rssi(&rssi_value) == ESP_OK) {
+            GLOBAL_STATE->SYSTEM_MODULE.wifi_rssi = rssi_value;
```
mutatrum, line 328: This is in the same function as where it's used. I think the `wifi_rssi` field can be removed from SYSTEM_MODULE, and the call can be done further down.

main/screen.c
```diff
@@ -417,7 +489,19 @@ static void screen_update_cb(lv_timer_t * timer)
...
+
+    // Loop to find the next screen that should be displayed
+    do {
+        next_scr++; // Advance to the next screen candidate
+        if (next_scr > SCR_CAROUSEL_END) { // If past the end of carousel
+            next_scr = SCR_CAROUSEL_START; // Wrap around to the start of carousel
+        }
+        // If the candidate screen is SCR_WIFI_RSSI AND this is NOT a bigger display,
+        // then this screen should be skipped, and the loop will continue to find the next one.
+    } while (next_scr == SCR_WIFI_RSSI && GLOBAL_STATE->DISPLAY_CONFIG.v_res / 8 != 8);
```
mutatrum, line 502: We have to think about this a bit more. For now I would revert this bit, and just add the new screen for all displays. If we want to have different screen designs for different resolutions, it would make sense to do that in `screen_start`, only once.

mutatrum: F.e. we could make 2 separate designs for the Wi-Fi, and either add one or the other on the SCR_WIFI_RSSI position. Or have a completely dynamic bit of the carousel, that a small screen has 4 screens (as an example), but a larger screen only has 3, with different designs.

PR #898, opened 2025-05-05
Statistics task for previous chart data

terratec:
with new API endpoint for statistics.
The configuration with 0 data points is live data only (old behavior) with statistics disabled.
The poll rate is automatically set based on the configured duration.

PR #940, opened 2025-05-21
FUP #898: Fix styling for statistics settings

duckAxe:
This change improves the layout of the statistics settings added by https://github.com/bitaxeorg/ESP-Miner/pull/898. Headlines spacing has also been adjusted.

### Before
Screenshot 2025-05-21 at 22 15 29

### After
Screenshot 2025-05-21 at 22 42 24

### Responsiveness

https://github.com/user-attachments/assets/7f796d4d-e60f-43c6-bd89-3fe690592c91


mutatrum:
Looking at the screens, it's a bit weird to have two sliders next to each other in the statistics part. Would it make more sense to have them below each other?

duckAxe:
Placing sliders next to each other saves space. However, I have placed them one below the other.

Regarding your suggestion to make the label a separate column: For now, I won't do that so that all sliders on the page look the same as the temperature slider. Once we have merged the [Display Slider](https://github.com/bitaxeorg/ESP-Miner/pull/942), we can make all the sliders the same and add a label column. 

Screenshot 2025-05-23 at 13 30 22

VortexRadar:
On Desktop, it can be a bit confusing to adjust the Statistics Limit slider since it doesn't allow you to make small incremental changes the way you can with the Statistics Duration slider. Unless you try moving the slider more, it may seem like it's broken or not working. This is less of an issue on mobile.

duckAxe:
> On Desktop, it can be a bit confusing to adjust the Statistics Limit slider since it doesn't allow you to make small incremental changes the way you can with the Statistics Duration slider. Unless you try moving the slider more, it may seem like it's broken or not working. This is less of an issue on mobile.

Yes, the limit slider steps are set to 60. @terratec can we reduce the steps?

P.S. with my [initial commit](https://github.com/bitaxeorg/ESP-Miner/pull/940#issue-3081370292) we wouldn't have this problem, because the slider would be half as wide.

mutatrum:
Repeating my comments from [Discord](https://discord.com/channels/1091348375301013615/1094385611718270977/1376919820196384859) here:

> I wouldn't know when you would limit the number of data points. And the other one might be changed into frequency, that would make more sense IMO. And similar to the screensaver timeout, it doesn't need to be a linear scale.

The sliders implicitly control the frequency that would be easier if that would be explicit. Total time is a result of that, with the total number of points.

duckAxe:
> The sliders implicitly control the frequency that would be easier if that would be explicit. Total time is a result of that, with the total number of points.

That means a weak UI concept for the whole stats functionality? Then we can also close the PR. Let's create an issue and bring all thoughts together.

terratec:
> Yes, the limit slider steps are set to 60. @terratec can we reduce the steps?

The steps can be reduced. There is a bug in the frequency calculation for values with fractions anyway :(

terratec:
> Repeating my comments from [Discord](https://discord.com/channels/1091348375301013615/1094385611718270977/1376919820196384859) here:
> 
> > I wouldn't know when you would limit the number of data points. And the other one might be changed into frequency, that would make more sense IMO. And similar to the screensaver timeout, it doesn't need to be a linear scale.
> 
> The sliders implicitly control the frequency that would be easier if that would be explicit. Total time is a result of that, with the total number of points.

I wasn't sure if 720 data points would be stable/fast enough with the json api as a bottleneck.
With two sliders it was possible to set a short duration with low resolution.

Maybe we can use a single frequency slider with exponential values and show an expected value for the duration? (Or vice versa with a duration slider?)

PR #942, opened 2025-05-22
Display Timeout Slider

duckAxe:
This PR replaces the display timeout input field with a slider. Since sliders can only work with ascending numbers (0 ... x ... -1 isn't possible), I use a hidden input field and synchronise the modified data between the slider and the input field.

The maximum adjustable value: 30 minutes. I can change this value if necessary. I consider this value to be optimal.

The layout is also optimised for mobile.

https://github.com/user-attachments/assets/2d9cae1b-a2ad-424c-bdc0-c4a32005be22

### DIFF
--- a/main/http_server/axe-os/src/app/components/edit/edit.component.html
+++ b/main/http_server/axe-os/src/app/components/edit/edit.component.html
@@ -86,6 +86,15 @@
             </div>
         </div>
 
+        <div class="field grid p-fluid">
+            <label htmlFor="displayTimeoutControl" class="col-12 md:col-2 hidden md:flex mb-2 md:mb-0">Display Sleep</label>
+            <div class="col-12 md:col-10">
+                <label class="block mb-2">{{getDisplayTimeoutFormated()}}</label>
+                <p-slider [min]="getDisplayTimeoutMin()" [max]="getDisplayTimeoutMax()" [step]="1" [formControl]="displayTimeoutControl"></p-slider>
+                <input type="hidden" formControlName="displayTimeout" />
+            </div>
+        </div>
+
         <div class="col-12 md:col-4">
             <div class="field-checkbox">
                 <p-checkbox name="flipscreen" formControlName="flipscreen" inputId="flipscreen"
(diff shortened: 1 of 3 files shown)


STSMiner1:
@duckAxe 

Found an issue.

Set timer for 1 minute, display turns off - all good there.

Wait 10 mins plus and try to change the slider back to 0 - the Save button does not activate to save the new setting.

duckAxe:
@STSMiner1 Good find. Fixed!

mutatrum:
On the original issue (#525) there was a lot of discussion on this. I also experimented with a slider, but your solution works better, what I did was cumbersome.

However, with a linear scale, there's no practical use to choose between 28, 29 or 30 minutes. I would opt for a longer maximum time, and to make that possible, have the values in an exponential scale. F.e. 1, 2, 5, 15, 30 minutes and then 1, 2, 4, 8 hours or something like that. I tried it, but with your solution this would work better. Back-end should still be able to accept all the values in minutes, so it's front-end only.

duckAxe:
> However, with a linear scale, there's no practical use to choose between 28, 29 or 30 minutes. I would opt for a longer maximum time, and to make that possible, have the values in an exponential scale. F.e. 1, 2, 5, 15, 30 minutes and then 1, 2, 4, 8 hours or something like that. I tried it, but with your solution this would work better. Back-end should still be able to accept all the values in minutes, so it's front-end only.

It is possible to work with fixed values like `1m, 2m, 5m, 1h, 2h` and so on. The problem is to map already stored values on such an exponential scale. For example, how do you want to display `21 minutes` that a user has already saved?

mutatrum:
> It is possible to work with fixed values like `1m, 2m, 5m, 1h, 2h` and so on. The problem is to map already stored values on such an exponential scale. For example, how do you want to display `21 minutes` that a user has already saved?

Simplest one would be to just round down, or if it's an invalid value, use default. The screensaver is a relatively new feature, and this falls in the same category as above: 21 minutes is so specific, so either pick 15m or 30m.

duckAxe:
> Simplest one would be to just round down, or if it's an invalid value, use default. The screensaver is a relatively new feature,

That is not good. We are showing the user something that is not right. The real screen timeout time is different from the time we show in the GUI.

Let me try one thing. I have an idea.

main/http_server/axe-os/src/app/components/edit/edit.component.html
```diff
@@ -86,6 +86,27 @@ <h3>Display</h3>
...
+            <label htmlFor="displayTimeoutControl" class="col-12 md:col-2 hidden md:flex mb-2 md:mb-0 relative" style="top: -5px;">
+                Display Sleep
+            </label>
+            <div class="col-12 md:col-10">
+                <label class="block mb-2">
+                    <ng-container *ngIf="this.form.controls['displayTimeout'].value == 0">
+                        Display always off
+                    </ng-container>
+                    <ng-container *ngIf="this.form.controls['displayTimeout'].value > 0">
+                        Display timeout: {{this.form.controls['displayTimeout'].value * 60 | dateAgo}}
```
mutatrum, line 99: Clever use of `dateAgo`.

mutatrum:
Is your elegant solution already in the code?

duckAxe:
> Is your elegant solution already in the code?

Is this elegant solution https://github.com/bitaxeorg/ESP-Miner/pull/942#issuecomment-2909597624 not elegant enough? :)

duckAxe:
@WantClue Conflicts solved.
Props to @mutatrum for the exponential scale idea.

PR #948, opened 2025-05-23
Add sorting to all swarm table columns

duckAxe:
This PR adds sorting to all columns in the swarm table. This allows devices to be sorted by more than just IP or host name. The feature is particularly useful when working with a large amounts of miners.

Also changed:

- [x] Order of the fields `Shares` and `Best Diff` (more to the left because more important)
- [x] Added missing tooltips to the fields `Best Diff` and `Temp`

https://github.com/user-attachments/assets/7b6b913d-976d-41c6-8896-be5828e0fc90


main/http_server/axe-os/src/app/components/swarm/swarm.component.html
```diff
@@ -62,33 +62,53 @@
...
+                    name: 'IP',
+                    label: 'IP'
+                },
+                {
+                    name: 'hostname',
+                    label: 'Hostname'
+                },
+                {
+                    name: 'hashRate',
+                    label: 'Hash Rate'
```
mutatrum, line 76: Had to look this up, didn't know there's supposed to be a space between hash and rate.

PR #955, opened 2025-05-25
Chart data source selection

terratec:
All data sources can be selected on-the-fly for the chart (y or y2 axis).

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.ts
+++ b/main/http_server/axe-os/src/app/components/home/home.component.ts
@@ -425,7 +425,7 @@ export class HomeComponent implements OnInit, OnDestroy {
           [
             this.pageDefaultTitle,
             info.hostname,
-            (info.hashRate ? HashSuffixPipe.transform(info.hashRate * 1000000000) : false),
+            (info.hashRate ? HashSuffixPipe.transform(info.hashRate) : false),
             (info.temp ? `${info.temp}${info.temp2 > -1 ? `/${info.temp2}` : ''}${info.vrTemp ? `/${info.vrTemp}` : ''} °C` : false),
             (!info.power_fault ? `${info.power} W` : false),
             (info.bestDiff ? info.bestDiff : false),


mutatrum:
Can you add some screenshots?

terratec:
Fan speed and rpm (manual and PID):
fan

terratec:
Power, current and voltages:
power

voltage

Maybe the `suggestedMax` can be improved.

terratec:
heap+wifi

mutatrum:
> Settings with source selection: settings

On initial though, I think the axis selection should be on the dashboard, near the graph, not in settings. Not sure how that should be UI wise.

Secondly, someone on Discord is working in the same area, but has a different direction: https://discord.com/channels/1091348375301013615/1094385595981246535/1375742408045760595

skot:
This is awesome! But I agree the source selector should be in the chart page.

mutatrum:
One top left and the other top right, above their respective axes? No need to have the labels then.

duckAxe:
Ah, I dont see, you uploaded a mockup ;)
I have scaled the dropdown a little, it doesn't have to be big on the dashboard.

localhost_4200_

duckAxe:
@terratec I added my changes to the two dropdowns on the dashboard. It works fine. Just apply the [attached patch file](https://github.com/user-attachments/files/20524846/Untitled.patch) to your code. What we need is to reset the chart dataset for the changed axis. See my video at the end.

https://github.com/user-attachments/assets/a1ddbe26-3fb6-4f94-a570-2c3a88cb5867

terratec:
This is fantastic, thank you!

Maybe we can call `initializeChart` to reset? Do we have to deinit/unsubscribe first?

duckAxe:
@terratec I tested your latest changes, it looks like the view is reloading, but the chart is still broken.

https://github.com/user-attachments/assets/5f199acf-b14c-41a4-b13d-4a17ebdc2b39

duckAxe:
@terratec I changed something and it looks much better without re-rendering the view. [Untitled.patch](https://github.com/user-attachments/files/20535015/Untitled.patch)

https://github.com/user-attachments/assets/c78c15d3-d3b1-4684-9c1b-06f53648419b

duckAxe:
Ah, you right. Okay, then we'll wait for the merge.

main/tasks/statistics_task.h
```diff
@@ -7,17 +7,17 @@ typedef struct StatisticsData * StatisticsNextNodePtr;
...
     float vrTemperature;
     float power;
     float voltage;
     float current;
-    int16_t coreVoltageActual;
-    uint16_t fanSpeed;
-    uint16_t fanRPM;
-    int8_t wifiRSSI;
-    uint32_t freeHeap;
+    float coreVoltageActual;
```
KillerInk: Old types are fine. They dont have to be a float.

main/tasks/statistics_task.c
```diff
@@ -127,17 +127,17 @@ void statistics_task(void * pvParameters)
...
-            statsData.current = Power_get_current(GLOBAL_STATE);
-            statsData.coreVoltageActual = VCORE_get_voltage_mv(GLOBAL_STATE);
-            statsData.fanSpeed = power_management->fan_perc;
-            statsData.fanRPM = power_management->fan_rpm;
-            statsData.wifiRSSI = wifiRSSI;
-            statsData.freeHeap = esp_get_free_heap_size();
+            statsData.voltage = power_management->voltage / 1000.0;
+            statsData.current = Power_get_current(GLOBAL_STATE) / 1000.0;
+            statsData.coreVoltageActual = VCORE_get_voltage_mv(GLOBAL_STATE) / 1000.0;
+            statsData.fanSpeed = (float)power_management->fan_perc;
```
KillerInk: now they need a cast^^

terratec: yes but is there a generic way for the dynamic assignment (label -> value) in `http_server.c` ? I didn't want to do a string comparison for each value individually.
And the unit conversion would be done for each value on each request in the frontend.

terratec:
@duckAxe With #996 a new subscription for `this.info$` was added. It looks like the interval object now survives the optimized reload logic.
Can I fix it like this or is there a better way?
`    this.titleSubscription = this.info$.subscribe(info => {
      this.titleService.setTitle(
        [ ...`
and in updateSystem():
`... next: () => { this.titleSubscription.unsubscribe(); ...`

duckaxe:
@terratec 👍

main/tasks/statistics_task.c
KillerInk: yeah but strings are always bad^^ for me the better way would be an enum that represent the labels.

KillerInk: and you use that enum already in the frontend

terratec: Reverted the float types

KillerInk: love it

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -92,35 +122,51 @@ export class HomeComponent {
...
-
+  public updateSystem() {
+    const form = this.form.getRawValue();
+
+    this.systemService.updateSystem(this.uri, form)
+      .pipe(this.loadingService.lockUIUntilComplete())
+      .subscribe({
+        next: () => {
+          this.titleSubscription?.unsubscribe();
+          // Clear and reload previous data
```
WantClue: remove unnecessary comment

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -42,24 +49,48 @@ export class HomeComponent {
...
     private quickLinkService: QuicklinkService,
     private titleService: Title,
+    private loadingService: LoadingService,
+    private toastr: ToastrService,
     private shareRejectReasonsService: ShareRejectionExplanationService
   ) {
     this.initializeChart();
+    this.loadPreviousData();
 
     // Subscribe to theme changes
```
WantClue: remove that as well please

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -202,29 +247,77 @@ export class HomeComponent {
...
+
+      if (stats.chartY1Data === stats.chartY2Data) {
+        idxChartY2Data = idxChartY1Data;
+      } else if (stats.chartY2Data === eChartLabel.hashrate) {
+        idxChartY2Data = 1;
+      } else if (stats.chartY2Data === eChartLabel.power) {
+        idxChartY2Data = 2;
+      } else if (stats.chartY2Data === eChartLabel.none) {
+        idxChartY2Data = -1;
+      } else if (idxChartY1Data < 3) { // no additional data for Y1
```
WantClue: remove comment please

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -202,29 +247,77 @@ export class HomeComponent {
...
+        idxChartY2Data = 3;
+      }
+
       stats.statistics.forEach(element => {
-        const idxHashrate = 0;
-        const idxTemperature = 1;
-        const idxPower = 2;
-        const idxTimestamp = 3;
+        element[idxHashrate] = element[idxHashrate] * 1000000000; // convert to H/s
```
WantClue: no explanaition needed

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -238,27 +331,48 @@ export class HomeComponent {
       switchMap(() => {
         return this.systemService.getInfo()
       }),
+      map(info => {
+        info.hashRate = info.hashRate * 1000000000; // convert to H/s
```
WantClue: here as well

mutatrum:
It seems the dropdown color doesn't play nice with the theme colors?

image

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -202,29 +247,77 @@ export class HomeComponent {
...
+    this.chartData.datasets[0].data = this.chartY1Data;
+    this.chartData.datasets[1].data = this.chartY2Data;
+  }
 
+  private loadPreviousData()
+  {
     // load previous data
     this.stats$ = this.systemService.getStatistics().pipe(shareReplay({ refCount: true, bufferSize: 1 }));
     this.stats$.subscribe(stats => {
+      const idxTimestamp = 0;
```
mutatrum: This section (L259-283) is confusing, what's exactly happening here?

mutatrum:
Another nit is the box and dropdown is scaled to the current selected item, causing a scrollbar and cut off labels:

image

main/http_server/axe-os/src/app/components/home/home.component.ts
mutatrum: After reading back and forth, I understand now. I really like this feature, but the code is quite complex. I wonder if - maybe on the next version iteration - it would be simpler to just have arrays for everything, instead 2 specials for hashrate and power, and 2 specials for y1 and y2. There's also quite some data mangling going on, those could all be delegated to the label callback.

main/nvs_config.h
```diff
@@ -34,6 +34,8 @@
 #define NVS_CONFIG_OVERCLOCK_ENABLED "oc_enabled"
 #define NVS_CONFIG_SWARM "swarmconfig"
 #define NVS_CONFIG_STATISTICS_FREQUENCY "statsFrequency"
+#define NVS_CONFIG_CHART_Y1_DATA "chartY1Data"
```
mutatrum: I think this should not be in NVS. Y1/Y2 data selection is front-end data only. We should limit writes to NVS. The browser storage is probably a better place for this.

main/http_server/http_server.c
```diff
@@ -733,25 +794,64 @@ int create_json_statistics_dashboard(cJSON * root)
 
     if (root) {
         // create array for dashboard statistics
+        char * chartY1Data = nvs_config_get_string(NVS_CONFIG_CHART_Y1_DATA, STATS_LABEL_HASHRATE);
```
mutatrum: A better design here would be to accept an array of field names for the back-end to fill in. Now the front-end design is tied into the REST API and the back-end.

Cleaner would be to have a request body with a list of fields the front-end wants. Obviously, hashrate and power will always be requested, but there can be other fields as well. The return payload is then similar to what it's now, so an array of data points. To be able to parse it on the front-end, a list of field names that denote the index in that array would be needed.

```
{
  "fields":["timestamp", "hashrate", "power", "voltage"],
  "data":[
      [12345636, 1325, 15, 5],
      [12345637, 1340, 15.1, 5.2],
     ... etc
   ]
}

This design leaves the door open to easily retrieve extra data, for future use-cases. It can then also be used for external tools.

**Comment by duckaxe** [comment]
@terratec [Dropdowns.patch](https://github.com/user-attachments/files/20922809/Dropdowns.patch)

File: main/http_server/axe-os/src/app/components/home/home.component.html [hunk 11]
```diff
@@ -56,7 +56,7 @@
                         <span class="block text-500 font-medium mb-3">Efficiency</span>
                         <div class="text-900 font-medium text-2xl flex align-items-center gap-2">
                             <span *ngIf="!info.power_fault">
-                                {{info.power / (info.hashRate/1000) | number: '1.2-2'}} <small>J/TH</small>
+                                {{info.power / (info.hashRate / 1000000000000) | number: '1.2-2'}} <small>J/TH</small>
```
WantClue: ideally we don't want any calc in the frontend html maybe we can move t his over to ts.

duckaxe: @WantClue We should generalize all hashrate values so we don't have to calculate them on the frontend side. Maybe an extra PR?
Screenshot 2025-07-07 at 11 40 02

main/http_server/http_server.c
terratec: The statistics API endpoint now supports parameters. After the redesign I can delete the separate dashboard API endpoint.

All data:
http://bitaxe/api/system/statistics

Selected data:
http://bitaxe/api/system/statistics?Hashrate&ASIC%20Temp&Current&Free%20Heap

main/http_server/http_server.c
```diff
@@ -831,11 +900,87 @@ static esp_err_t GET_system_statistics(httpd_req_t * req)
...
+    bool dataSelection[SRC_NONE] = {false};
+    bool selectionCheck = false;
+    int prebuffer = 0;
+
+    // Check query parameters
+    if (1 < bufLen) {
+        buf = (char *)malloc(bufLen);
+        if (buf) {
+            if (httpd_req_get_url_query_str(req, buf, bufLen) == ESP_OK) {
+                char * param = strtok(buf, "&");
```
mutatrum: You can `httpd_uri_decode` the buffer, and get rid of the `bool urlEnc` in `strToDataSource` and the double string constants at the top.

main/http_server/http_server.c
mutatrum: This simplified it a lot, IMO. It's a non-standard query convention though, as it's not key=value. 

Secondly, I would try to keep the labels front-end only. Now, the label (f.e. `ASIC Temp`) is used for display, but also in the query, data selection and NVS field. Ideally, use a URL/JSON/encoding safe version, which decouples the display value from the rest. So combined:

`http://bitaxe/api/system/statistics?columns=hashrate,asicTemp,current,freeHeap`

This also eliminated my earlier comment on URL decoding.

main/nvs_config.h
terratec: The `LocalStorageService` is now in use.

main/http_server/axe-os/src/app/components/home/home.component.ts
terratec: The mapping should be less complex now.

main/http_server/openapi.yaml
```diff
@@ -459,6 +467,40 @@ components:
...
+            - asicVoltage
+            - voltage
+            - power
+            - current
+            - fanSpeed
+            - fanRpm
+            - wifiRssi
+            - freeHeap
+          examples:
+            - "Hashrate"
```
mutatrum: Tiny remark: example needs to be updated, same below

mutatrum (approved):
LGTM, hope we can get this merged in the next version. Is there stull stuff to do as it's in draft?

terratec:
What about the hashrate calculation in the frontend html? Extra Issue?

WantClue:
> What about the hashrate calculation in the frontend html? Extra Issue?

yes this should be it's own issue please.
I'm making sure to review and test this PR so that we can merge this ASAP

main/http_server/axe-os/src/app/components/home/home.component.html
```diff
@@ -58,7 +58,7 @@
                         <span class="block text-500 font-medium mb-3">Efficiency</span>
                         <div class="text-900 font-medium text-2xl flex align-items-center gap-2">
                             <span *ngIf="!info.power_fault">
-                                {{info.power / (info.hashRate/1000) | number: '1.2-2'}} <small>J/Th</small>
+                                {{info.power / (info.hashRate / 1000000000000) | number: '1.2-2'}} <small>J/Th</small>
```
duckaxe: I would remove `<small>` element to be uniform with neighboring card designs.

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -105,33 +133,52 @@ export class HomeComponent implements OnInit, OnDestroy {
...
+  public updateSystem() {
+    const form = this.form.getRawValue();
+
+    this.storageService.setItem('chartDataSources', JSON.stringify(form));
+
+    this.systemService.updateSystem(this.uri, form)
+      .pipe(this.loadingService.lockUIUntilComplete())
+      .subscribe({
+        next: () => {
+          this.titleSubscription?.unsubscribe();
```
duckaxe, line 145: Why we need an unsubscribe?

duckaxe:
https://github.com/terratec/ESP-Miner/pull/1

Another idea would be to synchronize the two dropdowns. When something is selected in one dropdown, remove that entry from the other dropdown. But nice to have.

main/http_server/axe-os/src/app/components/home/home.component.ts
terratec: Each source selection would add a subscription with additional data points.

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -253,27 +339,48 @@ export class HomeComponent implements OnInit, OnDestroy {
       switchMap(() => {
         return this.systemService.getInfo()
       }),
+      map(info => {
+        info.hashRate = info.hashRate * 1000000000;
```
duckaxe, line 343: You need to adopt the hashrate @ https://github.com/bitaxeorg/ESP-Miner/blob/07d6371de69a50b92f301fcbb0af0ec0213c473c/main/http_server/axe-os/src/app/components/home/home.component.ts#L428 too.

main/http_server/axe-os/src/app/components/home/home.component.html
terratec: #1241

duckaxe:
Tested on my Bitaxe. Looks 👍

kakulukia:
Dont really know if this is the source of the problem, but as its realated i just wanted to report it here.

I installed the current git version and really like the new dashboard and also that it kinda took the freedom to auto activated data logging to not start without data to display.

But now every time i visit the settings page data logging is set to "Every 10 minutes, 28 seconds for 5 days, 5 hours, 43 minutes, 12 seconds" not respecting the previous setting. 

Was that introduced with the merge of this PR or where does that actually belong to?

Another side note: The dropdown entries for switching the data source look rather random - whats the logic behind that?

terratec:
> Dont really know if this is the source of the problem, but as its realated i just wanted to report it here.
> 
> I installed the current git version and really like the new dashboard and also that it kinda took the freedom to auto activated data logging to not start without data to display.
> 
> But now every time i visit the settings page data logging is set to "Every 10 minutes, 28 seconds for 5 days, 5 hours, 43 minutes, 12 seconds" not respecting the previous setting.
> 
> Was that introduced with the merge of this PR or where does that actually belong to?
> 
> Another side note: The dropdown entries for switching the data source look rather random - whats the logic behind that?

The default value for data logging should be 0 (disabled). This seems to be an user-defined value or maybe an old value from an older test version.
First I would try setting the value to 0, clicking save and restart.

The dropdown boxes contain all available data sources from the data logging task.

kakulukia:
Saving and restarting did not change the behaviour. Im on 1873664-dirty as i was testing the autotune feature.

Saving the 0 is respected since i dont see content in the chart right after starting, but as soon as i visit the systems settings the displayed valued is not initialized with the current value of the setting.

And my question about the dropdown content was more about: Why is it not sorted alphabetically? Its at least seems random to me if there was a reason to sort it that way: id like to know.

terratec:
> Saving and restarting did not change the behaviour. Im on [1873664](https://github.com/bitaxeorg/ESP-Miner/commit/1873664fa13ac5104a60b1118a2787aec748b822)-dirty as i was testing the autotune feature.
> 
> Saving the 0 is respected since i dont see content in the chart right after starting, but as soon as i visit the systems settings the displayed valued is not initialized with the current value of the setting.
> 
> And my question about the dropdown content was more about: Why is it not sorted alphabetically? Its at least seems random to me if there was a reason to sort it that way: id like to know.

Maybe it's a side effect of the merge. Have you also tested an image from the master branch?
Probably `info.statsFrequency` is not yet fully fixed @KillerInk
https://github.com/bitaxeorg/ESP-Miner/pull/1152/commits/257e246d0076259710391c9334bde5f23beeb38d


The chart is always empty after booting because RAM is used for data logging.

Ah yes the order is:
- system default values
- temperature related
- voltage related
- power related
- fan related
- system related

kakulukia:
What about also displaying the grouping that way: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/optgroup

No, I have not yet tested the master branch. I will have another look after 2.12 is done as it's not that serious.

terratec:
Something like this: https://v17.primeng.org/dropdown#group
Maybe create an issue?

duckaxe:
Grouping would require quite a bit of technical restructuring (eChartLabel must become an array, the related code must be refactored). Other simple ways would be to sort the options alphabetically or use a frequency-based order (values that are rarely used are placed at the bottom).

PR #958, opened 2025-05-26
fix: change back display conf to v_res

WantClue:


0xf0xx0 (approved):
fixes the issue on my axe

mutatrum (approved):

PR #962, opened 2025-05-27
fix: adds the GT offset

### LINKED ISSUE #881: need to tune temperature sensor for bitaxeGT 800
We tuned the gain and ideality settings on the EMC2101 for the BM1370 bitaxeGamma 600. bitaxeGamma Turbo uses the same ASIC, but of course we use a dual temp reader, the EMC2103 and it has different gain and ideality settings.

WantClue:
closes #881


benjamin-wilson (commented):
The property name should be external

benjamin-wilson:
Nack 
image
You're adding 5C to the external temp that was not there before

mutatrum:
Well spotted, thanks!

PR #963, opened 2025-05-27
Removal of unnecessary code @ edit.component.ts

duckAxe:
This change removes the remaining code from `edit.component.ts`. It was probably left behind when the network settings were moved to their own tab.


WantClue (approved):
Good catch

mutatrum:
Can you post your Discord username or id somewhere, or send a mail if you don't want to post that here, so the admins can unban you?

PR #967, opened 2025-05-28
Show restart warning only for relevant fields

duckaxe:
I have seen that the [request](https://github.com/bitaxeorg/ESP-Miner/pull/741) has stalled. I think I have found a simple solution: working with a whitelist of fields.

By default, all fields trigger the restart warning. However, you can add single fields to the whitelist to prevent this. This method is simple and easy to maintain, even with newly added fields.

The restart warning is available on the following pages:
* Network
_All fields trigger the restart warning._
* Pool
_All fields trigger the restart warning._
* Settings
_The following fields trigger the restart warning:_
-- `display`
-- `flipscreen`
-- `invertscreen`

Have I forgotten something? Please let me know.

https://github.com/user-attachments/assets/59d71242-8364-4642-ad97-7c77bcdc53c3

Closes https://github.com/bitaxeorg/ESP-Miner/issues/581

### DIFF
--- a/main/http_server/axe-os/src/app/components/edit/edit.component.ts
+++ b/main/http_server/axe-os/src/app/components/edit/edit.component.ts
@@ -252,6 +252,7 @@ export class EditComponent implements OnInit, OnDestroy {
 
   get noRestartFields(): string[] {
     return [
+      'displayTimeout',
       'coreVoltage',
       'frequency',
       'autofanspeed',


WantClue:
@duckAxe the `displayTimeout` should not trigger the restart warning

duckAxe:
> @duckAxe the `displayTimeout` should not trigger the restart warning

Easy. Just added to the whitelist. Fixed.

PR #986, opened 2025-05-30
Unify ASIC reset code

mutatrum:


components/asic/asic.c
```diff
@@ -9,13 +9,25 @@
...
 
 static const double NONCE_SPACE = 4294967296.0; //  2^32
 
 static const char *TAG = "asic";
 
 uint8_t ASIC_init(GlobalState * GLOBAL_STATE)
 {
+    ESP_LOGI(TAG, "Initializing %s", GLOBAL_STATE->DEVICE_CONFIG.family.asic.name);
+
+    esp_rom_gpio_pad_select_gpio(GPIO_ASIC_RESET);
```
mutatrum, line 24: It's debatable if this code belongs here, or outside of this component. Strictly seen it should be outside, but then the code needs to be duplicated in `main` and `self_test`. Any suggestions are welcome.

components/asic/bm1366.c
```diff
@@ -20,12 +20,6 @@
 #define BM1366_CHIP_ID 0x1366
 #define BM1366_CHIP_ID_RESPONSE_LENGTH 11
 
-#ifdef CONFIG_GPIO_ASIC_RESET
```
mutatrum, line 23: This probable never worked, `CONFIG_GPIO_ASIC_RESET` is undefined here.

main/main.c
```diff
@@ -102,6 +103,12 @@ void app_main(void)
     queue_init(&GLOBAL_STATE.stratum_queue);
     queue_init(&GLOBAL_STATE.ASIC_jobs_queue);
 
+    if (asic_reset() != ESP_OK) {
```
mutatrum, line 106: Subtle change here, needs some testing. The asic is reset _before_ the serial port is initialised.

WantClue: Will do some testing on this

0xf0xx0 (approved):

PR #990, opened 2025-05-31
Added share accepted LED functionality to accessory port GPIO39

Nexus9090:
Added functionality for an LED to be connected to the accessory port of a Bitaxe board using GPIO39 as the driver.

The purpose being to show activity when the board is headless or the display is turned off.

The code implements an ESP one-shot timer.

When a stratum share accepted message is received, the **accepted_led_trigger();** function is called setting GPIO39 high and starting the one-shot timer which when expired turns off the LED, thus flashing the LED to show share accepted activity.

An LED and suitable size resistor are required to be connected to the accessory port of a board with the anode of the LED attached to GPIO39 and the cathode connected via a resistor ~1K to GND

https://github.com/user-attachments/assets/2f4cd4f2-f65b-4c7f-bde3-f167a7f5597b

### DIFF
--- a/main/CMakeLists.txt
+++ b/main/CMakeLists.txt
@@ -1,5 +1,6 @@
 idf_component_register(
 SRCS
+    "accepted_led.c"
     "adc.c"
     "i2c_bitaxe.c"
     "main.c"
--- a/main/accepted_led.c
+++ b/main/accepted_led.c
@@ -1,0 +1,106 @@
+//////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Source file : accepted_led.c
+// Header file : accepted_led.h
+//
+// Description : For Open Source Bitaxe ESP-Miner project.
+//               implements one-shot timer to flash LED on GPIO upon share acceptance
+//
+// Requirements : An LED and suitable series resistor should be connected to the allocated GPIO port of the ESP32-S3 MCU
+//				  so that the anode of the LED is attached to the ESP32-S3 GPIO and the resistor ties the cathode of the LED to GND.
+//                It is intended to use the accesory port on the Bitaxe board to connect an LED to.
+//                The default GPIO port for the LED is GPIO39.
+//
+// accepted_led_init(); should be called in system.c at "SYSTEM_init_peripherals" to initialise the GPIO and Timer
+// accepted_led_trigger(); should be called in stratum_task.c when an approprate stratum share accepted message is received
+//
+// TODO: add error checking
+//////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include "esp_log.h"
+
+#include "esp_timer.h"
+#include "sdkconfig.h"
+#include "driver/gpio.h"
+
+//#define accepted_led_duration uint64_t 200000 //200mS 200000uS
+
+#define GPIO_ACCEPTED_LED CONFIG_GPIO_ACCEPTED_LED
+
+#ifndef CONFIG_GPIO_ACCEPTED_LED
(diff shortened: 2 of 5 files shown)


mutatrum:
I'm not sure we should be adding things to the firmware that are not part of default builds. At the very least, _if_ it's added, it should be configurable, either via a `Kconfig.projbuild` parameter (but this needs a custom firmware), and maybe via an NVS setting.

skot:
I think the ideal way to add this is make a BAP board and communicate over the (WIP) BAP protocol to turn your LED on and off. I realize this is a lot of work to be blinking a single LED, but we can't be adding code to mainline esp-miner for every accessory.

MyOwn2C:
ESP-Miner was meant to be a bare bone OS to mine with BM ASICS. 
IMO it is slowly getting bloated. 
There is another PR to add a dot to LCD when shares are accepted. That is much more viable than a separate LED using precious GPIO pins.

Nexus9090:
> ESP-Miner was meant to be a bare bone OS to mine with BM ASICS. IMO it is slowly getting bloated. There is another PR to add a dot to LCD when shares are accepted. That is much more viable than a separate LED using precious GPIO pins.

I wholehartedly disagree.

There are fundamental missing features and this in my opinion is one of them.

As for GPIO shortages, I think not, there's only a few allocated and in use.

MyOwn2C:
> I wholehartedly disagree.
> 
> There are fundamental missing features and this in my opinion is one of them.
> 
> As for GPIO shortages, I think not, there's only a few allocated and in use.

This is open source. 
So you are free to make your own branch, make your own LED board and load it with your own software. 
The PR for dot on LCD for shares accepted is already done and that is a much better implementation than hardware LED using GPIO to blink for a very minor function. 
You may even find yourself turning off your hardware shares accepted LED afterwards since it will blink all-the-time., thousands of times a day.

Nexus9090:
> > I wholehartedly disagree.
> > There are fundamental missing features and this in my opinion is one of them.
> > As for GPIO shortages, I think not, there's only a few allocated and in use.
> 
> This is open source. So you are free to make your own branch, make your own LED board and load it with your own software. The PR for dot on LCD for shares accepted is already done and that is a much better implementation than hardware LED using GPIO to blink for a very minor function. You may even find yourself turning off your hardware shares accepted LED afterwards since it will blink all-the-time., thousands of times a day.

Its clear you've no intention of discussing this positively or objectively, so I've close the PR.

So much for community spirit.

Nexus9090:
> I'm not sure we should be adding things to the firmware that are not part of default builds. At the very least, _if_ it's added, it should be configurable, either via a `Kconfig.projbuild` parameter (but this needs a custom firmware), and maybe via an NVS setting.
> 
> If you would get someone to actually make a board with this added, I would suggest not using one of the BAP GPIOs but have some other ESP32 pin used for this. Alternatively, maybe it's possible to make a tiny BAP LED PCB that one could hook into the bap port. I think this would make a stronger case for adding it to the firmware.
> 
> If there's only one out there, I don't think that warrants adding the code to the firmware.
> 
> Having said that: it's a cool addon!

Thank you @mutatrum I think it is a valuable addition. Never the less, I've been met with absolute negativity over what I started out with as an attempt to add something positive to the collaboration.

Anway, I appreciate your help along the way.

skot:
The first BitaxeMax HW had two RGB LEDs. But this was before adding the OLED, which kinda made status LEDs redundant.

I appreciate you taking the time to hack on this stuff and build what you want to see -- that is the way!

I personally think it's nice not having a status LED.. it's unique in a way. Concept NACK from me, on the HW.

Also, with MENUCONFIG maybe we could optionally add LED code into esp-miner?

Issue #657, opened 2025-01-16
Fix errors when accessing AxeOS by hostname


mutatrum:
Several calls are broken when opening AxeOS by hostname after #637. Swarm should also support adding devices by hostname.

benjamin-wilson:
Due to CORS this not not possible.

PR #996, opened 2025-06-01
Add browser tab title

duckAxe:
The title of the current browser tab is always the same: AxeOS. This PR expands the title for each page/view.
The home page title also displays the current hash rate and ASIC temperature. This allows you to view the current status without having the AxeOS tab open.

https://github.com/user-attachments/assets/136e423e-6c8a-49fe-b38a-84e7a1aef5a1


mutatrum:
What about showing `hostname`, for the people with more than one device?

duckaxe:
As you can see, the visible space inside the tab is very limited. We can only fit a small amount of information here. A hostname may be useful, but there is no room for it in the title. We can kill the hash rate & temp and include the hostname instead. But users with only one device will not benefit from this.

Screenshot 2025-06-02 at 10 27 10

mutatrum:
My counter example:

image

I have anywhere between 60 and 100 tabs open, I don't see titles, and maybe see icons if I'm lucky :laughing:

Chrome shows a preview of the window, with lots of room for a long title. Looking at other website (GitHub f.e.) they add the complete issue title on the tab. And for websites that are location relevant, most show the location information. F.e. the dashboard of a solar installations shows the location, not the stats.

So not sure, either way can be useful.

duckaxe:
> I have anywhere between 60 and 100 tabs open, I don't see titles, and maybe see icons if I'm lucky 😆

Haha. 

> Chrome shows a preview of the window, with lots of room for a long title.

Yes, and Chrome shows the host/domain of the tab. This alone allows you to identify your device. But I'm with you. How about this device information (like swarm)?

* host name
* hash rate
* ASIC/VR temp
* power
* best diff

Screenshot 2025-06-02 at 14 40 09

0xf0xx0 (approved):

mutatrum:
> Yes, and Chrome shows the host/domain of the tab. This alone allows you to identify your device. 

This is nuanced, as you can't properly access the devices by their hostname (see #657, needs to be revisited), I only see the ip. To detect if the page is accessed by either hostname or ip and only show hostname on the latter is probably overkill, but that would be the icing on the cake solution.

> But I'm with you. How about this device information (like swarm)?
> 
> * host name
> * hash rate
> * ASIC/VR temp
> * power
> * best diff

It's a lot, but it's not in the way so I think it's good.

A nice follow-up on this PR would be to show overheat, power faults or block found (and any other persistent messages that show in the top header on the dashboard), instead of hash rate and further, if any of those happen.

duckaxe:
> A nice follow-up on this PR would be to show overheat, power faults or block found (and any other persistent messages that show in the top header on the dashboard), instead of hash rate and further, if any of those happen.

We'll see what the user feedback is like. If it is well liked, we can expand the title to include the information you mention.

main/http_server/axe-os/src/app/app-routing.module.ts
```diff
@@ -11,14 +11,17 @@ import { PoolComponent } from './components/pool/pool.component';
 import { AppLayoutComponent } from './layout/app.layout.component';
 import { ApModeGuard } from './guards/ap-mode.guard';
 
+const TITLE_PREFIX = 'AxeOS';
```
mutatrum, line 14: `Axe-OS`

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -280,6 +287,19 @@ export class HomeComponent {
...
     );
+
+    this.info$.subscribe(info => {
+      this.titleService.setTitle(
+        [
+          this.pageDefaultTitle,
+          info.hostname,
+          (info.hashRate ? HashSuffixPipe.transform(info.hashRate * 1000000000) : false),
+          (info.temp ? `${info.temp}${info.vrTemp ? `/${info.vrTemp}` : ''} °C` : false),
+          (info.power ? `${info.power.toFixed()} W` : false),
```
mutatrum: One decimal for power if it's < 100W? Not sure how the dashboard shows this.

mutatrum (approved):
LGTM, small comments

main/http_server/axe-os/src/app/components/home/home.component.ts
duckaxe: Now I use `info.power`, as the dashboard also uses for the power output.

main/http_server/axe-os/src/app/app-routing.module.ts
duckaxe: I wouldn't change that here. We have too many places with `AxeOS`. If we want to change it, we should do so everywhere with a separate PR.

Screenshot 2025-06-10 at 08 50 08

WantClue (approved):

PR #1005, opened 2025-06-05
Add release notes

duckAxe:
This PR adds a link with release notes to the update section so that the user is informed about the release changes. 

The `Release Notes` link is grey colored because it is less important than the two `.bin` links.

https://github.com/user-attachments/assets/89d6d580-53c3-48b0-8fbf-15c3ebd6691d

### DIFF
--- a/main/http_server/axe-os/src/app/components/settings/settings.component.html
+++ b/main/http_server/axe-os/src/app/components/settings/settings.component.html
@@ -27,6 +27,17 @@
                         [href]="asset.browser_download_url">www.bin</a>
                 </div>
             </div>
+
+            <a class="text-500 underline block mt-3 cursor-pointer" type="button" (click)="showReleaseNotes = true">
+                Release Notes
+            </a>
+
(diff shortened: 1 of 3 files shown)


duckaxe:
Blocked by https://github.com/bitaxeorg/ESP-Miner/pull/1014

mutatrum:
For the paranoid out there, one feature would be to have a setting to categorically disallow external calls. That, or have a modal confirmation popup the first time Axe-OS calls out to GitHub, maybe? Don't think it should be with this PR.

duckaxe:
@mutatrum Paranoid people have already clicked the `release check` button. The link to the release notes will only appear after this step. If we want user confirmation for extern calls, then is the check button the right place. We can put the existing text "Clicking this button will connect to GitHub to check for the latest release, both files are required." in a [modal](https://github.com/bitaxeorg/ESP-Miner/pull/1014) and ask the user for confirmation. But it shouldn't be part of this PR.

WantClue (approved):

PR #1006, opened 2025-06-05
Add AxeOS version and mismatched version warning

### LINKED ISSUE #1003: Add firmware version to AxeOS / www.bin
I think this can be as "simple" as a static field in AxeOS Settings that gets automagically updated at build time with the current version string. This will be a very helpful indication for users that have forgotten to update www.bin or did run the update and it failed.

This is meant to be a stop-gap until we get get a unified update image like in #1002

mutatrum:
Fixes #1003

A new file `version.txt` with the app version according to `git describe --tags --always --dirty` is generated and added to `www.bin`. This file is uncompressed to make it simple to read. It's not served over the http server, but the value is exposed on the `info` API endpoint.

When the version of AxeOS and ESP-Miner are different, AxeOS shows a warning banner and an error is logged.

New field in Overview:

image

New log line on start-up:

image

When only updating `www.bin` it shows a warning on the dashboard:

image

and an error in the logs:

image

NOTE: Wording has changed, I didn't want to make new screenshots. Latest wording:
```
Firmware and Axe-OS versions do not match. Please make sure to update both esp-miner.bin and www.bin.
```
Axe-OS gets that information from the `/api/system/info` API endpoint:

image

It works for release (f.e. `v.2.9.0`) and pre-release (f.e. `v.2.9.0b3`), local (f.e. `v2.8.0-3-9ebfbfd-dirty`) and GitHub (f.e. `9ebfbfd`) builds.


mutatrum:
> Generated /home/runner/work/ESP-Miner/ESP-Miner/main/http_server/axe-os/dist/axe-os/version.json with version undefined

The `npm build` stage indeed doesn't know about the version information.

skot:
NPM is pretty fancy.. is there no way to find the current git tag?

duckaxe:
Regarding to https://stackoverflow.com/a/69919067 you can use `release.tag_name` for the current release tag.

Untested: [Untitled.patch](https://github.com/user-attachments/files/20608499/Untitled.patch)

mutatrum:
Only the git tag is not sufficient, it has to match whatever is in `esp_app_get_description.version`.

image

It wasn't clear how that was done. Problem was that esp version was set in the CMake part, which - on GitHub - is ran after the `npm build` stage. After some searching, I found the CMake bit that's responsible for that:

https://github.com/rpavlik/cmake-modules/blob/76c1eb4fbb2e5988a3bc0fdc1ff4c326b737a10c/GetGitRevisionDescription.cmake#L248

I added the same `git describe --tags --always` to get the version string. I'll clean up the code.

mutatrum:
New field in Overview:

image

New log line on start-up:

image

When only updating `www.bin` it shows a warning on the dashboard:

image

and an error in the logs:

image

NOTE: Wording has changed, I didn't want to make new screenshots.

> Firmware and Axe-OS versions do not match. Please make sure to update both esp-miner.bin and www.bin.

mutatrum:
It's also working for GitHub builds.

The version that goes into `www.bin`:
```
Generated /home/runner/work/ESP-Miner/ESP-Miner/main/http_server/axe-os/dist/axe-os/version.txt with version 9ebfbfd
```

and the version that goes into `esp-miner.bin`:
```
-- App "esp-miner" version: 9ebfbfd
```

mutatrum:
The file is also reachable:

image

Although I'm not sure if that's still needed. Axe-OS gets that information from the `/api/system/info` API endpoint:

image

mutatrum:
> The file is also reachable:
> 
> image

Decided to remove this. No need to actually serve `version.txt`.

WantClue:
> New field in Overview:
> 
> image
> 
> New log line on start-up:
> 
> image
> 
> When only updating `www.bin` it shows a warning on the dashboard:
> 
> image
> 
> and an error in the logs:
> 
> image
> 
> NOTE: Wording has changed, I didn't want to make new screenshots. Latest wording:
> 
> ```
> Firmware and Axe-OS versions do not match. Please make sure to update both esp-miner.bin and www.bin.
> ```

UUH I like that

skot:
This is excellent! Nice work @mutatrum I think this is going to eliminate a lot of problems users are seeing.

main/http_server/axe-os/src/app/components/logs/logs.component.html
```diff
@@ -39,9 +39,13 @@ <h2>Overview</h2>
                     <td>{{info.freeHeap}}</td>
                 </tr>
                 <tr>
-                    <td>Version:</td>
+                    <td>Firmware Version:</td>
                     <td>{{info.version}}</td>
                 </tr>
+                <tr>
+                    <td>Axe-OS Version:</td>
```
mutatrum, line 46: Maybe this should only show if it's actually different.

main/http_server/axe-os/src/app/components/system/system.component.html
```diff
@@ -39,9 +39,13 @@ <h2>System</h2>
                     <td>{{info.freeHeap}}</td>
                 </tr>
                 <tr>
-                    <td>Version:</td>
+                    <td>Firmware Version:</td>
                     <td>{{info.version}}</td>
                 </tr>
+                <tr>
+                    <td>Axe-OS Version:</td>
+                    <td>{{info.axeOSVersion}}</td>
```
mutatrum, line 47: Maybe this should only show if it's actually different.

skot: I like the idea of always showing both AxeOS and esp-miner versions.

WantClue (approved):

PR #1012, opened 2025-06-06
Optimised Fonts

duckaxe:
This PR introduces optimisations that reduce the size of the www bundle (898 KB vs. 82 KB).

All modern browsers [support the WOFF2 format](https://caniuse.com/woff2). Other web fonts formats such as WOFF, SVG, TTF, and OTF are unnecessary and must be deleted.

The following optimisations have been made:

1. Non-WOFF2 fonts for `AngelWish`, `Nippo-Regular` and `Primeicons` are removed.
2. Integrate `Primeicons` into the project so that we don't put unnecessary font files into the www bundle, which happens automatically as soon as it used as an `npm` package.

**Before (898 KB)**
Screenshot 2025-06-06 at 14 27 04

**After (82 KB)**
Screenshot 2025-06-10 at 19 14 24


0xf0xx0 (approved):

mutatrum:
What's the drawback of also removing the `woff` fonts?

duckaxe:
I didn't want to delete fonts too aggressively. I would keep the fallback font. However, if we need to reduce the package size further, then I will delete woffs.

duckaxe:
@mutatrum I checked the [NerdQAxePlus](https://github.com/shufps/ESP-Miner-NerdQAxePlus) repo, there is WOFF2 files only. It seems to be working fine for the end users. I have now deleted WOFF for us, too.

duckaxe:
PR description updated.

WantClue:
@duckAxe Please consider resolving the conflicts.
LGTM and can be merged after conflicts have been resolved

duckaxe:
Solved.

PR #1018, opened 2025-06-09
Custom board config

mutatrum:
Allows overriding board specific configurations through NVS configuration.

Currently, it's not possible to define a whole new family or ASIC, as there's no code for that yet. A follow-up PR could extend this to defining a custom family.

With this it's possible to add a new board version to an existing family.


WantClue (approved):
looks good to me, opens the doors for customization.



Please resolve the conflicts

PR #1022, opened 2025-06-09
SH1107 (64x128) display support completed

mutatrum:
By upgrading the LVGL 9.3, this fixes the display corruption when rotating the display which was needed to complete the support for the SH1107 display.

Replaced the Flip Screen setting by a Rotation setting.

Also fixed the Bitaxe logo alignment on bigger screens and improved the layout of other screens.

The SH1107 (128x128) is not supported yet: #1028

REST API changes:
- `flipscreen` is removed from `info` endpoint
- `rotation` is added to `info` endpoint


main/CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ EMBED_FILES "http_server/recovery_page.html"
 )
 
 idf_build_set_property(COMPILE_OPTIONS "-DLV_CONF_INCLUDE_SIMPLE=1" APPEND)
-idf_build_set_property(COMPILE_OPTIONS "-DLV_CONF_PATH= ${CMAKE_SOURCE_DIR}/main/lv_conf.h" APPEND)
+idf_build_set_property(COMPILE_OPTIONS "-DLV_CONF_PATH=\"${CMAKE_SOURCE_DIR}/main/lv_conf.h\"" APPEND)
```
mutatrum, line 72: This fixes the compilation problems. `LV_CONF_PATH` now needs to be a string.

PR #1023, opened 2025-06-10
Wait after socket creation

KillerInk:
sometimes it happens that the miner wont connect to the poolsocket.
with a small delay the socket has enough time not to be not null. 
for me that problem is now gone on all problematic pools and miner connect directly


KillerInk:
this is final. all 9 miner had today to reboot more often and all connected everytime.
bevor normaly 1-3 got stucked and need a manual reboot.

Issue #945, opened 2025-05-23
Unconfuse sensor readings and configuration settings


mutatrum:
Currently, the API endpoints have one bag for everything, sensor readings and configuration settings. This leads to confusing field names, such as `maxTemp` and `tempHighest`, and a few others.

One idea would be to separate these out, either in the JSON struct, or into separate endpoints. One for configuration, the other for sensor readings. Or maybe in another division, needs to be investigated.

It's also sort of continuation of the `asic` endpoint, which has hardware configuration. So separate endpoints would make sense IMO.

https://github.com/bitaxeorg/ESP-Miner/pull/888/files#r2103884794

0xf0xx0:
here's my proposal:

## `/api/system/info` renamed to `/api/system/status`
live stats snapshot
```json
{
  "bestDiff",
  "bestSessionDiff",
  "coreVoltageActual",
  "current",
  "expectedHashrate",
  "fanrpm",
  "fanspeed",
  "freeHeap",
  "hashRate",
  "power",
  "sharesAccepted",
  "sharesRejected",
  "sharesRejectedReasons",
  "stratumDiff",
  "temp",
  "uptimeSeconds",
  "voltage",
  "vrTemp",
  "wifiRSSI"
}
```
`wifiStatus` should be dropped, it's kinda a useless key. if you can see it, the axe is already connected to wifi.

## `/api/system/asic` -> `/api/system/board`
General board info, mostly static
```json
{
  "ASICModel",
  "apEnabled",
  "asicCount",
  "boardVersion",
  "defaultFrequency",
  "defaultVoltage",
  "frequencyOptions",
  "display",
  "idfVersion",
  "isPSRAMAvailable",
  "macAddr",
  "maxPower",
  "nominalVoltage",
  "runningPartition",
  "smallCoreCount",
  "version",
  "voltageOptions"
}
```
i feel like `isPSRAMAvailable` and `apEnabled` fit here better, they're not runtime things
`apEnabled` might be a pointless key like `wifiStatus`, the only usecase i can think of is a semi-automated setup process where you connect to the axe and run a script? but that could be done without needing `apEnabled`.

`display` also shouldn't change at runtime. that's... i'd be concerned for anyone that would hot-swap a display. why *is* it mutable?


## `/api/system/config` (or `/api/system/info`)
```json
{
  "autofanspeed",
  "coreVoltage",
  "displayTimeout",
  "fallbackStratumPort",
  "fallbackStratumURL",
[...]

PR #1029, opened 2025-06-11
Remove ASIC specific code from AxeOS

mutatrum:
This eliminated the last ASIC specific code from AxeOS. This is achieved by using the `swarm_color` field from `device_config.h`, which is exposed in the `info` endpoint.

The index of the swarm table also switches from ASIC model to Family name, this makes more sense IMO. It no longer shows family/model names that are not present.

This could be further improved if the swarm stuff first called the `asic` endpoint, and combined this with the `info` endpoint. I added `swarmColor` and `familyName` to the `info` endpoint, but it would be nicer if these were taken from the `asic` endpoint.

Furthermore, if this calls an older firmware, the color is white and the family name is empty. Nothing breaks, but it just looks off for a bit.

API CHANGE: `/api/system/asic`: `boardFamily` field is renamed to `familyName`.
BREAKING CHANGE: Swarm will not show colors and familyName of previous versions

### DIFF
--- a/doc/unit_testing.md
+++ b/doc/unit_testing.md
@@ -12,7 +12,7 @@ idf.py build
 ```
 
 ### Flashing
-**NOTE: Flashing the unit test binary will replace the existing firmware on the ESP32. For example, you will no longer have access to the Axe-OS web UI and must flash a release or build and flash a non-test binary to recover. Do not attempt to do this unless you are willing to spend time recovering (or have dedicated test devices)**
+**NOTE: Flashing the unit test binary will replace the existing firmware on the ESP32. For example, you will no longer have access to the AxeOS web UI and must flash a release or build and flash a non-test binary to recover. Do not attempt to do this unless you are willing to spend time recovering (or have dedicated test devices)**
 
 At the conclusion of the build, instructions are provided to flash it. To ensure esptool uses the correct serial port and has permission to do so, ensure the serial device of the esp32 is known (e.g. from `dmesg` output on connection) and ensure the user is in the `dialout` group.
 
--- a/main/http_server/axe-os/generate-version.js
+++ b/main/http_server/axe-os/generate-version.js
@@ -1,0 +1,9 @@
+const fs = require('fs');
+const path = require('path');
+
+const version = require('child_process').execSync('git describe --tags --always --dirty').toString().trim();
+
+const outputPath = path.join(__dirname, 'dist', 'axe-os', 'version.txt');
+fs.writeFileSync(outputPath, version);
+
+console.log(`Generated ${outputPath} with version ${version}`);
--- a/main/http_server/axe-os/package.json
+++ b/main/http_server/axe-os/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "ng": "ng",
     "start": "ng serve",
-    "build": "ng build --configuration=production && gzipper compress --verbose --gzip --gzip-level 9 ./dist/axe-os && node only-gzip.js",
+    "build": "ng build --configuration=production && gzipper compress --verbose --gzip --gzip-level 9 ./dist/axe-os && node only-gzip.js && node generate-version.js",
     "watch": "ng build --watch --configuration development",
     "test": "ng test",
     "bundle-report": "ng build --configuration=production --stats-json && webpack-bundle-analyzer dist/axe-os/stats.json"
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -16,6 +16,10 @@
         text="Using fallback pool - Share stats reset. Check Pool Settings and or / reboot Bitaxe">
     </p-message>
 
+    <p-message *ngIf="info.version !== info.axeOSVersion" severity="warn" styleClass="w-full mb-4 py-4 border-round-xl"
+        text="Firmware ({{info.version}}) and AxeOS ({{info.axeOSVersion}}) versions do not match. Please make sure to update both www.bin and esp-miner.bin.">
+    </p-message>
+
     <div class="grid">
         <div class="col-12 md:col-6 xl:col-3">
             <div class="card mb-0">
--- a/main/http_server/axe-os/src/app/components/system/system.component.html
+++ b/main/http_server/axe-os/src/app/components/system/system.component.html
@@ -45,8 +45,12 @@
                         <td>{{info.freeHeap}}</td>
                     </tr>
                     <tr>
-                        <td>Version:</td>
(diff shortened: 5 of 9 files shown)


mutatrum:
Example of one device with and one without this PR:

image

main/http_server/axe-os/src/app/components/edit/edit.component.html
```diff
@@ -5,7 +5,7 @@
             text="Custom settings can cause damage & system instability. Only modify these settings if you understand the risks of running outside designed parameters.">
         </p-message>
 
-        <ng-container *ngIf="!settingsUnlocked && restrictedModels.includes(ASICModel)">
+        <ng-container *ngIf="!settingsUnlocked">
```
mutatrum, line 8: No idea what the `restrictedModels` was for.

main/http_server/axe-os/src/app/components/swarm/swarm.component.html
```diff
@@ -182,25 +175,9 @@
...
-        <div class="flex align-items-center gap-1" [style.opacity]="hasModel('BM1370')">
-            <span class="text-green-500">●</span>
-            <span>BM1370 (Gamma)</span>
-        </div>
-        <div class="flex align-items-center gap-1" [style.opacity]="hasMultipleChips()">
-            <span class="text-orange-500">●</span>
-            <span>Multiple Chip Device</span>
+        <div *ngFor="let item of getFamilies" class="flex align-items-center gap-1">
+            <span [class]="'text-' + item.swarmColor + '-500'">●</span>
+            <span>{{ item.familyName }} ({{item.ASICModel}})</span>
```
mutatrum: This will now show family name.

duckaxe:
@mutatrum [Swarm.patch](https://github.com/user-attachments/files/20712742/Swarm.patch)
I moved `familyName` and `swarmColor` to the `asic` endpoint. I tested it with dummy data only because there are no real devices available. Feel free to change my patch.

mutatrum:
> @mutatrum [Swarm.patch](https://github.com/user-attachments/files/20712742/Swarm.patch) I moved `familyName` and `swarmColor` to the `asic` endpoint. I tested it with dummy data only because there are no real devices available. Feel free to change my patch.

I'm not sure if the patch is done on the correct base, it references `ISystemASIC` but there is no such file. Or I'm struggling to apply the patch.

I think it's possible to make a PR on this branch, that way I can just merge it into this branch and you'll be also mentioned as author.

duckaxe:
> I think it's possible to make a PR on this branch, that way I can just merge it into this branch and you'll be also mentioned as author.

https://github.com/mutatrum/ESP-Miner/pull/2

mutatrum:
> > I think it's possible to make a PR on this branch, that way I can just merge it into this branch and you'll be also mentioned as author.
> 
> [mutatrum#2](https://github.com/mutatrum/ESP-Miner/pull/2)

Thanks! Not sure why the patch didn't want to apply. Looking at this combined code, I think I'll move the other new fields also to the ASIC endpoint.

main/http_server/axe-os/api/system/asic_settings.c
```diff
@@ -36,7 +36,8 @@ esp_err_t GET_system_asic(httpd_req_t *req)
 
     // Add ASIC model to the JSON object
     cJSON_AddStringToObject(root, "ASICModel", GLOBAL_STATE->DEVICE_CONFIG.family.asic.name);
-    cJSON_AddStringToObject(root, "boardFamily", GLOBAL_STATE->DEVICE_CONFIG.family.name);
+    cJSON_AddStringToObject(root, "familyName", GLOBAL_STATE->DEVICE_CONFIG.family.name);
```
0xf0xx0, line 39: minor nit, consistent naming would be nice here; `boardVersion`, `boardFamily`, and `boardColor` tell you all you need to know and is extendable (eg `boardVendor` in the future)

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -119,20 +120,19 @@ export class SwarmComponent implements OnInit, OnDestroy {
...
-                IP: ipAddr,
-                ...result
-              };
+      mergeMap(IP =>
+        forkJoin({
+          info: this.httpClient.get(`http://${IP}/api/system/info`),
+          asic: this.httpClient.get(`http://${IP}/api/system/asic`)
+        }).pipe(
+          map(({ info, asic }) => {
+            if ('hashRate' in info) {
```
0xf0xx0: is this needed? wouldn't any errors be handled by `catchError` below?

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -210,20 +217,20 @@ export class SwarmComponent implements OnInit, OnDestroy {
...
-              IP: ipAddr,
-              ...result
-            }
+      mergeMap(IP =>
+        forkJoin({
+          info: this.httpClient.get(`http://${IP}/api/system/info`),
+          asic: this.httpClient.get(`http://${IP}/api/system/asic`)
+        }).pipe(
+          map(({ info, asic }) => {
+            return {IP, ...info, ...asic}
```
0xf0xx0: lol yeah it's not done here

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
duckaxe: I also wondered about this condition. But I didn't want to remove it, because there must be a reason why someone has added it.

0xf0xx0: scream test, imo we should drop it and we'll see  
if it's a problem we can re-add with a descriptive comment 😂

mutatrum: Extracted and re-used this method so they're at least the same.

main/http_server/axe-os/api/system/asic_settings.c
mutatrum: I opted to make it consistent with the internal naming. For the future, we can add `vendorName` or something like that. Might be more for #945

duckAxe (approved):

main/http_server/axe-os/src/app/components/edit/edit.component.html
WantClue: This is a filter to only show the specified values if it matches the build in asic

mutatrum: But `resitrctedModels` contained all the enum values, so what was it restricting?

See edit.component.ts:L35-36

PR #1036, opened 2025-06-15
faster frequency set and if not valid divider got found look for next…

KillerInk:
… best solution.
that fix at special values the frequency got not set and ignored if it failed. in that case it looks now for a solution that can get applied

### DIFF
--- a/components/asic/bm1370.c
+++ b/components/asic/bm1370.c
@@ -140,60 +140,81 @@ void BM1370_set_version_mask(uint32_t version_mask)
 }
 
 void BM1370_send_hash_frequency(float target_freq) {
-    // default 200Mhz if it fails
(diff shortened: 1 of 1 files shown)


KillerInk:
```
I (219596) bm1370: Setting Frequency to 543.75MHz (actual 543.75)
W (219706) bm1370: Using next best PLL settings for frequency 548.87 (actual 550.00, diff 1.13)
I (219706) bm1370: Setting Frequency to 548.87MHz (actual 550.00)
I (219706) frequency_transition: Successfully transitioned ASIC type 1370 to 548.87 MHz
I (219716) asic: Successfully transitioned to new ASIC frequency: 548.87 MHz
```

i have to fix the follow up logs

PR #1045, opened 2025-06-17
PLL lookup table

### LINKED ISSUE #1044: PLL lookup table investigation
With #1036, I wanted to take a look at generating a lookup table for the PLL settings, as the current code might be unpredictable/hard to control.

The idea is to calculate all possible PLL variables (`fb_divider`, `refdiv`, `postdiv1` and `postdiv2`) and the resulting frequency. This is done through a simple NodeJS script. Frequency is calculated by taking the `fb_divider` * 25, which is the VCO frequency, and divide that by `refdiv`, `postdiv1` and `postdiv2`.

As there are frequencies that can be achieved by different setting, the better option is to pick one with a lower VCO frequency. One example, for 525Mhz there are two options:

```
fb_divider=189 refdiv=1 postdiv1=3 postdiv2=3 for a vcoFreq of 4725MHz
fb_divider=168 refdiv=2 postdiv1=2 postdiv2=2 for a vcoFreq of 2100Mhz
```

In this case, the 2nd is preferable. Lower VCO freq is lower power consumption. When deduplicating the list, a set of 2424 unique frequencies, from 40.816327Mhz (160, 2, 7, 7) all the way up to 5975Mhz (239, 1, 1, 1).

Graphed everything is shown here:
Image

What's very noticeable is there are two bands of VCO Frequencies. One strictly below 3000Mhz, and the other strictly above 4000Mhz. With the dividers, this also means there are no possible frequencies between those two ranges, it's not possible to resolve anything there. This is linked to `refdiv`, all the high frequencies are with `refdiv=1`.

Restricting vcoFreq to 3000Mhz options invalidates all solutions with `refdif=1`:
Image

Interesting part here is there are still no resolvable frequencies between 1500Mhz and 2000Mhz.

This might be a pointer to why there sometimes are wild differences in performance and efficiency even with small frequency changes. If you are on a threshold setting, you can suddenly drop to way less efficient PLL combination. Exposing these values would certainly for OC people be interesting.

Script:
https://github.com/mutatrum/ESP-Miner/blob/pll-table/generate_pll_table.js

Generated table, limited to refdiv=2 and max frequency of 1500Mhz:
https://github.com/mutatrum/ESP-Miner/blob/pll-table/components/asic/include/pll_table.h

mutatrum:


mutatrum:
Learnings went into #1051 and #1069.

PR #1046, opened 2025-06-17
fix: process dirs recursively

WantClue:
Font files (AngelWish.woff2, Nippo-Regular.woff2, primeicons.woff2) will now be properly gzipped and served
Eliminates the 302 redirects that were preventing font loading
Resolves the "Failed to decode downloaded font" and "OTS parsing error" issues
Maintains the build optimization while preserving necessary assets

### DIFF
--- a/main/http_server/axe-os/only-gzip.js
+++ b/main/http_server/axe-os/only-gzip.js
@@ -3,28 +3,29 @@ const path = require('path');
 
 const directory = './dist/axe-os';
 
-fs.readdir(directory, (err, files) => {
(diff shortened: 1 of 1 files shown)


duckaxe:
@WantClue The current implementation of WOFF2 file handling (compression, http header) isn't perfect. Perhaps we should correct this finally with your PR? I've created a PR for this.

https://github.com/WantClue/ESP-Miner-WantClue/pull/1

eandersson (approved):
Tested locally

PR #1050, opened 2025-06-19
Refactored settings component

duckAxe:
Since the `settings` component is almost useless and only acts as a wrapper for the `edit` component, the component should be cleaned up.

There isn't much left.


0xf0xx0 (approved):

mutatrum (approved):
The best part is no part.

Issue #1052, opened 2025-06-20
Feature Request: Display HOSTNAME in <title> and top of AxeOS page


homecryptominer:
Not a bug, but to help improve UX. There are many people out there with many Bitaxes. When having multiple tabs opened in a browser, hovering over the tab just shows AxOS and the IP. It would be really useful to also include the actual HOSTNAME in the <title> tag.

In addition to this, it would also be useful to render the HOSTNAME at the top of the AxeOS Dashboard somewhere - perhaps to the right hand side of the hamburger menu?

Thank you so much for your continued support team! :-)
Steve

mutatrum:
You're in luck, this will be added in the next version: #996 

Not sure if the addition of hostname in the hamburger menu adds anything, the URL bar in the browser is there. We might need to revisit #657 for this.

duckaxe:
Maybe like that. Hostname & ASIC added to the page header. Mobile hostname only.

<img width="1152" alt="Image" src="https://github.com/user-attachments/assets/d67b30ce-69b9-4963-8e94-c3b2bd26b0e4" />

<img width="396" alt="Image" src="https://github.com/user-attachments/assets/86748b03-c373-4089-9b91-60432bca5cba" />

duckaxe:
I would prefer the following solution: We should show the hostname and Wi-Fi signal strength because that information is much more important and helpful than the ASIC.

On desktop, we display this information in the top right corner. Since there isn't enough space for it on mobile, we display it in the menu.

**Desktop**

https://github.com/user-attachments/assets/9935fa36-8d68-4272-aad3-bd88a3981d67

**Mobile**

https://github.com/user-attachments/assets/59b71a39-38e7-4482-9bff-c03a315b7e27

mutatrum:
I like that.

skot:
I like this too. My only suggestion is that the icon for hostname should be something looks like a bitaxe and not a computer (since it's the Bitaxe hostname)

duckaxe:
@skot We use [PrimeNg Icons](https://primeng.org/icons#list). There are not many icons that match this.

skot:
hmm, yeah no ideal icons in there. If we really can't have a Bitaxe icon, then maybe one of these?

<img width="109" alt="Image" src="https://github.com/user-attachments/assets/07860db2-c7fb-4767-a32d-917330edc6c0" />

[...]

PR #1056, opened 2025-06-21
Hostname & WiFi RSSI at header/menu

duckAxe:
Discussed https://github.com/bitaxeorg/ESP-Miner/issues/1052

This PR adds the hostname and the WiFI signal strength to the header (desktop) and the menu (mobile).

Changes summary:
* Added hostname & wifiRSSI to header (+ tooltips) on desktop
* Added hostname & wifiRSSI to menu on mobile
* Created Bitaxe SVG icon
* Fixed Content Type for SVG files
* Fixed some code indentations
* Removed unused styles

**Desktop**

https://github.com/user-attachments/assets/bb584691-3452-4ee3-aea5-07994614c3ac

**Mobile**
Screenshot 2025-06-21 at 07 46 08

Closed https://github.com/bitaxeorg/ESP-Miner/issues/1052


mutatrum:
Ok, maybe dumb, but would it look better with the icons on the right side? It _looks_ misaligned, but I think that's just the `-` sign of the RSSI that makes it look off.

Might also work better with long-ish hostnames.

main/http_server/axe-os/src/app/layout/app.menu.component.html
```diff
@@ -4,9 +4,27 @@
...
+        </li>
+        <li class="flex align-items-center gap-2">
+            <i class="pi pi-wifi" style="font-size: 1.143rem"></i>
+            <span [ngClass]="{
+                'text-green-500': info.wifiRSSI > -50,
+                'text-blue-500': info.wifiRSSI <= -50 && info.wifiRSSI > -60,
+                'text-orange-500': info.wifiRSSI <= -60 && info.wifiRSSI > -70,
+                'text-red-500': info.wifiRSSI <= -70
+            }">
+                {{info.wifiRSSI}} dBm
```
mutatrum: Is this updated after page load?

mutatrum (approved):
LGTM code wise, I think this is a really nice addition!

main/http_server/axe-os/src/app/layout/styles/layout/_topbar.scss
```diff
@@ -99,51 +80,5 @@
...
-                    font-size: 1rem;
-                    margin-right: .5rem;
-                }
-
-                span {
-                    font-weight: medium;
-                    display: block;
-                }
-            }
-        }
```
mutatrum, line 123: That's a lot of deleted scss code, all no longer used/needed?

main/http_server/axe-os/src/app/layout/app.menu.component.html
duckaxe: I think so. Why shouldn't it?

main/http_server/axe-os/src/app/layout/styles/layout/_topbar.scss
duckaxe: It looks like it is not needed. But I have reverted my change because it has nothing to do with the PR. I am doing a follow up PR with header code cleaning.

duckaxe:
> Ok, maybe dumb, but would it look better with the icons on the right side? It _looks_ misaligned, but I think that's just the `-` sign of the RSSI that makes it look off.
> 
> Might also work better with long-ish hostnames.

Looks much better. Done. Description updated.

main/http_server/axe-os/src/app/layout/app.menu.component.html
mutatrum: Not sure who is polling the info endpoint? But I'm not angular fluent, so maybe I just don't know how it works.

duckaxe: Loading the info$ is identical to other places, so I think it works here too.

skot:
Don't get me wrong, I love a good logarithmic power ratio like dBm, but maybe this is TMI in a header? Seems like most WiFi systems just show a number of bars (0-3) in the icon.

duckaxe:
> Don't get me wrong, I love a good logarithmic power ratio like dBm, but maybe this is TMI in a header? Seems like most WiFi systems just show a number of bars (0-3) in the icon.

Since we use text and icon for the hostname, we should do the same for Wi-Fi. Otherwise, it won't look good. Icon only should be shown on small displays like the NerdAxe display. The AxeOS header has enough space for more than just an icon. 

You mean like this?

Screenshot 2025-06-22 at 16 31 09

skot:
I think this looks much cleaner. The industry standard afaik is to change the number of "bars" shown on the icon (they're arcs really) to get an idea of the RSSI. Do you think that's doable? 

Here is from the menu in iOS:
image

duckaxe:
At this point, iOS displays icons only. No 5G provider is displayed; just an icon. If we decide to show icons only, the hostname should be removed. We should avoid mixing it. 

If we remove the hostname, we should close this pull request and create a new one, because this PR is about the hostname. WiFi was just a placeholder to fill the gap ;)

mutatrum:
I think the main use case is hostname, the WiFi icon was a nice addition. It's possible to have 3 or 4 WiFi icons for different strenghs with a bit of fiddling, and then have it indeed on the same line.

skot:
I think we should show the hostname still -- that seems very useful. The comparison with iOS was just to show the WiFi signal strength icons.

duckaxe:
> I think the main use case is hostname, the WiFi icon was a nice addition. It's possible to have 3 or 4 WiFi icons with a bit of fiddling, and then have it indeed on the same line.

From a usability standpoint, I would avoid mixing them. We should decide whether to display icons only or icons and text.

Let me explain. Suppose we display the WiFi icon without the RSSI value. The icon itself has a function, presenting the signal strength in color or by using bars. The Bitaxe icon on the left has no function because the hostname is the function. Having a combination of icons (with and without function) is confusing. This is why iOS (and every other operating system) uses a standardised approach on the top bar: only icons, no text.

We should follow best practice.

skot:
Hmm, I see what you mean. I guess we could solve this by requiring a click on the Bitaxe icon to show the hostname. But it is pretty handy to see the hostname at a glance.

duckaxe:
Since https://github.com/bitaxeorg/ESP-Miner/pull/996, we have been showing the hostname in the tab title. Perhaps it is enough for the hostname to be shown via mouse click/hover? In that case, we would display two icons that show more information when the mouse hovers over them.

Screenshot 2025-06-22 at 18 59 16

mutatrum:
We're not Apple and it's not the top bar of a mobile phone. It's unused real-estate on the Axe-OS dashboard. Hostname and WiFi are two stats that are very valuable to see without clicks.

Agreed, the bitaxe logo and WiFi logo don't do anything yet, but they can. Both can be extended, the WiFi logo can change color, and the bitaxe logo f.e. can at some point be leveraged with the status task.

It's not perfect, but fiddling further at this point feels like bikeshedding.

duckaxe:
Another idea is to insert the host name into the logo. Icons only on the right: One for Wi-Fi with fancy bars and one for restarting the device, which is always accessible, even when the menu sidebar is closed.

Screenshot 2025-06-23 at 12 01 36

mutatrum:
Only on desktop, and keep mobile on the bottom of the sidebar?

As this looks weird:
image

duckaxe:
On mobile we have to options: 
1. Put the hostname below the logo
2. Put the hostname to the bottom of the sidebar, as you mean it.

Screenshot 2025-06-23 at 14 18 03

And to keep it consistent (desktop vs mobile), we can also place the hostname below the logo on desktop too.

Screenshot 2025-06-23 at 14 26 03

skot:
I like this. Hostname below the logo looks good. We have to keep in mind that in most cases the hostname will be the default, forever.

duckaxe:
As soon https://github.com/bitaxeorg/ESP-Miner/pull/1058 is merged, I'll continue.

duckaxe:
For the mobile view, I would put the icons in the top right corner. This makes the icons easier to reach. Are there any objections? If not, I'll implement them there.

Screenshot 2025-06-24 at 10 05 51

duckaxe:
**Desktop**

https://github.com/user-attachments/assets/4c8057f6-0e27-430b-8aad-db065330a149

**Mobile**

https://github.com/user-attachments/assets/83cceaca-236d-429a-8ee5-ae30f64b881e

main/http_server/axe-os/src/app/layout/styles/layout/_utils.scss
```diff
@@ -22,3 +22,56 @@
...
+        }
+    }
+
+    &--fair > i:nth-of-type(3),
+    &--good > i:nth-of-type(3),
+    &--good > i:nth-of-type(2),
+    &--excellent > i:nth-of-type(3),
+    &--excellent > i:nth-of-type(2),
+    &--excellent > i:nth-of-type(1) {
+        opacity: 1;
```
mutatrum, line 75: :astonished:

mutatrum (approved):
That wifi-icon is a piece of art

PR #1026, opened 2025-06-11
Rethinking Navigation

duckAxe:
Discussed in https://github.com/bitaxeorg/ESP-Miner/discussions/994
_I opened this PR because https://github.com/bitaxeorg/ESP-Miner/pull/1017 had a wrong rebase. Sorry for that._

**1. Reorder the app navigation**
* Dashboard
* Swarm
* Logs
* System
* Pool
* Network
* Theme
* Settings
* Update
* Whitepaper

**2. Extract sections into a single page**
* System
* Update

Screenshot 2025-06-08 at 17 56 00

PR #1059, opened 2025-06-22
Configure min fan speed percentage

0xf0xx0:
i noticed my chip struggling to get up to temp from a cold start, so this adds a new nvs entry, `minpid` (this should be renamed honestly) that stores the minimum fan speed for the auto fan control  
could be expanded into `p`, `i`, and `d` being tunable?  
no danger of overheating when setting to 0, the fan runs at 70% before pid kicks in  
wildcard pr

i feel like there was a pid issue or pr but i cant find it

### DIFF
--- a/main/http_server/axe-os/src/app/components/edit/edit.component.ts
+++ b/main/http_server/axe-os/src/app/components/edit/edit.component.ts
@@ -162,6 +162,7 @@ export class EditComponent implements OnInit, OnDestroy, OnChanges {
           coreVoltage: [info.coreVoltage, [Validators.required]],
           frequency: [info.frequency, [Validators.required]],
           autofanspeed: [info.autofanspeed == 1, [Validators.required]],
+          minfanspeed: [info.minFanSpeed, [Validators.required]],
           fanspeed: [info.fanspeed, [Validators.required]],
           temptarget: [info.temptarget, [Validators.required]],
           overheat_mode: [info.overheat_mode, [Validators.required]],
--- a/main/nvs_config.h
+++ b/main/nvs_config.h
@@ -31,7 +31,7 @@
 #define NVS_CONFIG_DISPLAY_TIMEOUT "displayTimeout"
 #define NVS_CONFIG_AUTO_FAN_SPEED "autofanspeed"
 #define NVS_CONFIG_FAN_SPEED "fanspeed"
-#define NVS_CONFIG_MIN_FAN_PCT "minfanpct"
+#define NVS_CONFIG_MIN_FAN_SPEED "minfanspeed"
 #define NVS_CONFIG_TEMP_TARGET "temptarget"
 #define NVS_CONFIG_BEST_DIFF "bestdiff"
 #define NVS_CONFIG_SELF_TEST "selftest"
--- a/main/nvs_config.h
+++ b/main/nvs_config.h
@@ -31,6 +31,7 @@
 #define NVS_CONFIG_DISPLAY_TIMEOUT "displayTimeout"
 #define NVS_CONFIG_AUTO_FAN_SPEED "autofanspeed"
 #define NVS_CONFIG_FAN_SPEED "fanspeed"
+#define NVS_CONFIG_MIN_FAN_PCT "minfanpct"
 #define NVS_CONFIG_TEMP_TARGET "temptarget"
 #define NVS_CONFIG_BEST_DIFF "bestdiff"
 #define NVS_CONFIG_SELF_TEST "selftest"
--- a/main/tasks/power_management_task.c
+++ b/main/tasks/power_management_task.c
@@ -75,9 +75,8 @@ void POWER_MANAGEMENT_task(void * pvParameters)
     
     while (1) {
 
-        // Refresh PID setpoint and minimum output from NVS in case it was changed via API
+        // Refresh PID setpoint from NVS in case it was changed via API
         pid_setPoint = (double)nvs_config_get_u16(NVS_CONFIG_TEMP_TARGET, pid_setPoint);
-        min_fan_pct = (double)nvs_config_get_u16(NVS_CONFIG_MIN_FAN_SPEED, min_fan_pct);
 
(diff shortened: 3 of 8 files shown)


mutatrum:
I think it makes sense to expose this in the AxeOS settings. There's a double-sided slider we can use, so both min and max can be changed. https://primeng.org/slider#range. I started with this, but stopped as there was a lot of refactoring going on with #1026 and #1050. Should be clear now.

As for naming, it sets the minimum (and possible maximum) of the fan percentage, the fact it's controller by a PID is not relevant for the naming. So `min_fan_pct` / `max_fan_pct` or something like that?

0xf0xx0:
is there any use in making the max speed configurable?

0xf0xx0:
image

i put it behind `?oc` cause i assume for stock fans/heatsinks it may cause issues if the value is too low? also capped at 99, if you want 100 just use manual fan control lul

mutatrum:
IMO it doesn't need behind the OC flag. I can run my devices fine with 0% as minimum. One device I have has a fan that still spins at 3000 rpm at 0.

main/tasks/power_management_task.c
```diff
@@ -57,11 +58,12 @@ void POWER_MANAGEMENT_task(void * pvParameters)
     SystemModule * sys_module = &GLOBAL_STATE->SYSTEM_MODULE;
 
     pid_setPoint = (double)nvs_config_get_u16(NVS_CONFIG_TEMP_TARGET, pid_setPoint);
+    pid_min = (double)nvs_config_get_u16(NVS_CONFIG_MIN_FAN_PCT, pid_min);
```
mutatrum: Not a fan (hehehe) of changing the naming convention here, going from `MIN_FAN_PCT` in nvs to `pid_min`.

I would keep it min_fan_pct here as well, this makes searching for usages in the code a lot easier.

0xf0xx0:
squashed down to one commit and removed from behind the oc flag

main/tasks/power_management_task.c
0xf0xx0: renamed uwu

mutatrum:
> squashed down to one commit

Squashing down is not needed, that will happen on the merge. It's preferable not to, as it disconnects previous review comments.

main/nvs_config.h
```diff
@@ -27,6 +27,7 @@
 #define NVS_CONFIG_DISPLAY_TIMEOUT "displayTimeout"
 #define NVS_CONFIG_AUTO_FAN_SPEED "autofanspeed"
 #define NVS_CONFIG_FAN_SPEED "fanspeed"
+#define NVS_CONFIG_MIN_FAN_PCT "minfanpct"
```
mutatrum: `NVS_CONFIG_MIN_FAN_SPEED "minfanspeed"` to keep NVS in line with the rest.

I think it's fine to call it that, it's clear that it's in percentages.

0xf0xx0: done uwu

sz4bi:
Can you please add options to modify PWM parameters also? The default values ramping up and down the fans too rapidly,.

main/http_server/axe-os/src/app/components/edit/edit.component.ts
```diff
@@ -147,6 +147,7 @@ export class EditComponent implements OnInit, OnDestroy {
           coreVoltage: [info.coreVoltage, [Validators.required]],
           frequency: [info.frequency, [Validators.required]],
           autofanspeed: [info.autofanspeed == 1, [Validators.required]],
+          minfanspeed: [info.minFanSpeed, [Validators.required]],
```
duckaxe, line 165: It needs to be added to [noRestartFields](https://github.com/bitaxeorg/ESP-Miner/blob/3a163b60c6ea331307ec9361d04b518076056a75/main/http_server/axe-os/src/app/components/edit/edit.component.ts#L315).

0xf0xx0: it requires a restart to fully apply, updating it via the api just updates the nvs entry  
its not re-read like `tempTarget` is

duckaxe: Ah, ok.

duckAxe (approved):

duckaxe:
Some conflicts need to be resolved, @0xf0xx0

0xf0xx0:
resolved uwu

main/tasks/power_management_task.c
```diff
@@ -31,6 +31,7 @@ static const char * TAG = "power_management";
 
 double pid_input = 0.0;
 double pid_output = 0.0;
+double min_fan_pct = 25.0; // Default, 0-100%, will be overwritten by NVS
```
WantClue: remove comments

main/tasks/power_management_task.c
```diff
@@ -57,11 +58,12 @@ void POWER_MANAGEMENT_task(void * pvParameters)
     SystemModule * sys_module = &GLOBAL_STATE->SYSTEM_MODULE;
 
     pid_setPoint = (double)nvs_config_get_u16(NVS_CONFIG_TEMP_TARGET, pid_setPoint);
+    min_fan_pct = (double)nvs_config_get_u16(NVS_CONFIG_MIN_FAN_SPEED, min_fan_pct);
 
     // Initialize PID controller with pid_d_startup and PID_REVERSE directly
     pid_init(&pid, &pid_input, &pid_output, &pid_setPoint, pid_p, pid_i, pid_d_startup, PID_P_ON_E, PID_REVERSE);
     pid_set_sample_time(&pid, POLL_RATE - 1); // Sample time in ms
-    pid_set_output_limits(&pid, 25, 100); // Output limits 25% to 100%
+    pid_set_output_limits(&pid, min_fan_pct, 100); // Output limits fan_min% to 100%
```
WantClue: remove comments

kenz4284 (approved):

Issue #911, opened 2025-05-12
Stop mining if pools are not reachable


mutatrum:
We had an internet outage, and the result was the following:

![Image](https://github.com/user-attachments/assets/6527f9d1-88e3-4ec5-9574-beedfcdddd09)

In this case it would be prudent to stop mining, as it doesn't make sense to keep hashing the same work over and over again.

skot:
I'd love to get this one fixed soon. I made a "status task" issue a while back. 

My thought is that we have high priority task, running on a long interval that check to make sure mining is happening; stratum server is still sending work, ASIC is still sending shares, vreg is still enabled, etc.

If anything is amiss we should, at the very least show the alert in AxeOS and Bitaxe display.

mutatrum:
#272

RealHumanWithHair:
> We had an internet outage, and the result was the following:
> 
> ![Image](https://github.com/user-attachments/assets/6527f9d1-88e3-4ec5-9574-beedfcdddd09)
> 
> In this case it would be prudent to stop mining, as it doesn't make sense to keep hashing the same work over and over again.

From where did you buy your bitaxe?

RealHumanWithHair:
Should be nice if there have some kind a method with time limit per task, something like 1-2min even 5min to be safe.

WantClue:
Is planned for release 2.13.0

0xf0xx0:
closed by #1689

PR #1063, opened 2025-06-23
Add stratum response time

### LINKED ISSUE #921: Add Pool Latency Indicator to Shares Subsection in AxeOS Dashboard
**Description**
This feature request is to add a Pool Latency indicator to the Shares subsection of the AxeOS Dashboard. The goal is to calculate the latency between share submissions ("stratum_api: tx:" for mining.submit) and their corresponding responses ("stratum_task: rx:" for result accepted). Since the web GUI logs do not include timestamps, the firmware must perform these calculations internally.

**Rationale**
Latency between submitting shares and receiving pool responses is critical for mining efficiency. High latency can increase the risk of stale or rejected shares. By providing latency metrics on the dashboard, users can monitor and optimize their network and pool performance. Ping tests measure raw network latency, but even when the server allows Ping tests they don’t account for protocol overhead or pool-specific processing delays.

**Proposed Dashboard Display**
The latency metrics should be displayed in the Shares subsection as:
    [count] Pool Latency ([last share ms]) ([avg ms])

[count]: Number of latency calculations in the last 10 minutes.
[last share ms]: Latency of the most recent share (in milliseconds).
[avg ms]: Average latency over the last 10 minutes (in milliseconds).

Examples:
    1 Pool Latency (32ms) (32ms avg) (first calculation)
    123 Pool Latency (42ms) (38ms avg) (after multiple calculations)

**Implementation Approach**
The firmware could be modified in the following ways:

Capture Submission Time: When a share is submitted ("stratum_api: tx:"), record the current timestamp (e.g., using millis()).
Capture Response Time: When a response is received ("stratum_task: rx:"), record its timestamp and calculate latency by subtracting the submission timestamp.
Match Events: Use a unique identifier (e.g., job ID or share hash from the Stratum protocol) to pair each submission with its response.
Store Data: Maintain a rolling 10-minute window of latency values in a memory-efficient structure (e.g., a queue).
Calculate Metrics:
Count of latency calculations in the last 10 minutes.
Latency of the most recent share.
Average latency over the last 10 minutes.

Update Dashboard: Expose these metrics to the AxeOS Dashboard for display in the Shares subsection.

**Technical Considerations**
Timestamp Precision: Use millis() for millisecond precision, which should be sufficient for latency tracking.
Memory Management: Implement a queue or circular buffer to store latency data, ensuring it fits within the ESP32’s resource constraints.
Event Matching: Accurately match "tx" and "rx" events using Stratum protocol identifiers.
Edge Cases:
Handle missing responses with a timeout mechanism.
Decide whether to include latency for rejected shares (recommendation: calculate for accepted shares only).

**Additional Suggestions**
Configurable Window: Allow users to adjust the time window (e.g., 5, 10, 15 minutes) via the dashboard.
API Endpoint: Optionally provide latency metrics through an API for external use.

**Questions for Discussion**
Should latency be calculated only for accepted shares, or all shares?
Does the proposed display format suit the dashboard’s layout?

This feature will enhance the AxeOS Dashboard by providing actionable latency insights directly to users. Let’s discuss how to proceed with the implementation!

WantClue:
Instead of creating a ping towards the stratum server we calculate how long it takes the stratum server to respond to our submission. This allows us to play nicely with other issues and resolve them, like #911

This could also be used as the measure towards the connection with a pool in general, having a ping is using ICMP which might be (and already is on some pools) blocked

closes #921

### DIFF
--- a/main/tasks/stratum_task.c
+++ b/main/tasks/stratum_task.c
@@ -196,7 +196,6 @@ void stratum_task(void * pvParameters)
     int retry_attempts = 0;
     int retry_critical_attempts = 0;
 
-    struct timeval current_time;
 
     xTaskCreate(stratum_primary_heartbeat, "stratum primary heartbeat", 8192, pvParameters, 1, NULL);
 


mutatrum:
Did you see #921?

Not sure if this solution works, there's no correlation between the sending (e.g. start of measurement) and receiving (end of measurement). Best we can do in the current one is use the mining.submit, as that has an id field so send and receive times can be correlated, but as skot said in 921, a pool might take it's time to process the submit.

As for time measurement, it's probably simpler to use `esp_timer_get_time`, instead of date/time functions.

Which pools have blocked ICMP ping?

components/stratum/stratum_api.c
```diff
@@ -375,6 +378,9 @@ int STRATUM_V1_configure_version_rolling(int socket, int send_uid, uint32_t * ve
 
 static void debug_stratum_tx(const char * msg)
 {
+    // Record the timestamp when sending
+    gettimeofday(&last_tx_time, NULL);
```
johnny9: It is more appropriate to use the monotonic clock for this. using the timeofday can get skewed if it adjusts under you for whatever reason. it should be easier to use too. https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/esp_timer.html#_CPPv418esp_timer_get_timev

johnny9, line 382: its more appropriate to use a monotonic clock for this instead of wall time. https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/esp_timer.html#_CPPv418esp_timer_get_timev

main/tasks/stratum_task.c
```diff
@@ -309,6 +314,15 @@ void stratum_task(void * pvParameters)
                 break;
             }
 
+            // Calculate response time
+            if (is_tracking_response) {
+                gettimeofday(&current_time, NULL);
```
johnny9: I think it might be best to move the implementation here either all into the stratum_api.c or another timer.h header that has static inline header only utility definitions.

```
/* stratum_api.c */

#include "esp_timer.h" 

static int64_t last_tx_us = 0;
static bool    tracking   = false;

/* call from debug_stratum_tx() */
void STRATUM_V1_stamp_tx(void)
{
    last_tx_us = esp_timer_get_time(); 
    tracking   = true;
}

/* call from the reader thread */
double STRATUM_V1_get_response_time_ms(void)
/* returns -1 if no request was outstanding */
{
    if (!tracking) return -1;

    int64_t delta_us = esp_timer_get_time() - last_tx_us;
    tracking = false;
    return delta_us / 1000.0;            /* convert to ms */
}
```
something like this maybe and call STRATUM_V1_get_response_time_ms

That way we don't have new variables across modules

johnny9 (commented):
Concept ACK

I think this is a good thing to measure and I imagine this is what user are interested in knowing. We likely need to actually track request/response ids for this to be correct, though.

johnny9:
> Which pools have blocked ICMP ping?

apparently ocean does

johnny9:
> Not sure if this solution works, there's no correlation between the sending (e.g. start of measurement) and receiving (end of measurement). Best we can do in the current one is use the mining.submit, as that has an id field so send and receive times can be correlated, but as skot said in 921, a pool might take it's time to process the submit.
>

This is right but wouldn't this be what we're interested in as a miner? submission, processing, and response. 


> As for time measurement, it's probably simpler to use `esp_timer_get_time`, instead of date/time functions.

agree with this

WantClue:
> Did you see #921?
> 
> 
> 
> Not sure if this solution works, there's no correlation between the sending (e.g. start of measurement) and receiving (end of measurement). Best we can do in the current one is use the mining.submit, as that has an id field so send and receive times can be correlated, but as skot said in 921, a pool might take it's time to process the submit.
> 
> 
> 
> As for time measurement, it's probably simpler to use `esp_timer_get_time`, instead of date/time functions.
> 
> 
> 
> Which pools have blocked ICMP ping?

ACK,
I'll refactor some of the code. And add an avg to it.

main/tasks/stratum_task.c
WantClue: Done

components/stratum/stratum_api.c
WantClue: Done

WantClue:
Now we can call it with `STRATUM_V1_get_response_time_ms` 
I'll also add it to the dashboard

WantClue:
image

duckAxe:
@WantClue I would keep url, port and user together. Ping at the end.

components/stratum/stratum_api.c
```diff
@@ -19,6 +20,26 @@ static const char * TAG = "stratum_api";
 static char * json_rpc_buffer = NULL;
 static size_t json_rpc_buffer_size = 0;
 
+static int64_t last_tx_us = 0;
+static bool tracking = false;
```
mutatrum: If you make this an int (or whatever the `id` field is), you can track the actual tx/rx id. There is no guarantee they come in order, e.g. it can happen multiple submits are sent before a rx comes back.

WantClue: True totally forgot about that, should make sure that this is sticking to the `id` field

components/stratum/stratum_api.c
```diff
@@ -375,6 +396,8 @@ int STRATUM_V1_configure_version_rolling(int socket, int send_uid, uint32_t * ve
 
 static void debug_stratum_tx(const char * msg)
 {
+    // Record the timestamp when sending
```
johnny9: Comments like this aren't necessary. The function name is clear enough.

main/tasks/stratum_task.c
```diff
@@ -309,6 +315,13 @@ void stratum_task(void * pvParameters)
                 break;
             }
 
+            // Calculate response time
```
johnny9: Comment not necessary

components/stratum/stratum_api.c
WantClue: Fixed

components/stratum/stratum_api.c
```diff
@@ -10,15 +10,80 @@
...
+
+static RequestTiming* get_request_timing(int request_id) {
+    if (request_id < 0 || request_id >= MAX_REQUEST_IDS) return NULL;
+    return &request_timings[request_id];
+}
+
+void STRATUM_V1_stamp_tx(const char *msg)
+{
+    init_request_timings();
+    int request_id = get_request_id(msg);
```
johnny9: I think we can avoid having to parse the message again here. Maybe pass in the request id from the function that is creating the request

components/stratum/stratum_api.c
```diff
@@ -10,15 +10,80 @@
...
+            timing->timestamp_us = esp_timer_get_time();
+            timing->tracking = true;
+        }
+    }
+}
+
+double STRATUM_V1_get_response_time_ms(const char *msg)
+{
+    init_request_timings();
+    int request_id = get_request_id(msg);
```
johnny9: I think it should be possible to avoid having to parse the message again here as well. Pass in an id and maybe use stratum_api_v1_message from the STRATUM parse method.

mutatrum:
Agree with johnny9, i seems you're in the wrong place to do the time measurement. And I think it's overkill to track multiple id's. If there's overlap between submits, it's fine to only one use, as it's overwritten immediately anyways.

I would add something like this to `system.c`:
```
static uint16_t current_stratum_message = -1;
static uint64_t current_stratum_message_start_time = 0;

void SYSTEM_stratum_response_time_start(uint16_t id) {
    current_stratum_message_start_time = esp_timer_get_time();
    current_stratum_message = id;
}

void SYSTEM_stratum_response_time_end(uint16_t id)
{
    uint64_t duration = esp_timer_get_time() - current_stratum_message_start_time;
    if (current_stratum_message == id) {
        GLOBAL_STATE->SYSTEM_MODULE.stratum_response_time = duration;
    }
}
```

Then call `SYSTEM_stratum_response_time_start` from `asic_result_task.c`, right before `STRATUM_V1_submit_share`. 
And call   `SYSTEM_stratum_response_time_end` from `stratum_task.c`, right before `SYSTEM_notify_accepted_share`. In both locations `id` is already available.

No locking, but by setting/reading the time before setting/checking the id, I _think_ race conditions should be avoided. Unless the compiler throws execution out of order, not 100% sure here. Otherwise it might need a semaphore around those two blocks.

johnny9:
I think mutatrum's might be better as I think it does end up being a simpler implementation but i don't necessarily have too strong of an opinion. I primarily just don't want to have the stratum message being parsed twice.

WantClue:
> Agree with johnny9, i seems you're in the wrong place to do the time measurement. And I think it's overkill to track multiple id's. If there's overlap between submits, it's fine to only one use, as it's overwritten immediately anyways.
> 
> I would add something like this to `system.c`:
> 
> ```
> static uint16_t current_stratum_message = -1;
> static uint64_t current_stratum_message_start_time = 0;
> 
> void SYSTEM_stratum_response_time_start(uint16_t id) {
>     current_stratum_message_start_time = esp_timer_get_time();
>     current_stratum_message = id;
> }
> 
> void SYSTEM_stratum_response_time_end(uint16_t id)
> {
>     uint64_t duration = esp_timer_get_time() - current_stratum_message_start_time;
>     if (current_stratum_message == id) {
>         GLOBAL_STATE->SYSTEM_MODULE.stratum_response_time = duration;
>     }
> }
> ```
> 
> Then call `SYSTEM_stratum_response_time_start` from `asic_result_task.c`, right before `STRATUM_V1_submit_share`. And call `SYSTEM_stratum_response_time_end` from `stratum_task.c`, right before `SYSTEM_notify_accepted_share`. In both locations `id` is already available.
> 
> No locking, but by setting/reading the time before setting/checking the id, I _think_ race conditions should be avoided. Unless the compiler throws execution out of order, not 100% sure here. Otherwise it might need a semaphore around those two blocks.

I don't think this should go into system.c we should have the measurement in the appropriate stratum code. I removed now the double parsing

components/stratum/stratum_api.c
```diff
@@ -10,14 +10,63 @@
...
+        for (int i = 0; i < MAX_REQUEST_IDS; i++) {
+            request_timings[i].timestamp_us = 0;
+            request_timings[i].tracking = false;
+        }
+        initialized = true;
+    }
+}
+
+static RequestTiming* get_request_timing(int request_id) {
+    if (request_id < 0 || request_id >= MAX_REQUEST_IDS) return NULL;
```
mutatrum: I don't understand, it only does a measurement for the first 1024 message ids?

WantClue: yes that was a bug , it's fixed now
thanks

main/tasks/stratum_task.c
```diff
@@ -193,6 +196,9 @@ void stratum_task(void * pvParameters)
     int retry_attempts = 0;
     int retry_critical_attempts = 0;
 
+    struct timeval current_time;
```
johnny9:
```suggestion
```

main/tasks/stratum_task.c
```diff
@@ -193,6 +196,9 @@ void stratum_task(void * pvParameters)
...
+    double response_time_ms;
```
johnny9:
```suggestion
```

main/tasks/stratum_task.c
johnny9: delete, this is no longer used

main/tasks/stratum_task.c
johnny9: delete, this is no longer used

johnny9 (changes requested):
Some left over static variables need to be removed in stratum_task.c

johnny9 (approved):
tACK

Tested on a 204

PR #1067, opened 2025-06-25
reduce gap between hamburger and logo in AxeOS

0xf0xx0:
old: 
image

new: 
image

doesnt need anything done for the smaller scales

### DIFF
--- a/main/http_server/axe-os/src/app/layout/styles/layout/_topbar.scss
+++ b/main/http_server/axe-os/src/app/layout/styles/layout/_topbar.scss
@@ -6,6 +6,7 @@
     top: 0;
     width: 100%;
     padding: 0 2rem;
+    gap: 1rem;
     background-color: var(--surface-card);
     transition: left $transitionDuration;
     display: flex;


0xf0xx0:
an alternative could be to swap the logo and the hamburger, this might fit better with the expanded ui
image
image

duckaxe:
My 2 cent:

1. Please wait for https://github.com/bitaxeorg/ESP-Miner/pull/1056
2. I wouldn't touch the logo and menu button again. With https://github.com/bitaxeorg/ESP-Miner/pull/1056 they are perfectly tuned. Both of your suggestions are not good, they cause disorder in the design. Things need to be aligned.

Screenshot 2025-06-25 at 18 59 00

0xf0xx0:
with your pr the hamburger is still floating by itself. to me it feels off, especially once the sidebar has been folded. i'll set this as a draft until yours is merged uwu

duckaxe:
> with your pr the hamburger is still floating by itself.

Thats right. The menu button is not part of the logo, no matter where it is located - left or right next to the logo. Your suggestions imply an association between the two elements due to their proximity. That's not how it should be, though. They have nothing to do with each other. That's why it's good that they are far apart. This is not a bug, it's a feature.

0xf0xx0:
while i see that, i think its not an issue? most sites place their logo on the right side of the hamburger, like amazon. and again, it being so far off looks weird when the sidebar is folded. maybe the logo should be place to the right edge and the hamburger on the left, like on the mobile ui?

duckaxe:
@0xf0xx0 I've been thinking about it again. But I don't have a better idea. Personally, I think the current situation is fine. Maybe we should close the PR until there's a better suggestion?

0xf0xx0:
yea i dont have anything either xmx closing

PR #1051, opened 2025-06-19
Accept and store float frequency values

mutatrum:
One of the things I found in #1045 is that frequency values are currently bound to integer values, however, the PLL settings can resolve way more frequencies. By accepting frequency as float, way more values can be used.

Currently it only accepts exact values, a next step could be to make it more lenient.

PR #1069, opened 2025-06-26
Add frequency ramp for BM1397

mutatrum:
Follow-up of #1051, this adds experimental support for PLL calculations for the BM1397. This is untested, and assumes the PLL has a similar structure as the other chips. The [documentation](https://github.com/skot/BM1397/blob/master/registers.md#pll0-parameter) seems to state this.

| PLL parameter | Old name | Old range | New range | Remark |
| --- | --- |:---:|:---:| --- |
| `fb_divider` | `fa` | 60 - 200 | 60 - 200 |
| `refdiv` | `fb` | 1 - 2 | 1 - 2 |
| `postdiv1` | `fc1` | 1 - 3 | 1 - 7 | `postdiv1` > `postdiv2` |
| `postdiv2` | `fc2` | 5 | 1 - 7 |
| Frequency | | 50 - 650 Mhz | 17 $\frac{6}{7}$ - 1500 Mhz | 

Commit f0cbc054818e761b21a0e2f5d56fc87abe0ba6ba.

[BM1397 Frequency analysis.ods](https://github.com/user-attachments/files/20921852/BM1397.Frequency.analysis.ods)


johnny9:
Concept ACK

I think we're going to need some unit tests for these functions at this point. Ideally using known working examples as well as the edge cases to document the expectations of the frequency calculations. It's a pretty clean function so it should be simple enough to add tests around it. So please add something like test_pll.c to our tests for this component

components/asic/test/test_pll.c
```diff
@@ -0,0 +1,18 @@
+#include "unity.h"
+
+#include "pll.h"
+
+TEST_CASE("Check PLL frequency calculation", "[pll]")
```
johnny9, line 5: This is a great start. Add a case using each of the know divider constants and have a clear testcase name that includes the asic that uses those. We should test some of the limits of the function as well. Dump pll.c into your favorite llm and ask it to come up with a list of testcases if you need some ideas to get good coverage

test/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ set(EXTRA_COMPONENT_DIRS "../components")
 # - when invoking CMake directly: cmake -D TEST_COMPONENTS="xxxxx" ..
 # - when using idf.py: idf.py -T xxxxx build
 #
-set(TEST_COMPONENTS "bm1397 stratum" CACHE STRING "List of components to test")
+set(TEST_COMPONENTS "bm1397 stratum asic" CACHE STRING "List of components to test")
```
johnny9: bm1397 doesnt appear to be a component. wonder how this was working

components/asic/test/test_pll.c
mutatrum: Will add more, I first wanted to see if the tests would run.

test/CMakeLists.txt
mutatrum: It used to be a component, until the `bm1397` module got renamed to `asic` and this was overlooked: a7bae3ce7112c1585addd4313edbde3f013cbd6a.

Seems the code of `test_job_command.c` is quite out of date. Not sure if this test can run without a device, so unsure how to proceed from here.

mutatrum: Commented out `test_job_command`, at least `pll_test` is running now:
image
No idea why this doesn't fail the build though.

WantClue:
will do testing on this weekend

WantClue:
Any frequency change seems to drastically reduce the produced hashrate, still investigating

mutatrum:
> Any frequency change seems to drastically reduce the produced hashrate, still investigating

Verified this on latest master + #1318 and it seems to work. The registered hashrate briefly drops to 0 when changing frequency, but it seems to settle around the expected hashrate. Tested several adjustments up and down.

WantClue (approved):
code lgtm testing inc, we might also remove the commented out parts

PR #1071, opened 2025-06-26
Add files via upload

STSMiner1:
Stratum V1 Protocol Test Tool with client.reconnect Trigger

This tool models the Stratum V1 protocol as used for SHA256-based coins (e.g., Bitcoin). It simulates a minimalist pool that demonstrates core protocol behavior, and now includes a "reconnect storm" after 5 job notifications.

### DIFF
--- a/Stratumv1_Protocol_Test_Tool.py
+++ b/Stratumv1_Protocol_Test_Tool.py
@@ -1,0 +1,246 @@
+#!/usr/bin/env python3
+"""
+Stratum V1 Protocol Test Tool with client.reconnect Trigger
+
+This tool models the Stratum V1 protocol as used for SHA256-based coins (e.g., Bitcoin).
+It simulates a minimalist pool that demonstrates core protocol behavior,
+and now includes a "reconnect storm" after 5 job notifications.
+"""
+
+import asyncio
+import json
+import time
+import os
+from datetime import datetime
+
+LOG_FILE = "stratum_log.txt"
+
+def log(msg):
+    timestamp = datetime.now().strftime("[%Y-%m-%d %H:%M:%S]")
+    line = f"{timestamp} {msg}"
+    print(line)
+    with open(LOG_FILE, "a", encoding="utf-8") as f:
+        f.write(line + "\n")
+
+def print_handshake_diagram():
+    diagram = [
+        "        Miner Side                              Pool Side",
+        "        ----------                              ----------",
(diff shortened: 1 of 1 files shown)


Stratumv1_Protocol_Test_Tool.py
```diff
@@ -0,0 +1,265 @@
...
+    log("")
+    print_handshake_diagram()
+
+    server = await asyncio.start_server(handle_client, "0.0.0.0", 3333)
+    log(f"Stratum mock pool listening on {server.sockets[0].getsockname()}")
+    async with server:
+        await server.serve_forever()
+
+if __name__ == "__main__":
+    asyncio.run(main())
```
johnny9, line 265: new line at the end of the file

ghost: Thanks for the thoughtful feedback—makes sense.

I originally placed the script in the root to keep it visible during early development and discussion, but I agree that "contrib/" is a better long-term home for tools like this, feel free to move it there if that helps with organization.

The tool itself is a lightweight Stratum server simulator. I built it to test "mining.extranonce.subscribe" and "mining.set_extranonce" behaviors—just a basic sandbox pool stripped of unnecessary parts. Python made that easy and portable. It helps catch miner edge cases and protocol assumptions that might not show up with a typical pool setup.

Also comes in handy for testing the Bitaxe fallback mechanism.

Stratumv1_Protocol_Test_Tool.py
```diff
@@ -0,0 +1,265 @@
...
+                    "error": None
+                })
+                await asyncio.sleep(0.2)
+                await send_msg(generate_fake_job(format(job_id_counter, "04x")))
+                job_id_counter += 1
+
+                # Push extranonce update if we're in Nicehash mode
+                if personality == "nicehash":
+                    await asyncio.sleep(0.5)
+                    new_extra = random_extranonce(9)
```
adammwest, line 195: why length 9? 
extranonce length would be made with *2 so 18 hex, 18 hex i believe is unsupported in stratum 
i think 8 bytes ~ 16 hex is max for most pools

I think its useful to have this feature but, controlled via command line options

like  `--extranonce2_length x`

ghost:
It was done as a test tool, so should not really matter with that, the tool is a passive test harness, nothing more.

Psst .... there are two version of this test tool here.

PR #1072, opened 2025-06-26
Increased contrast ratio for purple

duckaxe:
The Google Accessibility Tool suggests increasing the contrast of the purple color (current contrast ratio: 2.98, recommended contrast ratio: 4.5) to make the layout easier to read. This PR uses a slightly brighter purple to improve contrast.

**Before**
Screenshot 2025-06-26 at 14 12 56

**After**
Screenshot 2025-06-26 at 14 12 26

### DIFF
--- a/main/http_server/axe-os/src/app/components/design/theme-config.component.ts
+++ b/main/http_server/axe-os/src/app/components/design/theme-config.component.ts
@@ -184,31 +184,31 @@ export class ThemeConfigComponent implements OnInit {
     },
     {
       name: 'Purple',
-      primaryColor: '#9c27b0',
+      primaryColor: '#b340fa',
       accentColors: {
-        '--primary-color': '#9c27b0',
+        '--primary-color': '#b340fa',
         '--primary-color-text': '#ffffff',
-        '--highlight-bg': '#9c27b0',
+        '--highlight-bg': '#b340fa',
         '--highlight-text-color': '#ffffff',
         '--focus-ring': '0 0 0 0.2rem rgba(156,39,176,0.2)',
         // PrimeNG Slider
         '--slider-bg': '#dee2e6',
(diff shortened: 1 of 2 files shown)


main/http_server/axe-os/src/app/components/design/theme-config.component.ts
```diff
@@ -184,31 +184,31 @@ export class ThemeConfigComponent implements OnInit {
...
-        '--button-bg': '#9c27b0',
+        '--button-bg': '#b340fa',
         '--button-hover-bg': '#8e24aa',
-        '--button-focus-shadow': '0 0 0 2px #ffffff, 0 0 0 4px #9c27b0',
+        '--button-focus-shadow': '0 0 0 2px #ffffff, 0 0 0 4px #b340fa',
         // Toggle button
-        '--togglebutton-bg': '#9c27b0',
-        '--togglebutton-border': '1px solid #9c27b0',
+        '--togglebutton-bg': '#b340fa',
+        '--togglebutton-border': '1px solid #b340fa',
```
mutatrum, line 211: dumb question, but is it possible to define the hex color code only once?

duckaxe: If I do what you want, we have to make the same for all colors, not just for purple. I wouldn't do that in this PR, though. Too much changes.

mutatrum: Agree that it should be for all colors then, but I'm not sure if it's a good idea, and if it is, if it can be done in a nice way. Agreed it should be a separate PR.

Thinking of that, you can get really funky with a color picker then :rofl:

mutatrum (approved):
Not sure which mad person uses purple, but now at least it's readable 😆

PR #1075, opened 2025-06-26
Uniform notifications design

duckAxe:
Currently, Toastr notifications are not uniform. This pull request ensures consistency.

1. Notifications title is always `Success` | `Warning` | `Error`
2. Notification message is always the message not the title
3. ~~Bitaxe~~ → Device

### DIFF
--- a/main/http_server/axe-os/src/app/components/edit/edit.component.ts
+++ b/main/http_server/axe-os/src/app/components/edit/edit.component.ts
@@ -213,7 +213,7 @@ export class EditComponent implements OnInit, OnDestroy {
           if (this.isRestartRequired) {
             this.toastr.warning('You must restart this device after saving for changes to take effect', 'Warning');
           }
-          this.toastr.success(successMessage, 'Success!');
+          this.toastr.success(successMessage, 'Success');
           this.savedChanges = true;
         },
         error: (err: HttpErrorResponse) => {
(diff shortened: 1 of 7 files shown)


mutatrum:
It's weird a method called X is called, but also the string X is added. `toastr.success('blabla', 'Success')`. But maybe that's a weird toastr thing.

duckaxe:
The second arg is the title. We don't need this title. I don't know why it was used before. Toaster uses colors for notifications to communicate whether it is a success, error or warning, so we don't need the title. We should just display the message with no frills. What du you mean, @mutatrum 

Screenshot 2025-06-27 at 13 05 45


Screenshot 2025-06-27 at 13 03 19

mutatrum:
> The second arg is the title. We don't need this title. I don't know why it was used before. Toaster uses colors for notifications to communicate whether it is a success, error or warning, so we don't need the title. We should just display the message with no frills. What du you mean, @mutatrum

That makes sense. Colors and icon both convey the status.

And if you're fiddling with it, I would add punctuation to the sentences.

duckaxe:
> That makes sense. Colors and icon both convey the status.

Check.

> And if you're fiddling with it, I would add punctuation to the sentences.

Done. Short messages without punctuation, sentences with punctuation.

Screenshot 2025-06-27 at 14 24 56

mutatrum (approved):

WantClue:
@duckAxe please resolve conflicts for merging

duckaxe:
@WantClue Done.

PR #1089, opened 2025-06-27
Swarm should handle devices without asic endpoint

mutatrum:
This is best effort backwards compatible leniency towards devices that don't have the `/api/system/asic` endpoint yet. It also points to a way to move towards a more organised api which doesn't break the whole world.

API CHANGES: 
- `/api/system/asic`
  - renamed `familyName` to `deviceModel`
  - added `asicCount`
- `/api/system/info`
  - removed `asicCount`

### DIFF
--- a/main/http_server/axe-os/src/models/ISystemASIC.ts
+++ b/main/http_server/axe-os/src/models/ISystemASIC.ts
@@ -1,6 +1,6 @@
 export interface ISystemASIC {
   ASICModel: string;
-  familyName: string;
+  deviceModel: string;
   swarmColor: string;
   asicCount: number;
   defaultFrequency: number;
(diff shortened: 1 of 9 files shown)


main/http_server/axe-os/src/models/ISystemASIC.ts
```diff
@@ -2,6 +2,7 @@ export interface ISystemASIC {
   ASICModel: string;
   familyName: string;
   swarmColor: string;
+  asicCount: number;
```
duckaxe, line 5: Wouldn't it be better to use a unified spelling for ASIC properties? `ASICModel` & `ASICCount` or `asicModel` & `asicCount`? I tend towards the last option.

WantClue: I think the last would be ideal

mutatrum: I would opt to do that in another PR, first I want to get this sorted. Especially since `ASICModel` is an existing property.

WantClue (approved):

PR #1098, opened 2025-06-29
The Great Rename

mutatrum:
The `family` nomenclature was never a good fit. When renaming `familyName` to `deviceModel`, the naming for hardware configuration finally made sense:

Instead of:
- Device -> Family -> Asic

It should be:
- Board -> Device -> Asic

Board is the specific hardware revision, e.g. 204, 601, etc, which sensors are used, etc.
Device is Max, Ultra, Gamma, etc, with number of chips and some other things
Asic is the actual chip, number of cores in the chip, etc.

Then the `deviceName` field as part of `DeviceConfig` suddenly makes a lot of sense, as does `boardVersion` as part of `BoardConfig`.

This changeset mainly does a lot of renaming. Other subtle changes are passing `BoardConfig` instead of `GlobalState` if possible, and pass them all as pointer, instead of pass by value. Also fixed some capitalization of GLOBAL_STATE fields.


components/asic/asic.c
```diff
@@ -62,27 +62,9 @@ int ASIC_set_max_baud(GlobalState * GLOBAL_STATE)
     return 0;
 }
 
-void ASIC_set_job_difficulty_mask(GlobalState * GLOBAL_STATE, uint8_t mask)
```
mutatrum, line 65: This was only used in self_test. By setting the desired difficulty before `ASIC_init` is called, this is no longer needed.

main/board_config.c
```diff
@@ -0,0 +1,79 @@
+#include <string.h>
```
mutatrum, line 1: Unfortunately, GitHub refuses to show this as a rename. To see the diff, look at the first commit (a3bf3d580df77e19e0207d3ff9b802ca7d2fb0d6), the 2nd commit is just the rename of this file.

johnny9:
I personally wouldn't bother with this name change. I don't think Device/Board add any clarity to the configuration situation and are interchangeable names. These sort of names are largely up to us to assign meaning to them and Family made just as much sense in my opinion.  But I think it's mostly because we just have a static and mostly flat relationship between all of these concepts.

I like all of the other cleanup though so I'd probably merge it just for that reason.

mutatrum:
Drifted too far away. Might redo this at some point.

PR #1100, opened 2025-06-30
Preset Stratum Pools

duckaxe:
This PR introduces a set of predefined pools, which makes it easier for users to switch them. Of course, users still have the option to add a custom pool. 

The pool list comes from [quicklink.service.ts](https://github.com/bitaxeorg/ESP-Miner/blob/master/main/http_server/axe-os/src/app/services/quicklink.service.ts) (`solo.d-central.tech` is dead, added `Blitz Pool` instead). The list is sorted by name. We can adjust the list at any time.

https://github.com/user-attachments/assets/d0f97856-1944-405f-91e7-0eb7a57160b7

### DIFF
--- a/main/http_server/axe-os/src/app/components/pool/pool.component.ts
+++ b/main/http_server/axe-os/src/app/components/pool/pool.component.ts
@@ -5,9 +5,9 @@ import { ToastrService } from 'ngx-toastr';
 import { startWith, takeUntil, Subject } from 'rxjs';
 import { LoadingService } from 'src/app/services/loading.service';
 import { SystemService } from 'src/app/services/system.service';
+import { PRESET_POOLS, PresetPool } from 'src/app/constants/preset-pools';
 
 type PoolType = 'stratum' | 'fallbackStratum';
-type PresetPool = { name: string, value: { url: string, port: number } | {} };
 
 @Component({
   selector: 'app-pool',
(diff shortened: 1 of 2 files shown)


johnny9:
I don't think we should hide the url and port with a predefined pool. I think we should keep it clear to the user what a connection to a pool means and so they can learn these basic concepts. Maybe just lock the inputs if a predefined pool is selected.

duckaxe:
> I don't think we should hide the url and port with a predefined pool. I think we should keep it clear to the user what a connection to a pool means and so they can learn these basic concepts. Maybe just lock the inputs if a predefined pool is selected.

Good idea. Code changed. PR description updated.

remcoros:
maybe the list of known pools can be extracted to a separate file and imported, and used here too: https://github.com/bitaxeorg/ESP-Miner/blob/master/main/http_server/axe-os/src/app/services/quicklink.service.ts to avoid duplication or miss future additions

skot:
I'm conflicted about this feature. On one hand just choosing a pool from the list does make life easier for new miners. On the other hand this list could quickly get unmanageable (both in length and drama) for esp-miner maintainers.

xr1140:
The benefit of such list would be if the user ca populate the list with his custom pools.

duckaxe:
> maybe the list of known pools can be extracted to a separate file and imported, and used here too: 

Done.

> I'm conflicted about this feature. On one hand just choosing a pool from the list does make life easier for new miners. On the other hand this list could quickly get unmanageable (both in length and drama) for esp-miner maintainers.

Before implementing the feature, I considered beginners who could easily set up a (fallback) pool without knowing what a port is. Since most solo pools use the same format for the username and password, it's easy for users to switch pools. They don't need to search for the pool URL and port; they can just use our new dropdown menu.

We have to maintain [the pools](https://github.com/bitaxeorg/ESP-Miner/blob/master/main/http_server/axe-os/src/app/services/quicklink.service.ts) anyway. We just need to update [one more file](https://github.com/bitaxeorg/ESP-Miner/blob/cf6e31b59f297a5025d807665ae8ea7f4d2bb62c/main/http_server/axe-os/src/app/constants/preset-pools.ts).

> The benefit of such list would be if the user ca populate the list with his custom pools.

That's a huge construction site because we need to implement a whole management to add, edit & remove custom pools.

duckaxe:
I close the PR. We can take it out at a later time and build pool management on it.

PR #1102, opened 2025-06-30
Common difficulty mask function

mutatrum:
Extracted some cleanups from #1098

Extract `BM##_set_job_difficulty_mask` into a single function
Set difficulty in self-test before ASIC_init to eliminate `ASIC_set_job_difficulty_mask`
Pass `DEVICE_CONFIG` by pointer instead of by value in `thermal.c`
Fixed some capitalisation of `GLOBAL_STATE` and some others

### DIFF
--- a/components/asic/asic.c
+++ b/components/asic/asic.c
@@ -62,24 +62,6 @@ int ASIC_set_max_baud(GlobalState * GLOBAL_STATE)
     return 0;
 }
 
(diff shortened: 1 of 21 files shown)


johnny9:
I know it might feel arbitrary to start requesting this all of a sudden but we really need to start getting unit test coverage on our core components. I would like to request some basic unit tests are added at the very least around the new common function. Doesn't have to necessarily be with this PR, though.

johnny9 (approved):

PR #1106, opened 2025-07-01
Fix strncpy buffer overflow warning in WiFi configuration

ahmedalalousi:
## Description
This PR fixes a compilation error when building with `-Werror=stringop-truncation`. Also fixes config-custom.cvs file so that custom boards with BM1368 ASICs are recognised as such.

## Problem
- The current code uses `strncpy` without ensuring null termination, which triggers:
error: 'strncpy' specified bound 32 equals destination size [-Werror=stringop-truncation]

## Solution
- Reduced copy length by 1 to leave room for null terminator
- Explicitly set null terminator at the end of the buffer
- Ensures SSID string is always properly terminated

## Testing
- Built successfully with ESP-IDF v5.x
- Tested WiFi AP functionality - working correctly
- No strncpy compilation errors

## Problem
If you want to use a custom board, with one or more BM1368 ASICs, then using the default config-custom.cvs configuration file will simply ignore the asicmodel string setting.

## Solution
- To use BM1368-based custom boards, ensure to use a unique string that's not already being used. Otherwise, the default "207" string in the example will default to BM1366, no matter what you user for the asicmodel string
 
## Testing
- New example custom configuration file created config-custom-1368.cvs
- Compiled and flashed image
- Interface now recognises the new custom board, with BM1368


johnny9 (approved):
ACK

Changes look good. Be sure to squash when merging as the cleanup commits aren't needed.

johnny9:
Create another PR for the custom board csv is you would like that fix as well.

PR #1103, opened 2025-06-30
Pool difficulty column on swarm

mutatrum:
Fixes #1093

This (again) renames `stratumDifficulty` to `poolDifficulty` to align with NerdAxe. Swarm display has fallback code to use `stratumDiff` for older Bitaxe devices.

As the previous rename from `stratumDiff` to `stratumDifficulty` was done in #1064 and is also part of 2.9.0, it would be good to get this in as well, to reduce fallback code.

---
API CHANGE: `apy/system/info`: `stratumDiff`/`stratumDifficulty` is renamed to `poolDifficulty`.

PR #1107, opened 2025-07-01
FUP #1103: Fix sorting for undefined poolDifficulty

duckAxe:
If a device does not provide the `poolDifficulty`, the swarm column "Pool Diff" will not sort properly and will cause an error. The fix adds a check for sorting columns/fields, skipping type === undefined.

Screenshot 2025-07-01 at 19 56 27


mutatrum:
For this to work with old devices that still have `stratumDiff`, I moved the fallback code from the html to the ts:
https://github.com/mutatrum/ESP-Miner/commit/9c647646225e3adde5dd290efca2bc8a40d058e7

duckaxe:
Merged, thanks.

mutatrum (approved):

PR #1108, opened 2025-07-01
Add workflow for building and publishing our devcontainer

johnny9:
Splits the devcontainer Dockerfile into 2 stages, one for the base esp-idf tooling and a second that includes qemu support. This reduces the size of the base Dockerfile while keeping our devcontainer functionality.

Each stage is published as a container to ghcr.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,23 +1,43 @@
-FROM espressif/idf:v5.4.1
+# Stage 1: Base image with ESP-IDF and common dependencies
+FROM espressif/idf:v5.4.1 AS base
 
 ARG DEBIAN_FRONTEND=nointeractive
 
+# Install common dependencies
 RUN apt-get update \
   && apt install -y -q \
   cmake \
   git \
   hwdata \
-  libglib2.0-0 \
+  libglib2.0-dev \
   libnuma1 \
-  libpixman-1-0 \
+  libpixman-1-dev \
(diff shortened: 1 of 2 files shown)


johnny9:
Here is a successful instance from my fork. https://github.com/johnny9/ESP-Miner/actions/runs/16010388987/job/45166922414

The base image from my fork can be pulled with `docker pull ghcr.io/johnny9/esp-miner/devcontainer:latest`. In our repo the image location will be `ghcr.io/bitaxeorg/esp-miner/devcontainer:latest`

This is a step towards having more reproducible builds integrated directly into our repo by having a base docker image that we can all  build our images against. They will be archived in in ghcr and images will be tagged with the git sha so we can map the container image with a commit in git.

.devcontainer/Dockerfile
```diff
@@ -1,42 +1,53 @@
...
 
+# Set up usbip for QEMU USB/IP functionality
 RUN update-alternatives --install /usr/local/bin/usbip usbip `ls /usr/lib/linux-tools/*/usbip | tail -n1` 20
 
-# Node.js and npm (v20 is required)
-RUN curl -fsSL https://deb.nodesource.com/setup_22.x | bash -
-RUN apt-get update && apt-get install -y nodejs
-
 # QEMU
 ENV QEMU_REL=esp-develop-20220919
```
eandersson, line 43: We should bump this to esp-develop-9.2.2-20250228

PR #1110, opened 2025-07-02
Handle multiple websocket connections

mutatrum:
Currently, a new logs websocket connection overwrites an existing connection. This PR adds support for 10 concurrent websocket connections for logs, as well as proper disconnect handling. Websocket code is extracted from `http_server.c` out into a separate `websocket.c` file.


mutatrum:
I can open 10 tabs with logs all streaming. 11th tab gives an error, and if I close any of the previous 10, the 11th can reload the logs screen and start streaming.

However, when websocket connections are exhausted, best I could do is emit a generic toaster error. I think the 101 Upgrade is sent before the 429, so the browser initially thinks the websocket is good, but then receives an invalid frame. At least an error is shown, so I think it's good enough.

johnny9:
Ok it appears to be working correctly for me too.

Is it better to kill old connections if we hit the limit rather than stopping the new connection?

main/http_server/websocket.c
```diff
@@ -0,0 +1,232 @@
...
+#include "http_server.h"
+
+static const char * TAG = "websocket";
+
+static QueueHandle_t log_queue = NULL;
+static int clients[MAX_WEBSOCKET_CLIENTS];
+
+int log_to_queue(const char *format, va_list args)
+{
+    va_list args_copy;
```
johnny9, line 24: Add a NULL check for log_queue at the top of the function to be safe

main/http_server/websocket.c
```diff
@@ -0,0 +1,232 @@
...
+
+static esp_err_t add_client(int fd)
+{
+    for (int i = 0; i < MAX_WEBSOCKET_CLIENTS; i++) {
+        if (clients[i] == -1) {
+            if (is_empty()) {
+                esp_log_set_vprintf(log_to_queue);
+            }
+
+            clients[i] = fd;
```
johnny9, line 76: might need a mutex to guard clients

main/http_server/websocket.c
```diff
@@ -0,0 +1,232 @@
...
+        free((void*)log_buffer);
+    }
+
+    return 0;
+}
+
+static void init_clients(void)
+{
+    for (int i = 0; i < MAX_WEBSOCKET_CLIENTS; i++) {
+        clients[i] = -1;
```
johnny9: safer to initialize clients statically `static int clients[MAX_WEBSOCKET_CLIENTS] = { [0 ... MAX_WEBSOCKET_CLIENTS-1] = -1 };`

main/http_server/axe-os/src/app/components/logs/logs.component.ts
```diff
@@ -78,6 +80,9 @@ export class LogsComponent implements OnInit, OnDestroy, AfterViewChecked {
           if (this.logs.length > 256) {
             this.logs.shift();
           }
+        },
+        error: (error) => {
+          this.toastr.error("Error opening websocket connection");
```
johnny9, line 85: I'm not sure why but this toast never appeared on my tabs when i opened too many

mutatrum:
Not sure what happened with the force-push? Was that to merge in master, as that is better done by the author. This disrupts the local checkout of the branch, in case I want to add stuff to it.

If there are no conflicts at the time of merge, it's not even needed to update the branch, GitHub will squash and merge cleanly.

main/http_server/axe-os/src/app/components/logs/logs.component.ts
mutatrum: This is on Chrome:
image

main/http_server/websocket.c
mutatrum: I understand the defensiveness, but if it either is NULL we cannot do anything sane here anyways, we cannot even log an error log, as we're inside the log handler.

main/http_server/websocket.c
mutatrum: Mutex added.

johnny9:
> Not sure what happened with the force-push? Was that to merge in master, as that is better done by the author. This disrupts the local checkout of the branch, in case I want to add stuff to it.
> 
> If there are no conflicts at the time of merge, it's not even needed to update the branch, GitHub will squash and merge cleanly.

Yeah I hit the rebase button. Won't do that again

main/http_server/websocket.c
johnny9: what will xQueueSendToBack do though?

johnny9: xQueueSendToBack doesnt appear to have NULL protections so if it is null this will cause a fault.
If log_queue is null just return.

main/http_server/axe-os/src/app/components/logs/logs.component.ts
johnny9: This is working well for me now. I needed to clean up my build config for the new 26 sockets limit to work.

johnny9 (approved):
tACK

Tested on a 204 with chrome and firefox by opening 10 tabs and seeing that the 11th failed with the toast appearing.

main/http_server/axe-os/src/app/components/logs/logs.component.ts
mutatrum: Yeah, every time `sdkconfig.defaults` changes, you need a full clean, or at the least delete `sdkconfig`. Not sure why the build process doesn't pick this up.

main/http_server/websocket.c
mutatrum: Added NULL check on queue creation

WantClue (approved):

PR #1111, opened 2025-07-02
Auto refresh current version @ update view

duckaxe:
The `update` page does not reload after the firmware update (AxeOS update does). This means that the display of the current version does not change after the firmware update. This PR makes sure that the version is checked in the background and refreshed on update.

Screenshot 2025-07-02 at 22 30 25

### DIFF
--- a/main/http_server/axe-os/src/app/components/update/update.component.ts
+++ b/main/http_server/axe-os/src/app/components/update/update.component.ts
@@ -1,5 +1,5 @@
 import { Component, ViewChild } from '@angular/core';
-import { Observable, switchMap, shareReplay, map, timer } from 'rxjs';
+import { Observable, switchMap, shareReplay, map, timer, distinctUntilChanged } from 'rxjs';
 import { HttpErrorResponse, HttpEventType } from '@angular/common/http';
 import { ToastrService } from 'ngx-toastr';
 import { FileUploadHandlerEvent, FileUpload } from 'primeng/fileupload';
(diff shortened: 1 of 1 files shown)


main/http_server/axe-os/src/app/components/update/update.component.ts
```diff
@@ -38,7 +38,10 @@ export class UpdateComponent {
       return releases[0];
     }));
 
-    this.info$ = this.systemService.getInfo().pipe(shareReplay({refCount: true, bufferSize: 1}))
+    this.info$ = timer(0, 5000).pipe(
```
mutatrum, line 41: Why isn't the page reload on L104 enough to trigger this? And can the reload of `info` be done in that code-block?

duckaxe: Line 104?

duckaxe: Resolve conversation after Discord discussion.

johnny9:
Is it accurate to do this?

main/http_server/axe-os/src/app/components/update/update.component.ts
johnny9: what was the resolution?

duckaxe: My code modification is necessary because after the firmware update the ESP restarts but the webpage does not reload. Therefore, we need my changes so that the new firmware version can be updated.

Issue #1109, opened 2025-07-02
Support unit tests in forked PRs


mutatrum:
Currently, unit tests on PRs are run, but the results are not posted to GitHub. If we want to increase test coverage, we need to support this.

See: https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#support-fork-repositories-and-dependabot-branches

PR #1113, opened 2025-07-03
Split unittest workflow up it can work with pull requests

johnny9:
Due to the limited permissions on pull requests, the unittest publisher should be run in its own workflow.

https://github.com/EnricoMi/publish-unit-test-result-action/tree/v2.0.0?tab=readme-ov-file#support-fork-repositories-and-dependabot-branches

https://github.com/bitaxeorg/ESP-Miner/issues/1109

### DIFF
--- a/.github/workflows/unittest-results.yml
+++ b/.github/workflows/unittest-results.yml
@@ -1,0 +1,45 @@
+name: Publish Test Results
+
+on:
+  workflow_run:
+    workflows: ["Unit Test"]
+    types:
+      - completed
+permissions: {}
+
+jobs:
+  test-results:
+    name: Unit Test Results
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion != 'skipped'
+
+    permissions:
+      checks: write
+      pull-requests: write
+      contents: read
+      issues: read
+      actions: read
+
+    steps:
+      - name: Download and Extract Artifacts
+        env:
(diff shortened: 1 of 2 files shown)


mutatrum:
Can you temporarily add a failing test, to see how it's being picked up by GitHub?

.github/workflows/unittest-results.yml
```diff
@@ -0,0 +1,45 @@
+name: Publish Test Results
```
eandersson, line 1: Maybe Publish Unit Test Results just for added clarity

.github/workflows/unittest.yml
```diff
@@ -2,33 +2,48 @@ name: Unit Test
...
+
+      - name: Upload event file
+        uses: actions/upload-artifact@v4
+        with:
+          name: event-file
+          path: ${{ github.event_path }}
+
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@v4
```
mutatrum, line 37: If there is a test failure in the `report.xml`, doesn't GitHub then automatically flag the build as failed? I thought GitHub can read that format.

johnny9: The publish action was doing that previously but the publish action doesn't work on pull requests.

.github/workflows/unittest-results.yml
johnny9: Yeah i flip-flopped on this. It can technically be used for any of our test scenarios that output junit xml. likely wont for a little while anyway

PR #1120, opened 2025-07-04
Introduce dockerbuild.py

johnny9:
dockerbuild.py is a python3 script that will launch our dev container and run the build commands for our project. This should make it easier for people to building the latest firmware.

### DIFF
--- a/dockerbuild.py
+++ b/dockerbuild.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python3
+"""dockerbuild.py - helper to run ESP-Miner builds inside a Docker/Podman container.
+
+Prerequisites (choose one):
+  • Podman (rootless-friendly OCI runtime).
+    Ubuntu/Debian:  `sudo apt install podman`
+    Arch Linux   :  `sudo pacman -S podman`
+    macOS (brew) :  `brew install podman`
+  • Docker (classic container runtime).
+    Ubuntu/Debian:  `sudo apt install docker.io`
+    Arch Linux   :  `sudo pacman -S docker`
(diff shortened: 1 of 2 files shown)


eandersson:
Looks good. Should we maybe put all scripts into something like a contrib or tools folder?

dockerbuild.py
```diff
@@ -0,0 +1,94 @@
...
+        workspace,
+        image,
+    ]
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Run build inside Docker/Podman container")
+    parser.add_argument("build_cmd", nargs=argparse.REMAINDER, help="Build command to run (default: idf.py build)")
+    parser.add_argument("--image", help="Container image to use",
+                    default="ghcr.io/bitaxeorg/esp-miner/devcontainer:sha-6a7c499a5dd8f985a578a05e04eba3fa9f93f1f7")
```
eandersson: Why don't we use something like `latest`, `stable` or a specific tag / version by default?

johnny9: Problem with using latest is that if you check out an older commit, latest might not work for it anymore.

johnny9: Version string might work if it's specific enough

johnny9:
> Looks good. Should we maybe put all scripts into something like a contrib or tools folder?

Yeah I think that makes sense. I can start that trend with this script and the rest can be migrated over separately. I think tools is a good folder for these things that will be necessary for either deployment and proper builds. The goal with this script is to be able to have reproducible builds.

WantClue (approved):

PR #1128, opened 2025-07-05
feat: Add basic voltage monitoring implementation

ahmedalalousi:
## Problem Description ##
When operating in series-chained ASICs, voltage discrepancy per ASIC will occur

## Solution ##
Implement chain- and ASIC-level voltage monitoring in hardware, and the associated software update

## Solution Description ##

1. A voltage follower arrangement is implemented around the cost effective and widely available LM358 dual OpAmp, configured as a voltage follower. This will effectively act as a voltage sensor to read each ASIC's (the BM1368 in this case) voltage, with reference to its ground. There are half as many LM358 ICs as there are ASICs, with each LM358 responsible for reading the voltage on two ASICs
2. 1× CD74HC4067 (16:1 mux) cascaded with 1× CD74HC4051 (8:1 mux) will allow us to feed measurements from a chain of up to 40 ASICs into an ADC
3. The ADS1115, a low cost 16-bit ADC with I2C interface, will then feed these voltage measurements back to the ESP32 over I2C
4. This branch mod adds the necessary driver for the ADS115, as well as the necessary HTTP endpoint (/api/voltage) to expose expose the voltage readouts

## Future Direction ##

1. Update the UI with a section to graphically show the ASICs, their status and voltage measurements
2. To further implement diagnostic and self-healing, so that failed ASICs in a chain can be flagged, and optionally either automatically or manually bypassed. This is especially important in seriesed chains, where a single failed ASIC will bring down an entire chain
3. To implement ASIC-level voltage tuning
4. To add the necessary API control endpoints, so this can be done with an API call

### DIFF
--- a/components/stratum/stratum_api.c
+++ b/components/stratum/stratum_api.c
@@ -162,8 +162,8 @@ void STRATUM_V1_parse(StratumApiV1Message * message, const char * stratum_json)
     int64_t parsed_id = -1;
     if (id_json != NULL && cJSON_IsNumber(id_json)) {
         parsed_id = id_json->valueint;
-        last_parsed_request_id = parsed_id;
     }
+    last_parsed_request_id = parsed_id;
     message->message_id = parsed_id;
 
     cJSON * method_json = cJSON_GetObjectItem(json, "method");


johnny9 (changes requested):
Don't include binaries and extra diff files.

Check the build errors in the unittest and build github checks.sha-6a7c499a5dd8f985a578a05e04eba3fa9f93f1f7

ahmedalalousi:
Fixed, updated, got tree control to work and tested with build/flash - multiple changes, so closing PR. Will create a fresh PR for all of this.

PR #1131, opened 2025-07-05
Add mDNS hostname advertisement

korbin:
This PR adds Espressif's mDNS component, enabling consistent hostname-based access of Bitaxe devices in varying network topologies.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,9 @@ jobs:
       with:
         esp_idf_version: v5.4.2
         target: esp32s3
-        command: GITHUB_ACTIONS="true" idf.py build
+        command: |
+          GITHUB_ACTIONS="true" idf.py add-dependency "espressif/mdns^1.8.2"
+          GITHUB_ACTIONS="true" idf.py build
         path: '.'
     - uses: actions/setup-python@v5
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -22,7 +22,9 @@ jobs:
       with:
         esp_idf_version: v5.4.2
         target: esp32s3
-        command: GITHUB_ACTIONS="true" idf.py build
+        command: |
+          GITHUB_ACTIONS="true" idf.py add-dependency "espressif/mdns^1.8.2"
```
johnny9, line 26: Can this get added to our devcontainer as well?

johnny9: or does it not need it because its included in the idf dependancies for our main component?

johnny9: I think I see the problem, its just the unittest target that needs it.

johnny9:
Can we get is_network_allowed(httpd_req_t * req) to allow "hostname.local" and the hostname should be what is currently set for mdns during the init and not the nvs config since that can be overwritten? mdns_hostname_get can get the hostname

korbin:
I'm opening a new PR with a lot more.

PR #1146, opened 2025-07-11
splitt modules. 

KillerInk:
this makes it easier to use modules and remove the error prone passing the pointer around^^
pls look through carfully! but i think i removed all shit i made to my own repo that do not affect this.

### DIFF
--- a/components/asic/asic.c
+++ b/components/asic/asic.c
@@ -14,42 +14,42 @@ static const double NONCE_SPACE = 4294967296.0; //  2^32
 
 static const char *TAG = "asic";
 
-uint8_t ASIC_init(GlobalState * GLOBAL_STATE)
+uint8_t ASIC_init()
 {
-    ESP_LOGI(TAG, "Initializing %s", GLOBAL_STATE->DEVICE_CONFIG.family.asic.name);
+    ESP_LOGI(TAG, "Initializing %s", DEVICE_CONFIG.family.asic.name);
+    switch (DEVICE_CONFIG.family.asic.id) {
 
-    switch (GLOBAL_STATE->DEVICE_CONFIG.family.asic.id) {
         case BM1397:
-            return BM1397_init(GLOBAL_STATE->POWER_MANAGEMENT_MODULE.frequency_value, GLOBAL_STATE->DEVICE_CONFIG.family.asic_count, GLOBAL_STATE->DEVICE_CONFIG.family.asic.difficulty);
+            return BM1397_init(POWER_MANAGEMENT_MODULE.frequency_value, DEVICE_CONFIG.family.asic_count, DEVICE_CONFIG.family.asic.difficulty);
         case BM1366:
-            return BM1366_init(GLOBAL_STATE->POWER_MANAGEMENT_MODULE.frequency_value, GLOBAL_STATE->DEVICE_CONFIG.family.asic_count, GLOBAL_STATE->DEVICE_CONFIG.family.asic.difficulty);
+            return BM1366_init(POWER_MANAGEMENT_MODULE.frequency_value, DEVICE_CONFIG.family.asic_count, DEVICE_CONFIG.family.asic.difficulty);
         case BM1368:
-            return BM1368_init(GLOBAL_STATE->POWER_MANAGEMENT_MODULE.frequency_value, GLOBAL_STATE->DEVICE_CONFIG.family.asic_count, GLOBAL_STATE->DEVICE_CONFIG.family.asic.difficulty);
+            return BM1368_init(POWER_MANAGEMENT_MODULE.frequency_value, DEVICE_CONFIG.family.asic_count, DEVICE_CONFIG.family.asic.difficulty);
         case BM1370:
-            return BM1370_init(GLOBAL_STATE->POWER_MANAGEMENT_MODULE.frequency_value, GLOBAL_STATE->DEVICE_CONFIG.family.asic_count, GLOBAL_STATE->DEVICE_CONFIG.family.asic.difficulty);
+            return BM1370_init(POWER_MANAGEMENT_MODULE.frequency_value, DEVICE_CONFIG.family.asic_count, DEVICE_CONFIG.family.asic.difficulty);
         default:
     }
     return ESP_OK;
 }
 
-task_result * ASIC_process_work(GlobalState * GLOBAL_STATE)
+task_result * ASIC_process_work()
 {
-    switch (GLOBAL_STATE->DEVICE_CONFIG.family.asic.id) {
+    switch (DEVICE_CONFIG.family.asic.id) {
         case BM1397:
-            return BM1397_process_work(GLOBAL_STATE);
+            return BM1397_process_work();
         case BM1366:
-            return BM1366_process_work(GLOBAL_STATE);
+            return BM1366_process_work();
         case BM1368:
-            return BM1368_process_work(GLOBAL_STATE);
+            return BM1368_process_work();
         case BM1370:
-            return BM1370_process_work(GLOBAL_STATE);
+            return BM1370_process_work();
     }
     return NULL;
 }
 
-int ASIC_set_max_baud(GlobalState * GLOBAL_STATE)
+int ASIC_set_max_baud()
 {
-    switch (GLOBAL_STATE->DEVICE_CONFIG.family.asic.id) {
+    switch (DEVICE_CONFIG.family.asic.id) {
         case BM1397:
             return BM1397_set_max_baud();
         case BM1366:
@@ -62,27 +62,27 @@ int ASIC_set_max_baud(GlobalState * GLOBAL_STATE)
     return 0;
 }
 
-void ASIC_send_work(GlobalState * GLOBAL_STATE, void * next_job)
+void ASIC_send_work(void * next_job)
 {
-    switch (GLOBAL_STATE->DEVICE_CONFIG.family.asic.id) {
+    switch (DEVICE_CONFIG.family.asic.id) {
         case BM1397:
-            BM1397_send_work(GLOBAL_STATE, next_job);
+            BM1397_send_work(next_job);
             break;
         case BM1366:
-            BM1366_send_work(GLOBAL_STATE, next_job);
+            BM1366_send_work(next_job);
             break;
         case BM1368:
-            BM1368_send_work(GLOBAL_STATE, next_job);
+            BM1368_send_work(next_job);
             break;
         case BM1370:
-            BM1370_send_work(GLOBAL_STATE, next_job);
+            BM1370_send_work(next_job);
             break;
     }
 }
 
-void ASIC_set_version_mask(GlobalState * GLOBAL_STATE, uint32_t mask)
+void ASIC_set_version_mask(uint32_t mask)
 {
-    switch (GLOBAL_STATE->DEVICE_CONFIG.family.asic.id) {
+    switch (DEVICE_CONFIG.family.asic.id) {
         case BM1397:
             BM1397_set_version_mask(mask);
             break;
@@ -98,12 +98,12 @@ void ASIC_set_version_mask(GlobalState * GLOBAL_STATE, uint32_t mask)
     }
 }
 
-bool ASIC_set_frequency(GlobalState * GLOBAL_STATE, float target_frequency)
+bool ASIC_set_frequency(float target_frequency)
 {
-    ESP_LOGI(TAG, "Setting ASIC frequency to %.2f MHz", target_frequency);
+    //ESP_LOGI(TAG, "Setting ASIC frequency to %.2f MHz", target_frequency);
     bool success = false;
     
-    switch (GLOBAL_STATE->DEVICE_CONFIG.family.asic.id) {
+    switch (DEVICE_CONFIG.family.asic.id) {
         case BM1366:
             success = BM1366_set_frequency(target_frequency);
             break;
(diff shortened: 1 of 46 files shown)


components/asic/bm1366.c
```diff
@@ -195,13 +195,13 @@ void BM1366_send_hash_frequency(float target_freq)
     ESP_LOGI(TAG, "Setting Frequency to %.2fMHz (%.2f)", target_freq, newf);
 }
 
-static void do_frequency_ramp_up(float target_frequency) {
+static void do_frequency_ramp_up(double target_frequency) {
```
0xf0xx0: change missed

components/asic/bm1366.c
```diff
@@ -195,13 +195,13 @@ void BM1366_send_hash_frequency(float target_freq)
...
     ESP_LOGI(TAG, "Ramping up frequency from %.2f MHz to %.2f MHz", current_frequency, target_frequency);
     do_frequency_transition(target_frequency, BM1366_send_hash_frequency, 1366);
 }
 
 // Add a public function for external use
-bool BM1366_set_frequency(float target_freq) {
+bool BM1366_set_frequency(double target_freq) {
```
0xf0xx0: here too

johnny9 (commented):
NACK

This doesn't address the issue with the dependancies between these components and main and in fact just makes a lot of this harder to eventually test.

What needs  to be done is asic.h, for instance, needs to no longer know about global state.

On top of that this is far too large of a change to go in without any additional tests being written.

KillerInk:
> NACK
> 
> This doesn't address the issue with the dependancies between these components and main and in fact just makes a lot of this harder to eventually test.
> 
> What needs to be done is asic.h, for instance, needs to no longer know about global state.
> 
> On top of that this is far too large of a change to go in without any additional tests being written.

the thing is that the GLOBAL_STATE is static in main but not accessible without routing it through methods, in fact its static^^
other thing is that everything that is stored in that pointer struct can lead to mem leaks when a new value get set and the old one is not freed.
https://github.com/bitaxeorg/ESP-Miner/blob/d47275b94231a272fbab63395e075725f07857a6/main/tasks/stratum_task.c#L343-L346
like this
this makes sure that every struct with extern flag can exists only once.
i can also move the structs into own headers, then they can get included as needed

johnny9:
> the thing is that the GLOBAL_STATE is static in main but not accessible without routing it through methods, in fact its static^^ other thing is that everything that is stored in that pointer struct can lead to mem leaks when a new value get set and the old one is not freed.
> 

Ok, this might be worthwhile and I might agree, but you called the PR splitt modules but there wasn't any change in dependencies. You need to work on the commit and PR clarity here. It changes 1000 lines of code. Lets be clearer on the intent

components/asic/asic.c
```diff
@@ -1,55 +1,53 @@
...
 #include <esp_log.h>
-
 #include "bm1397.h"
 #include "bm1366.h"
 #include "bm1368.h"
 #include "bm1370.h"
-
 #include "asic.h"
 #include "device_config.h"
+#include "power_management_task.h"
```
johnny9: damn this sucks. I know you didn't cause the problem and this change is just exposing it more but asic shouldnt care at all about the tasks that are leveraging it.

johnny9:
> other thing is that everything that is stored in that pointer struct can lead to mem leaks when a new value get set and the old one is not freed.
> 
> https://github.com/bitaxeorg/ESP-Miner/blob/d47275b94231a272fbab63395e075725f07857a6/main/tasks/stratum_task.c#L343-L346
> 
> like this

Not everything in the struct will cause memory leaks. Only the alloc'd values like the strings. That is unavoidable. we need to free the dynamically allocated strings regardless of if GLOBAL_STATE holds the pointer or some other module.

KillerInk:
> > other thing is that everything that is stored in that pointer struct can lead to mem leaks when a new value get set and the old one is not freed.
> > https://github.com/bitaxeorg/ESP-Miner/blob/d47275b94231a272fbab63395e075725f07857a6/main/tasks/stratum_task.c#L343-L346
> > 
> > like this
> 
> Not everything in the struct will cause memory leaks. Only the alloc'd values like the strings. That is unavoidable. we need to free the dynamically allocated strings regardless of if GLOBAL_STATE holds the pointer or some other module.

i did not say it will, it can. but its now more safe because its using the same memory location and make the free obsolete.
on my side i never noted that leak^^

> > the thing is that the GLOBAL_STATE is static in main but not accessible without routing it through methods, in fact its static^^ other thing is that everything that is stored in that pointer struct can lead to mem leaks when a new value get set and the old one is not freed.
> 
> Ok, this might be worthwhile and I might agree, but you called the PR splitt modules but there wasn't any change in dependencies. You need to work on the commit and PR clarity here. It changes 1000 lines of code. Lets be clearer on the intent

I probably have to, if you have a good title feel free to change it

components/asic/asic.c
KillerInk: there will more come up. its not the first time im doing such things^^

johnny9:
Concept ACK.

This is starting to look much better. I'll make an effort to test this out this week and do a more thorough review.

KillerInk:
Take your time. i dont expect this get merged soon^^
i realy tried to move the modules into a "modules" folder. but no idea why esp idf dont like header only folders.

components/asic/asic.c
```diff
@@ -1,125 +1,87 @@
...
-        case BM1366:
-            return BM1366_init(GLOBAL_STATE->POWER_MANAGEMENT_MODULE.frequency_value, GLOBAL_STATE->DEVICE_CONFIG.family.asic_count, GLOBAL_STATE->DEVICE_CONFIG.family.asic.difficulty);
-        case BM1368:
-            return BM1368_init(GLOBAL_STATE->POWER_MANAGEMENT_MODULE.frequency_value, GLOBAL_STATE->DEVICE_CONFIG.family.asic_count, GLOBAL_STATE->DEVICE_CONFIG.family.asic.difficulty);
-        case BM1370:
-            return BM1370_init(GLOBAL_STATE->POWER_MANAGEMENT_MODULE.frequency_value, GLOBAL_STATE->DEVICE_CONFIG.family.asic_count, GLOBAL_STATE->DEVICE_CONFIG.family.asic.difficulty);
+void ASIC_init_methods(int id)
+{
+    switch (id) {
+        case 0:
```
mutatrum: Don't switch on enum ids.

Also not sure why we're returning to function pointer, we were moving away from that.

Maybe it's better to have 2 PRs, one focussing on the GLOBAL_STATE, and only that, and another on revisiting the function delegation pattern.

components/asic/bm1370.c
```diff
@@ -195,18 +197,20 @@ void BM1370_send_hash_frequency(float target_freq) {
     ESP_LOGI(TAG, "Setting Frequency to %.2fMHz (%.2f)", target_freq, newf);
 }
 
-static void do_frequency_ramp_up(float target_frequency) {
+static void do_frequency_ramp_up(float target_frequency)
```
mutatrum: For a big PR like this, please don't add in whitespace and formatting changes. It's big enough already.

main/asic_task_module.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
```
mutatrum: Headers in rest of code are guarded by ifndef blocks, maybe keep consistent to rest of code?

components/asic/asic.c
KillerInk: in that case function pointer make sense.
the pointers are sealed from the rest of the source and get only set once at the beginning and avoid unnecessary repeating lookups.

also the enum lookup with ids is fine in that case, i changed the enum^^
better would be to move the enum in its own header and include it then here.

components/asic/bm1370.c
KillerInk: to be fair i only pressed shift+alt+f, sorry

KillerInk:
> Initial code review, not finished yet.
> 
> I'm not 100% sure the division between modules is correct, especially asic_task, mining, state and system are very close together. This might warrant another look.
> 
> It would help to eliminate all whitespace/formatting changes, and the bm#### function delegation changes. Can you try to make this PR as small as possible, with only the GLOBAL_STATE stuff?

i can... you guys should press more often shift+alt+f

aisc changes come from this conversation

https://github.com/bitaxeorg/ESP-Miner/pull/1146#discussion_r2202421092

components/asic/bm1370.c
mutatrum: I understand, problem is there's no clear defined code-style - at least, that I know - and contributors use all kinds of editors, so this makes it difficult.

KillerInk (commented):
i rechecked everything requested

main/http_server/http_server.c
```diff
@@ -27,19 +27,25 @@
...
-#include "http_server.h"
+#include "theme_api.h" // Add theme API include
+#include "vcore.h"
 #include "system.h"
+#include "power_management_module.h"
+#include "system_module.h"
+#include "device_config.h"
+#include "wifi_module.h"
+#include "pool_module.h"
+#include "state_module.h"
```
mutatrum, line 50: All these includes is a sign that we need to overhaul this class as well. But that's outside the scope of this PR.

main/self_test/self_test.h
```diff
@@ -1,6 +1,7 @@
 #ifndef SELF_TEST_H_
 #define SELF_TEST_H_
 
-bool self_test(void * pvParameters);
+bool self_test();
+bool should_test();
```
mutatrum: This doesn't need to be in the header file, it's not called from outside.

main/tasks/asic_task.c
```diff
@@ -6,44 +6,62 @@
...
+#include "mining_module.h"
+#include "device_config.h"
+#include "asic_task_module.h"
 
 static const char *TAG = "asic_task";
 
 // static bm_job ** active_jobs; is required to keep track of the active jobs since the
+const double NONCE_SPACE = 4294967296.0; //  2^32
+
+double ASIC_get_asic_job_frequency_ms(float frequency)
```
mutatrum, line 21: Interesting to see this moved here. Kinda makes sense, at least in the current version.

main/main.c
```diff
@@ -1,26 +1,45 @@
...
+PowerManagementModule POWER_MANAGEMENT_MODULE;
+DeviceConfig DEVICE_CONFIG;
+DisplayConfig DISPLAY_CONFIG;
+AsicTaskModule ASIC_TASK_MODULE;
+SelfTestModule SELF_TEST_MODULE;
+StatisticsModule STATISTICS_MODULE;
+mining_queues MINING_MODULE;
+WifiSettings WIFI_MODULE;
+PoolModule POOL_MODULE;
+StateModule STATE_MODULE;
```
mutatrum, line 43: Why are these all defined here?

main/system.c
```diff
@@ -37,94 +43,91 @@ static void _suffix_string(uint64_t, char *, size_t, int);
...
 
-esp_err_t SYSTEM_init_peripherals(GlobalState * GLOBAL_STATE) {
+esp_err_t SYSTEM_init_peripherals() {
     
     ESP_RETURN_ON_ERROR(gpio_install_isr_service(0), TAG, "Error installing ISR service");
 
     // Initialize the core voltage regulator
-    ESP_RETURN_ON_ERROR(VCORE_init(GLOBAL_STATE), TAG, "VCORE init failed!");
-    ESP_RETURN_ON_ERROR(VCORE_set_voltage(GLOBAL_STATE, nvs_config_get_u16(NVS_CONFIG_ASIC_VOLTAGE, CONFIG_ASIC_VOLTAGE) / 1000.0), TAG, "VCORE set voltage failed!");
+    ESP_RETURN_ON_ERROR(VCORE_init(NULL), TAG, "VCORE init failed!");
```
mutatrum: Why pass `NULL`?

main/system.c
```diff
@@ -29,6 +29,12 @@
 #include "screen.h"
 #include "vcore.h"
 #include "thermal.h"
+#include "asic_task_module.h"
+#include "system_module.h"
+#include "device_config.h"
+#include "pool_module.h"
+#include "state_module.h"
```
mutatrum, line 37: Similar commend to http_server, this file might need to be broken up as well. But not for this PR.

main/system_module.h
```diff
@@ -0,0 +1,77 @@
...
+{
+    char message[64];
+    uint32_t count;
+} RejectedReasonStat;
+
+
+
+typedef struct
+{
+    // The starting time for a certain period of mining activity.
```
mutatrum: This comment is not correct I think. AFAIK it has something to do with the statistics. Maybe those can be moved into a Statistics_module?

main/main.c
KillerInk: its only for exploring the code.
normaly when you start reading new code you start with the main class and that makes visible that there are different structs in use. and if you look into the header you see it as extern and know its global usable.
it can replaced with static inside the module header and removed from main if needed

main/system.c
KillerInk: good question^^

main/system_module.h
KillerInk: its correct^^
https://github.com/bitaxeorg/ESP-Miner/blob/27c8dc44565fe4830d1177915c28ea26f2105508/main/system.c#L213

KillerInk: its only used in system.c it dont have to be global

mutatrum:
As an exercise to the reader (me) I tried to _only_ make `GLOBAL_STATE` static, and having done that, I'm not sure if this is the right direction to go. It makes testing harder, as you can't easily inject mock or test data anymore. To be able to do this, we would need to revert back to accepting parameters again. Furthermore, it makes lazy programming easier, by not having to think about responsibilities of a function, as everything is available everywhere.

My try is in https://github.com/mutatrum/ESP-Miner/tree/static-global-state. Some stuff could be done further, f.e. the `BM####_init` functions don't need these parameters anymore, as the `bm####.c` files can access them directly. And that's exactly where it gets tricky IMO.

Having said that, I think the changes on how the modules are separated do make sense. Maybe a better way would be to indeed remove GlobalState, have the modules live in `main.c` and only pass around the specific modules that functions need. This forces a better encapsulation of responsibilities.

KillerInk:
this is only the top of the iceberg^^
the main problem are the cross references. class a need b and b need a.
dont know but maybe things should get a little bit more modular.
as sample the stratum and asic task. 
stratum does one thing with the asic task, its sends a new notification to asic. and asic send the result to stratum
that could get done different with function callbacks. then both classes dont rely on each other and you can easly replace the asic task with a different, if needed, without touching the stratum code or vise versa.
i think most the of the module structs can get removed/replaced by a local usage with a more modular approach.
that could also help to simplyfy and improve testing and reduce code complexity and force a better encapsulation of responsibilities^^

https://github.com/KillerInk/ESP-Miner/blob/385739054fc73e7516e0fc36e39533da44d010de/main/main.c#L91-L92

https://github.com/KillerInk/ESP-Miner/blob/8f43e63d151c35542534d4719d2a6153c4e7ddc5/main/tasks/asic_task.h#L12-L13
https://github.com/KillerInk/ESP-Miner/blob/8f43e63d151c35542534d4719d2a6153c4e7ddc5/main/tasks/asic_task.c#L29-L30
https://github.com/KillerInk/ESP-Miner/blob/8f43e63d151c35542534d4719d2a6153c4e7ddc5/main/tasks/asic_task_common.c#L27-L31

https://github.com/KillerInk/ESP-Miner/blob/8f43e63d151c35542534d4719d2a6153c4e7ddc5/main/tasks/stratum_task.h#L8
https://github.com/KillerInk/ESP-Miner/blob/8f43e63d151c35542534d4719d2a6153c4e7ddc5/main/tasks/stratum_task.c#L33
https://github.com/KillerInk/ESP-Miner/blob/8f43e63d151c35542534d4719d2a6153c4e7ddc5/main/tasks/stratum_task.c#L336

PR #1148, opened 2025-07-13
change icon pi-refresh to pi-power-off to disambiguate

eyelight:
The 'pi-refresh' icon used is so similar to the iOS browser's refresh icon that (without tooltip) it took a while to realize this menu selection performed a device reboot.

This change to 'pi-power-off' icon removes the ambiguity as it is clearly power-related.

Until the icon pack adds a power icon that includes arrow tips to indicate 'reset' / 'power cycle', maybe this is second-best?


duckaxe:
It's not an easy topic. I also racked my brains when choosing the icon. PrimeNG provides a limited number of icons (bad), which improves app speed (good).

I'm not sure if an icon that clearly indicates a function, like Power-Off (the icon matches 100 % for power off), is really better than a less clear icon, like the current one.

When searching for "restart" on flaticon, I'd say our current icon isn't too bad.

Screenshot 2025-07-14 at 08 13 24

mutatrum:
Last one on 3rd row is best IMO. Try searching for `reboot`.

duckaxe:
Not all of our icons are perfect. We simply picked the best ones that PrimeNG has to offer. The same is true for the Restart icon.

> Try searching for `reboot`.

Mostly the same "reload" icons. 

Screenshot 2025-07-14 at 19 58 52

eyelight:
Yes, the third row mentioned above is superior with the vertical in the middle... 

So, it seems to be a speed/bloat tradeoff of adding the second icon pack?

Yes, and my "power off" icon does not eliminate all ambiguity.

One way to meet in the middle (*not* add the second set of icons) might be this -- it seems to be true that there is no "off" for a bitaxe unless you unplug it physically, so, what about adding a confirmation modal to that button which teaches/clarifies the intent with something like, "reboot bitaxe? ok/cancel". In addition to training users on the function of the ambiguous icon, the confirm modal would also prevent accidental reboots for those who intended, oops, to refresh the page content.

duckaxe:
I think it's excessive to put so much effort into one small function. Yes, the current icon is imperfect. "Power-Off" is not ideal either. However, I believe users will understand the icon's meaning after using it for the first time. Besides, we've seen that the icon isn't bad.

If we want to do it right, we should eliminate PrimeNG icons and introduce custom SVG icons. However, I think we have other priorities waiting to be implemented.

Since we don't have any better alternatives for the current icon, I wouldn't change it. My opinion. The decision is up to the community.

WantClue:
I agree on this with duckaxe and NACK this concept of chaning the restart button again. Ofc this adds an initial confusion to the user especially with the design concept that is similar to some apple styles but people will get used to this really quickly and adding another pack just adds to it instead of making the design maintanable

mutatrum:
NACK, I think adding another icon pack is indeed overkill. For now, I guess it's good enough:tm:. Maybe we can move to all SVG icons at some point, and we can do what we want - and get rid of `pi` in total, but not sure if it's worth the effort.

PR #1134, opened 2025-07-06
Add tooltips for suggest difficulty and extranonce subscribe pool fields

mutatrum:

PR #1149, opened 2025-07-15
Show/Hide pool advanced options

duckAxe:
Since v2.9.0, I've often seen people asking for the new pool options, `Suggested Difficulty` and `Extranonce Subscribe`. To avoid confusing beginners and reduce the number of support requests, this pull request puts both options in a show/hide wrapper and marks them as "advanced" fields.

https://github.com/user-attachments/assets/9a8ba46b-6c5e-4fbb-ba57-0e1d5d1c60fe

### DIFF
--- a/main/http_server/axe-os/src/app/components/pool/pool.component.html
+++ b/main/http_server/axe-os/src/app/components/pool/pool.component.html
@@ -34,24 +34,35 @@
                             [type]="showPassword[pool] ? 'text' : 'password'" />
                     </div>
                 </div>
-                <div class="field grid p-fluid">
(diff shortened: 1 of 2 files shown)


WantClue (approved):
ACK, added a tooltip for the suggestDiff as well

johnny9 (approved):
tACK

tested on a 201. Looks great.

duckaxe:
@WantClue I would't add tooltip in this PR because @mutatrum created already a PR for that https://github.com/bitaxeorg/ESP-Miner/pull/1134

duckaxe:
After talking to @WantClue , I removed the tooltip so that we can add it with https://github.com/bitaxeorg/ESP-Miner/pull/1134 .

Issue #1099, opened 2025-06-30
Enhancement request: overheat control to lower settings till temperature stabilizes


plmarin:
It's absolutely annoying to be checking the device status time to time, restore parameters and restart  every time overheat is enabled. Please allow us OC in our devices. 

Thanks.

mutatrum:
#987 should help a bit. And I think it could be done without a restart as well, with #734. But overheat protection is there for a reason, if that's disabled, the result is a fried chip. And we don't want fire in your house 😉 

Currently it's 75 °C for the ASIC and 105 °C for the voltage regulator (TPS546).

plmarin:
OK, good point — I definitely don’t want my house on fire! But… I have a couple of suggestions that might be worth considering:

If overheat mode is enabled, is there any chance the settings could automatically revert to their previous configuration once temperatures return to normal?

Could you add more fine-tuned control? For example, manual input boxes for Frequency and Core Voltage, and also for the overheat threshold. The idea would be to tweak performance within a safe range, increasing or decreasing slightly while being careful with those changes. Would that be possible?

Many thanks!

KillerInk:
i made autoclock for my miners. you can set the max fanspeed or powerlimit. and with rising temps it starts to throttle and when it gets cooler it ramp up. runs here stable now for a week and room temps are between 16-34°C^^ i hate summer....
https://github.com/KillerInk/ESP-Miner/tree/autoclock

skot:
> If overheat mode is enabled, is there any chance the settings could automatically revert to their previous configuration once temperatures return to normal?

This is tough. What if the Bitaxe is overheating at defaults? (This can happen very quickly if the fan stops or the heatsink isn't right)

plmarin:
> This is tough. What if the Bitaxe is overheating at defaults? (This can happen very quickly if the fan stops or the heatsink isn't right)

This could be addressed by enabling a dynamic threshold setting for the thermal protection trigger.

skot:
I'm not sure I follow.. are you proposing making the overheat temp threshold adjustable? Why would you want it lower than the max?

KillerInk:
@plmarin you need auto tune. overheat is fine as it is. its the last protection bevor you burn you house.

![Image](https://github.com/user-attachments/assets/d657ac0f-c837-4165-8f2b-d331ddb82a14)

plmarin:
[...]

PR #1152, opened 2025-07-17
Auto-Tuning Capabilities

KillerInk:
## What This Does
This PR introduces a new set of features that significantly enhance the auto-tuning capabilities of the ESP-Miner. The auto-tune functionality is crucial for optimizing mining performance by adjusting key parameters such as power limit, fan speed, voltage, and frequency based on real-time conditions.

### Auto-Tune Enhancements:
- **Dynamic Power Management**: Users can now adjust the power limit to ensure optimal energy usage while maintaining high hashrates.
- **Temperature Control**: Fine-tuning fan limits allows for better thermal management, preventing overheating and potential hardware damage.
- **Voltage and Frequency Adjustments**: The ability to modify step voltage and frequency settings helps in achieving the best possible performance from ASICs under varying conditions.

### Benefits:
- **Optimized Performance**: By dynamically adjusting parameters, users can maximize mining efficiency and profitability.
- **Improved Hardware Longevity**: Better temperature control and power management contribute to the longevity of the mining hardware.
- **Remote Configuration**: The API-driven approach allows for remote adjustments, making it easier to manage multiple miners or integrate with external monitoring systems.

By enhancing the auto-tuning features, this PR aims to provide users with more control over their mining operations, leading to improved performance and potentially higher returns.

### DIFF
--- a/main/http_server/axe-os/src/app/components/autotune/autotune.component.html
+++ b/main/http_server/axe-os/src/app/components/autotune/autotune.component.html
@@ -1,0 +1,35 @@
+<ng-container>
+    <div class="card">
+        <form [formGroup]="autotuneForm" (ngSubmit)="updateAutotune()" *ngIf="autotuneForm">
+            <h3>Autotune Settings</h3>
+            <div class="form-check mb-3">
+                <label for="auto_tune_hashrate">Auto Tune Hashrate</label>
+                <input type="checkbox" id="auto_tune_hashrate" formControlName="auto_tune_hashrate" />
+            </div>
+            <div class="row">
+                <div class="col">
+                    <label>Power Limit</label>
+                    <input type="number" class="form-control" formControlName="power_limit">
+                </div>
+                <div class="col">
+                    <label>Fan Limit</label>
+                    <input type="number" class="form-control" formControlName="fan_limit">
+                </div>
+                <div class="col">
+                    <label>Max Voltage ASIC</label>
+                    <input type="number" class="form-control" formControlName="max_voltage_asic">
+                </div>
+                <div class="col">
+                    <label>Max Frequency ASIC</label>
+                    <input type="number" class="form-control" formControlName="max_frequency_asic">
+                </div>
+                <div class="col">
+                    <label>Max ASIC Temperatur</label>
+                    <input type="number" class="form-control" formControlName="max_asic_temperatur">
+                </div>
+            </div>
+            <button type="submit" class="btn btn-primary" [disabled]="!autotuneForm.valid">Save Autotune
+                Settings</button>
+        </form>
+    </div>
+</ng-container>
--- a/main/http_server/axe-os/src/app/components/autotune/autotune.component.ts
+++ b/main/http_server/axe-os/src/app/components/autotune/autotune.component.ts
@@ -1,0 +1,60 @@
+import { Component, OnInit, ViewChild } from '@angular/core';
+import { FormBuilder, FormGroup, Validators } from '@angular/forms';
+import { HttpErrorResponse, HttpEventType } from '@angular/common/http';
+import { SystemService } from 'src/app/services/system.service';
+import { ToastrService } from 'ngx-toastr';
+
+@Component({
+  selector: 'autotune',
+  templateUrl: './autotune.component.html',
+})
+export class AutotuneComponent implements OnInit {
+  constructor(private fb: FormBuilder,
+    private systemService: SystemService,
+    private toastr: ToastrService,) { }
+  public autotuneForm!: FormGroup;
+  public autotuneInfo: any = {};
+
+  //private toastr: ToastrService;
+  //private toastrService: ToastrService;
+  //private loadingService: LoadingService;
+
+
+  ngOnInit() {
+    this.autotuneForm = this.fb.group({
+      auto_tune_hashrate: [true, [Validators.required, Validators.required]],
+      power_limit: [20, [Validators.required, Validators.min(1)]],
+      fan_limit: [75, [Validators.required, Validators.min(1)]],
+      max_voltage_asic: [1400, [Validators.required, Validators.min(1)]],
+      max_frequency_asic: [1000, [Validators.required, Validators.min(1)]],
+      max_asic_temperatur: [65, [Validators.required, Validators.min(1)]],
+    });
+
+    // Load autotune settings from API and patch the form if available
+    this.systemService.getAutotune().subscribe({
+      next: autotune => {
+        this.autotuneInfo = autotune;
+        this.autotuneForm.patchValue({
+          auto_tune_hashrate: autotune.auto_tune_hashrate,
+          power_limit: autotune.power_limit ?? 20,
+          fan_limit: autotune.fan_limit ?? 75,
+          max_voltage_asic: autotune.max_voltage_asic ?? 1400,
+          max_frequency_asic: autotune.max_frequency_asic ?? 1000,
+          max_asic_temperatur: autotune.max_asic_temperatur ?? 65,
+        });
+      },
+      error: err => { this.toastr.error('Failed to load autotune settings'); }
+    });
+  }
+
+
+
+  public updateAutotune() {
+    if (!this.autotuneForm.valid) return;
+    this.systemService.updateAutotune(this.autotuneForm.value)
+      .subscribe({
+        next: () => this.toastr.success('Autotune settings saved!', 'Success'),
+        error: (err: HttpErrorResponse) => this.toastr.error('Could not save autotune settings.', err.message)
+      });
+  }
+}
--- a/main/http_server/http_server.c
+++ b/main/http_server/http_server.c
@@ -36,6 +36,7 @@
 #include "asic.h"
 #include "TPS546.h"
 #include "statistics_task.h"
+#include "auto_tune.h"
 #include "theme_api.h"  // Add theme API include
 #include "axe-os/api/system/asic_settings.h"
 #include "http_server.h"
@@ -1123,6 +1124,117 @@ void websocket_log_handler()
     }
 }
 
+esp_err_t GET_autotune_info(httpd_req_t * req)
+{
+    if (is_network_allowed(req) != ESP_OK) {
+        return httpd_resp_send_err(req, HTTPD_401_UNAUTHORIZED, "Unauthorized");
+    }
+    httpd_resp_set_type(req, "application/json");
+    if (set_cors_headers(req) != ESP_OK) {
+        httpd_resp_send_500(req);
+        return ESP_OK;
+    }
+
+    cJSON *root = cJSON_CreateObject();
+    cJSON_AddNumberToObject(root, NVS_CONFIG_KEY_POWER_LIMIT, AUTO_TUNE.power_limit);
+    cJSON_AddNumberToObject(root, NVS_CONFIG_KEY_FAN_LIMIT, AUTO_TUNE.fan_limit);
+    cJSON_AddNumberToObject(root, NVS_CONFIG_KEY_STEP_VOLT, AUTO_TUNE.step_volt);
+    cJSON_AddNumberToObject(root, NVS_CONFIG_KEY_STEP_FREQ_RAMPUP, AUTO_TUNE.step_freq_rampup);
+    cJSON_AddNumberToObject(root, NVS_CONFIG_KEY_STEP_FREQ, AUTO_TUNE.step_freq);
+    cJSON_AddNumberToObject(root, NVS_CONFIG_KEY_AUTOTUNE_STEP_FREQ, AUTO_TUNE.autotune_step_frequency);
+    cJSON_AddNumberToObject(root, NVS_CONFIG_KEY_MAX_VOLTAGE_ASIC, AUTO_TUNE.max_voltage_asic);
+    cJSON_AddNumberToObject(root, NVS_CONFIG_KEY_MAX_FREQUENCY_ASIC, AUTO_TUNE.max_frequency_asic);
+    cJSON_AddNumberToObject(root, NVS_CONFIG_KEY_MAX_ASIC_TEMPERATUR, AUTO_TUNE.max_asic_temperatur);
+    cJSON_AddNumberToObject(root, NVS_CONFIG_KEY_AUTO_TUNE_ENABLE, AUTO_TUNE.auto_tune_hashrate);
+
+    const char *response = cJSON_Print(root);
+    httpd_resp_sendstr(req, response);
+    free((void *)response);
+    cJSON_Delete(root);
+    return ESP_OK;
+}
+
+esp_err_t POST_autotune_update(httpd_req_t * req)
+{
+    if (is_network_allowed(req) != ESP_OK) {
+        return httpd_resp_send_err(req, HTTPD_401_UNAUTHORIZED, "Unauthorized");
+    }
+    if (set_cors_headers(req) != ESP_OK) {
+        httpd_resp_send_500(req);
+        return ESP_OK;
+    }
+
+    int total_len = req->content_len;
+    char buf[512];
+    int received = 0, cur_len = 0;
+    if (total_len >= sizeof(buf)) {
+        httpd_resp_send_err(req, HTTPD_500_INTERNAL_SERVER_ERROR, "content too long");
+        return ESP_OK;
+    }
+    while (cur_len < total_len) {
+        received = httpd_req_recv(req, buf + cur_len, total_len - cur_len);
+        if (received <= 0) {
+            httpd_resp_send_err(req, HTTPD_500_INTERNAL_SERVER_ERROR, "Failed to receive data");
+            return ESP_OK;
+        }
+        cur_len += received;
+    }
+    buf[total_len] = '\0';
+
+    cJSON *root = cJSON_Parse(buf);
+    if (!root) {
+        httpd_resp_send_err(req, HTTPD_400_BAD_REQUEST, "Invalid JSON");
+        return ESP_OK;
+    }
+
+    cJSON *item;
+    if ((item = cJSON_GetObjectItem(root, NVS_CONFIG_KEY_POWER_LIMIT)) && cJSON_IsNumber(item)) {
+        AUTO_TUNE.power_limit = item->valuedouble;
+        nvs_config_set_u16(NVS_CONFIG_KEY_POWER_LIMIT, (uint16_t)AUTO_TUNE.power_limit);
+    }
+    if ((item = cJSON_GetObjectItem(root, NVS_CONFIG_KEY_FAN_LIMIT)) && cJSON_IsNumber(item)) {
+        AUTO_TUNE.fan_limit = item->valuedouble;
+        nvs_config_set_u16(NVS_CONFIG_KEY_FAN_LIMIT, (uint16_t)AUTO_TUNE.fan_limit);
+    }
+    if ((item = cJSON_GetObjectItem(root, NVS_CONFIG_KEY_STEP_VOLT)) && cJSON_IsNumber(item)) {
+        AUTO_TUNE.step_volt = item->valuedouble;
+        // Optionally store in NVS if you add a key
+    }
+    if ((item = cJSON_GetObjectItem(root, NVS_CONFIG_KEY_STEP_FREQ_RAMPUP)) && cJSON_IsNumber(item)) {
+        AUTO_TUNE.step_freq_rampup = item->valuedouble;
+        // Optionally store in NVS if you add a key
+    }
+    if ((item = cJSON_GetObjectItem(root, NVS_CONFIG_KEY_STEP_FREQ)) && cJSON_IsNumber(item)) {
+        AUTO_TUNE.step_freq = item->valuedouble;
+        // Optionally store in NVS if you add a key
+    }
+    if ((item = cJSON_GetObjectItem(root, NVS_CONFIG_KEY_AUTOTUNE_STEP_FREQ)) && cJSON_IsNumber(item)) {
+        AUTO_TUNE.autotune_step_frequency = item->valuedouble;
+        // Optionally store in NVS if you add a key
+    }
+    if ((item = cJSON_GetObjectItem(root, NVS_CONFIG_KEY_MAX_VOLTAGE_ASIC)) && cJSON_IsNumber(item)) {
+        AUTO_TUNE.max_voltage_asic = item->valuedouble;
+        nvs_config_set_u16(NVS_CONFIG_KEY_MAX_VOLTAGE_ASIC, (uint16_t)AUTO_TUNE.max_voltage_asic);
+    }
+    if ((item = cJSON_GetObjectItem(root, NVS_CONFIG_KEY_MAX_FREQUENCY_ASIC)) && cJSON_IsNumber(item)) {
+        AUTO_TUNE.max_frequency_asic = item->valuedouble;
+        nvs_config_set_u16(NVS_CONFIG_KEY_MAX_FREQUENCY_ASIC, (uint16_t)AUTO_TUNE.max_frequency_asic);
+    }
+    if ((item = cJSON_GetObjectItem(root, NVS_CONFIG_KEY_MAX_ASIC_TEMPERATUR)) && cJSON_IsNumber(item)) {
+        AUTO_TUNE.max_asic_temperatur = item->valuedouble;
+        nvs_config_set_u16(NVS_CONFIG_KEY_MAX_ASIC_TEMPERATUR, (uint16_t)AUTO_TUNE.max_asic_temperatur);
+    }
+
+    if ((item = cJSON_GetObjectItem(root, NVS_CONFIG_KEY_AUTO_TUNE_ENABLE)) && cJSON_IsNumber(item)) {
+        AUTO_TUNE.auto_tune_hashrate = item->valuedouble;
+        nvs_config_set_u16(NVS_CONFIG_KEY_AUTO_TUNE_ENABLE, (uint16_t)AUTO_TUNE.auto_tune_hashrate);
+    }
+
+    cJSON_Delete(root);
+
+    httpd_resp_sendstr(req, "{\"status\":\"ok\"}");
+    return ESP_OK;
+}
 esp_err_t start_rest_server(void * pvParameters)
 {
     GLOBAL_STATE = (GlobalState *) pvParameters;
@@ -1256,6 +1368,22 @@ esp_err_t start_rest_server(void * pvParameters)
         .user_ctx = NULL
     };
     httpd_register_uri_handler(server, &update_post_ota_www);
+
+    httpd_uri_t autotune_get_uri = {
+        .uri = "/api/system/autotune",
+        .method = HTTP_GET,
+        .handler = GET_autotune_info,
+        .user_ctx = rest_context
+    };
+    httpd_register_uri_handler(server, &autotune_get_uri);
+
+    httpd_uri_t autotune_post_uri = {
+        .uri = "/api/system/autotune",
+        .method = HTTP_POST,
+        .handler = POST_autotune_update,
+        .user_ctx = rest_context
+    };
+    httpd_register_uri_handler(server, &autotune_post_uri);
 
     httpd_uri_t ws = {
         .uri = "/api/ws", 
--- a/main/nvs_config.h
+++ b/main/nvs_config.h
@@ -39,6 +39,16 @@
 #define NVS_CONFIG_OVERCLOCK_ENABLED "oc_enabled"
 #define NVS_CONFIG_SWARM "swarmconfig"
 #define NVS_CONFIG_STATISTICS_FREQUENCY "statsFrequency"
+#define NVS_CONFIG_KEY_POWER_LIMIT "power_limit"
+#define NVS_CONFIG_KEY_FAN_LIMIT "fan_limit"
+#define NVS_CONFIG_KEY_STEP_VOLT "step_volt"
+#define NVS_CONFIG_KEY_STEP_FREQ_RAMPUP "step_freq_rampup"
+#define NVS_CONFIG_KEY_STEP_FREQ "step_freq"
+#define NVS_CONFIG_KEY_AUTOTUNE_STEP_FREQ "autotune_step_frequency"
+#define NVS_CONFIG_KEY_MAX_VOLTAGE_ASIC "max_voltage_asic"
+#define NVS_CONFIG_KEY_MAX_FREQUENCY_ASIC "max_frequency_asic"
+#define NVS_CONFIG_KEY_MAX_ASIC_TEMPERATUR "max_asic_temperatur"
+#define NVS_CONFIG_KEY_AUTO_TUNE_ENABLE "auto_tune_hashrate"
 
 // Theme configuration
 #define NVS_CONFIG_THEME_SCHEME "themescheme"
--- a/main/tasks/power_management_task.h
+++ b/main/tasks/power_management_task.h
@@ -1,6 +1,8 @@
 #ifndef POWER_MANAGEMENT_TASK_H_
 #define POWER_MANAGEMENT_TASK_H_
 
+#define POLL_RATE 500
+
 typedef struct
 {
     uint16_t fan_perc;
--- a/main/thermal/auto_tune.c
+++ b/main/thermal/auto_tune.c
@@ -1,0 +1,264 @@
+#include "auto_tune.h"
+#include "PID.h"
+#include "esp_log.h"
+#include "nvs_config.h"
+#include "power_management_task.h"
+#include <math.h>
+#include <float.h>
+#include "global_state.h"
+
+static const char * TAG = "auto_tune";
+
+auto_tune_settings AUTO_TUNE = {.power_limit = 20,
+                                .fan_limit = 75,
+                                .step_volt = 0.5,
+                                .step_freq_rampup = 0.5,
+                                .step_freq = 0.5,
+                                .autotune_step_frequency = 0,
+                                .max_voltage_asic = 1400,
+                                .max_frequency_asic = 1000,
+                                .max_asic_temperatur = 65,
+                                .frequency = 525,
+                                .voltage = 1150,
+                                .auto_tune_hashrate = true};
+
+double last_core_voltage_auto;
+double last_asic_frequency_auto;
+double last_hashrate_auto;
+double current_hashrate_auto;
+double avg_hashrate_auto;
+
+bool lastVoltageSet = false;
+const int waitTime = 30;
+int waitCounter = 0;
+double freq_step;
+double volt_step;
+GlobalState * GLOBAL_STATE;
+
+#define INIT_FREQ 525
+#define INIT_VOLTAGE 1150
+#define MIN_FREQ 401
+#define MIN_VOLTAGE 1001
+
+enum TuneState
+{
+    sleep_before_warmup,
+    warmup,
+    working
+};
+
+enum TuneState state;
+
+void auto_tune_init(GlobalState * _GLOBAL_STATE)
+{
+    GLOBAL_STATE = _GLOBAL_STATE;
+    AUTO_TUNE.frequency = nvs_config_get_u16(NVS_CONFIG_ASIC_FREQ, AUTO_TUNE.frequency);
+    AUTO_TUNE.voltage = nvs_config_get_u16(NVS_CONFIG_ASIC_VOLTAGE, AUTO_TUNE.voltage);
+    AUTO_TUNE.power_limit = nvs_config_get_u16(NVS_CONFIG_KEY_POWER_LIMIT, AUTO_TUNE.power_limit);
+    AUTO_TUNE.fan_limit = nvs_config_get_u16(NVS_CONFIG_KEY_FAN_LIMIT, AUTO_TUNE.fan_limit);
+    AUTO_TUNE.max_voltage_asic = nvs_config_get_u16(NVS_CONFIG_KEY_MAX_VOLTAGE_ASIC, AUTO_TUNE.max_voltage_asic);
+    AUTO_TUNE.max_frequency_asic = nvs_config_get_u16(NVS_CONFIG_KEY_MAX_FREQUENCY_ASIC, AUTO_TUNE.max_frequency_asic);
+    AUTO_TUNE.max_asic_temperatur = nvs_config_get_u16(NVS_CONFIG_KEY_MAX_ASIC_TEMPERATUR, AUTO_TUNE.max_asic_temperatur);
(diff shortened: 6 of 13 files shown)


KillerInk:
this is related to #1099

KillerInk:
have to extend the chart, that it shows the applied voltage, frequency and fanspeed to have somekind of feedback for the user what is is happening.

duckaxe:
Still draft?

KillerInk:
Yes hope I have tomorrow time to extend the chart and add the missing stuff on esp side to statistic and livedata.
Think that my current chart is a little overdone and contains to many datasets.
Or I keep it as it is and add it as extra menu item where the dashboard, pool, update items are. 
But then I still have to cleanup unused datasets. 
I'm open to every suggestion

KillerInk:
Unbenannt-1

main/http_server/axe-os/src/app/components/autotune/autotune.component.ts
```diff
@@ -0,0 +1,60 @@
...
+    });
+  }
+
+
+
+  public updateAutotune() {
+    if (!this.autotuneForm.valid) return;
+    this.systemService.updateAutotune(this.autotuneForm.value)
+      .subscribe({
+        next: () => this.toastr.success('Autotune settings saved!', 'Success'),
```
duckaxe: Pls remove the title from toastr, see https://github.com/bitaxeorg/ESP-Miner/pull/1075/commits/65ec957f0c3b961764de1c7934831a4c29d9999e

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -201,13 +248,43 @@ export class HomeComponent {
...
           },
           suggestedMax: 80
+        },
+        y3: {
+          type: 'linear',
+          display: true,
+          position: 'right',
+          ticks: {
+            color: frequencyColor,
+            callback: (value: number) => value + 'mhz'
```
duckaxe: ` MHz` (with a space before)

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -201,13 +248,43 @@ export class HomeComponent {
...
+          },
+          suggestedMax: 1200
+        },
+        y4: {
+          type: 'linear',
+          display: true,
+          position: 'right',
+          ticks: {
+            color: voltageColor,
+            callback: (value: number) => value + 'mv'
```
duckaxe: ` mV` (with a space before)

main/http_server/axe-os/src/app/components/autotune/autotune.component.html
```diff
@@ -0,0 +1,35 @@
...
+        <form [formGroup]="autotuneForm" (ngSubmit)="updateAutotune()" *ngIf="autotuneForm">
+            <h3>Autotune Settings</h3>
+            <div class="form-check mb-3">
+                <label for="auto_tune_hashrate">Auto Tune Hashrate</label>
+                <input type="checkbox" id="auto_tune_hashrate" formControlName="auto_tune_hashrate" />
+            </div>
+            <div class="row">
+                <div class="col">
+                    <label>Power Limit</label>
+                    <input type="number" class="form-control" formControlName="power_limit">
```
duckaxe: No min/max attributes?

duckaxe:
Maybe it's a good idea to wait for https://github.com/bitaxeorg/ESP-Miner/pull/955 before any changes on the home chart?

main/http_server/axe-os/src/app/components/autotune/autotune.component.html
KillerInk: for power its enough to have only a max. the lower limit is defined through the lowest frequency 400 MHz(with space :D) and voltage 1000 mV.

KillerInk:
Maybe it is, no hurry. will keep this up to date with master

main/http_server/axe-os/src/app/components/autotune/autotune.component.ts
KillerInk: Roger that!

KillerInk:
well the chart part is more or less obsolete with #955 
so i can drop it an keep only the settings part

chillingbee:
Hey! 
I like this feature a lot... Very nice work!
But it would be nice to have a sort of button to set all the values back to "default" in case you played around and messed something up... 
Further more it would be nice to have a more detailed explantion on the ratio values. It was a little bit confusing to me how to set them.

KillerInk:
Hi @chillingbee 
thx for the suggestion. i will improve the v/r ratio limits description and add the reset button the next days

KillerInk:
well im to lazy to add a reset button^^
but i added default values to tooltips

Evanito:
Running this on my Gamma 602 - very cool stuff. The standard 1.2TH/s is up to about 1.4TH/s now, so there is good reason to ship this soon!

A few points of feedback I am going to leave on this feature as a whole as a user, but **I am not saying these need to be addressed for this PR**.
1. Not currently sure how safe the default limits are. Comparing them to the preset options for the Gamma 602 raised an eyebrow, so I lowered the maximums down to the high limits in the existing settings.
2. Decimals aren't rounded in the UI, leading to: 
image
3. Log is full of voltage updates to the same value. Can probably optimize this to only change voltage when it differs from the existing value.

```
₿ (3000971) power_management: setting new vcore voltage to 1159.599976mV
₿ (3000981) vcore: Set ASIC voltage = 1.160V
₿ (3000990) TPS546: Vout changed to 1.16 V
₿ (3001514) power_management: Temp: 54.9 °C, SetPoint: 55.0 °C, Output: 76.1% (P:15.0 I:0.2 D_val:3.0 D_start_val:20.0)
₿ (3001516) power_management: setting new vcore voltage to 1159.599976mV
₿ (3001523) vcore: Set ASIC voltage = 1.160V
₿ (3001540) TPS546: Vout changed to 1.16 V
₿ (3002055) power_management: Temp: 55.0 °C, SetPoint: 55.0 °C, Output: 78.7% (P:15.0 I:0.2 D_val:3.0 D_start_val:20.0)
₿ (3002056) power_management: setting new vcore voltage to 1159.599976mV
₿ (3002064) vcore: Set ASIC voltage = 1.160V
₿ (3002074) TPS546: Vout changed to 1.16 V
₿ (3002581) power_management: Temp: 55.0 °C, SetPoint: 55.0 °C, Output: 78.0% (P:15.0 I:0.2 D_val:3.0 D_start_val:20.0)
₿ (3002582) power_management: setting new vcore voltage to 1159.599976mV
₿ (3002590) vcore: Set ASIC voltage = 1.160V
₿ (3002598) TPS546: Vout changed to 1.16 V
₿ (3003022) bm1370: Job ID: 30, Core: 48/6, Ver: 00C8C000
```

4. It is not clear how fan limits in the autotune settings work with the automatic fan control. It seems that the autotune's default limit of 75% will try to make it so the automatic fan control is able to maintain the automatic temperature at 75% fan speed? It might be easier to combine these in some way... maybe move the existing fan control to the auto-tune settings page and include a note with how they interact.
5. Defaults in descriptions of some settings do not match reality.

I don't want to ask too much of this, so it makes sense to me if this is merged as-is and cleaned up before full release.

KillerInk:
> 1. Not currently sure how safe the default limits are. Comparing them to the preset options for the Gamma 602 raised an eyebrow, so I lowered the maximums down to the high limits in the existing settings.

i tested this excessive^^ the most important is the powerlimit because it dependes on the powersupply.
a stock miner will never pass 25Watts and hit befor the temperature or fan limit.
it looks different with a changed heatsink, then the miner can consume more power and its easy to hit the power limit. 
in that case you can incerase the powerlimit that it match your powersupply 80%rule. 
with 30-40watts on the miner the frequency and voltage limits get interessting.

> 2.Decimals aren't rounded in the UI, leading to:

> Maybe it's a good idea to wait for #955 before any changes on the home chart?

this should be obsolete when its merged. 

> 3. Log is full of voltage updates to the same value. Can probably optimize this to only change voltage when it differs from the existing value. 

https://github.com/bitaxeorg/ESP-Miner/blob/8a4e8b38e8034b0e27316187ba6b775dd6d2e809/main/tasks/power_management_task.c#L202-L206

there is a check in the code. that only set it when it have change. i think that there is something off with the float precission in the log.

answer the 4. 5. later. work waiting for me...

KillerInk:
> 4.It is not clear how fan limits in the autotune settings work with the automatic fan control. It seems that the autotune's default limit of 75% will try to make it so the automatic fan control is able to maintain the automatic temperature at 75% fan speed? It might be easier to combine these in some way... maybe move the existing fan control to the auto-tune settings page and include a note with how they interact.

it have something todo with pid and not. the temperature limit should match the pid temperature.
it try to tune the miner that way pid can keep the 75%fanspeed while maintaining the temperature. when it reach that 75 % the miner stops tuning till the 75+overshoot is reached, in that case it reduce voltage and frequency till its below 75%. if its back below 75% it resume tuning.

you can also disable pid(note to me i havto remove the pidactive flag) and let the fan run manual at specific speed, lets say 75%^^
in that case miner would tune until it hit another limit like temperature or power.

> 5.Defaults in descriptions of some settings do not match reality.

reality is relative^^ you mean the default:values?
for me that looks fine but i lost user sight 2 or 3 months ago^^
well the fanspeed description could cointain a note that its based on pid fanspeed 🐰

0xf0xx0:
liking this, its working pretty well, but i have some nits:

- related to 2, decimals should be rounded to 3 or 4 places. i doubt the asic can get this precise, and my vreg is 3 decimals max.
image 
image
- the logspam is a lot, maybe this could run every 3-5 ticks? a bit of delay isnt going to hurt anything. E: itll also let the fan catch up with the new chip state, right now it fluctuates audibly as the tuning happens.

KillerInk:
@0xf0xx0 its not using that precssion, frequency get increased by 0.2 and voltage by 0.1.
in auto tune voltage is 1093.920000, when it reaches power_managment its 1093.9200**44** and that seems to change with every call. so after the 7th digit we are loosing precision with float type and that is due how its stored in exponent and mantissa.
the only solution i see is to change everything that have todo with voltage and frequency to double.

KillerInk:
so that trim voltage and frequency to 2 decimals in ui. and logs trigger now only when voltage or frequency realy changed except the pid log

KillerInk:
> E: itll also let the fan catch up with the new chip state, right now it fluctuates audibly as the tuning happens.

that is a problem from pid. its to agressive^^
with that old 1800ms delay for the power_manangment_task.c pid jumped between 50-100% in one cycle with autoclock.
on my own branch i added an avg for the pid output to smooth it.
alternative you can adjust the fan limit and overshoot. 70% +10%overshoot.
at somepoint it should stop tuning and then the fanspeed get relativ constant.
its also up to the fan used. a 10000rpm fan at 50% is alot louder then a 6000rpm at same %.

0xf0xx0:
10% overshoot helped a bit, and logspam is gone :D

main/http_server/axe-os/src/app/components/autotune/autotune.component.html
```diff
@@ -0,0 +1,66 @@
+<form [formGroup]="autotuneForm" *ngIf="autotuneForm">
+    <div class="card">
+      <h2>Autotune Settings</h2>
+
+      <p-message severity="warn" styleClass="w-full mb-4 border-round-sm" text="The following options should only be set by expert users"></p-message>
```
0xf0xx0: nit: this should probably be closer to the oc settings warning

main/http_server/http_server.c
```diff
@@ -1035,6 +1041,121 @@ esp_err_t http_404_error_handler(httpd_req_t * req, httpd_err_code_t err)
...
+        nvs_config_set_double(NVS_CONFIG_KEY_VF_RATIO_MIN, AUTO_TUNE.vf_ratio_min);
+    }
+     if ((item = cJSON_GetObjectItem(root, NVS_CONFIG_KEY_MAX_VR_TEMPERATUR)) && cJSON_IsNumber(item)) {
+        AUTO_TUNE.max_vr_temperatur = (uint16_t)item->valueint;
+        nvs_config_set_u16(NVS_CONFIG_KEY_MAX_VR_TEMPERATUR, (uint16_t)AUTO_TUNE.max_vr_temperatur);
+    }
+
+    cJSON_Delete(root);
+
+    httpd_resp_sendstr(req, "{\"status\":\"ok\"}");
```
0xf0xx0: nit: status code is enough

main/nvs_config.h
```diff
@@ -40,6 +41,17 @@
...
+#define NVS_CONFIG_KEY_FAN_LIMIT "fan_limit"
+#define NVS_CONFIG_KEY_MAX_VOLTAGE_ASIC "max_volt_asic"
+#define NVS_CONFIG_KEY_MAX_FREQUENCY_ASIC "max_freq_asic"
+#define NVS_CONFIG_KEY_MAX_ASIC_TEMPERATUR "max_temp_asic"
+#define NVS_CONFIG_KEY_AUTO_TUNE_ENABLE "auto_tune"
+#define NVS_CONFIG_KEY_OVERSHOT_POWER_LIMIT "osh_pow_limit"
+#define NVS_CONFIG_KEY_OVERSHOT_FAN_LIMIT "osh_fan_limit"
+#define NVS_CONFIG_KEY_VF_RATIO_MAX "vf_ratio_max"
+#define NVS_CONFIG_KEY_VF_RATIO_MIN "vf_ratio_min"
+#define NVS_CONFIG_KEY_MAX_VR_TEMPERATUR "max_temp_vr"
```
0xf0xx0: nit: `_TEMP`, also the keys should be `lowercase` or `camelCase` to match the others  
preferably camel, as you use these as the api keys too

main/tasks/statistics_task.c
```diff
@@ -143,6 +143,8 @@ void statistics_task(void * pvParameters)
                 statsData.fanRPM = power_management->fan_rpm;
                 statsData.wifiRSSI = wifiRSSI;
                 statsData.freeHeap = esp_get_free_heap_size();
+            statsData.frequency = power_management->frequency_value;
+            statsData.core_voltage = power_management->core_voltage;
```
0xf0xx0: indent

main/thermal/auto_tune.c
```diff
@@ -0,0 +1,305 @@
...
+void auto_tune_init(GlobalState * _GLOBAL_STATE)
+{
+    GLOBAL_STATE = _GLOBAL_STATE;
+    AUTO_TUNE.frequency = nvs_config_get_u16(NVS_CONFIG_ASIC_FREQUENCY, AUTO_TUNE.frequency);
+    AUTO_TUNE.voltage = nvs_config_get_u16(NVS_CONFIG_ASIC_VOLTAGE, AUTO_TUNE.voltage);
+    AUTO_TUNE.power_limit = nvs_config_get_u16(NVS_CONFIG_KEY_POWER_LIMIT, AUTO_TUNE.power_limit);
+    AUTO_TUNE.fan_limit = nvs_config_get_u16(NVS_CONFIG_KEY_FAN_LIMIT, AUTO_TUNE.fan_limit);
+    AUTO_TUNE.max_voltage_asic = nvs_config_get_u16(NVS_CONFIG_KEY_MAX_VOLTAGE_ASIC, AUTO_TUNE.max_voltage_asic);
+    AUTO_TUNE.max_frequency_asic = nvs_config_get_u16(NVS_CONFIG_KEY_MAX_FREQUENCY_ASIC, AUTO_TUNE.max_frequency_asic);
+    AUTO_TUNE.max_asic_temperatur = nvs_config_get_u16(NVS_CONFIG_KEY_MAX_ASIC_TEMPERATUR, AUTO_TUNE.max_asic_temperatur);
```
0xf0xx0: similar, `_temp`

0xf0xx0 (changes requested):

0xf0xx0:
encountered an interesting edge case where the autotuning dropped my axe to the minimum clock and voltage, i think the fix is to check if the fan speed is near its set limit when the chip is at the temp limit. you can see in the image where it dropped and where i finally noticed. my settings were targeting 55° chip temp and 70% fan speed (+10% overshoot), and the actual fan speed during the drop was around 26%.

image

KillerInk:
im back from vacation^^
hm yes that happen when temp limit get set below pid temp. 
then pid falls to its set min speed or far below its limit.
so code doing its jop^^

temp limit is seprate because you can disable pid and set fan speed manual to 75%.
with pid enabled it must be temp limit >= pid temp

0xf0xx0:
> im back from vacation^^

welcome back :3 

> hm yes that happen when temp limit get set below pid temp.
then pid falls to its set min speed or far below its limit.
so code doing its jop^^

> temp limit is seprate because you can disable pid and set fan speed manual to 75%.
with pid enabled it must be temp limit >= pid temp

that might be it, id say its `temp limit > pid temp` as i had both set to 55°. testing again, in the meantime a nice ui thing would be to set the minimum slider value to `the current pid temp target + 1` if pid is active.

0xf0xx0:
yep, that was it. +1 nit uwu

main/http_server/http_server.c
KillerInk: roger that

main/thermal/auto_tune.c
KillerInk: temp_asic, if we take it seriously^^

main/nvs_config.h
KillerInk: hmm 
`#define NVS_CONFIG_OVERCLOCK_ENABLED "oc_enabled" `

gives me the right to do so^^

main/http_server/axe-os/src/app/components/autotune/autotune.component.ts
```diff
@@ -158,4 +171,22 @@ export class AutotuneComponent implements OnInit {
       error: (err: HttpErrorResponse) => this.toastr.error(`Could not save autotune settings. ${err.message}`)
     });
   }
+
+  private updateSliderMinForPid(info: any): void {
```
duckaxe: Avoid `any`, use `ISystemInfo`

main/http_server/axe-os/src/app/components/autotune/autotune.component.html
duckaxe: I would keep this message design. Will create a separate PR to uniform all message designs based on this one.

main/http_server/http_server.c
```diff
@@ -1175,6 +1295,22 @@ esp_err_t start_rest_server(void * pvParameters)
     };
     httpd_register_uri_handler(server, &update_post_ota_www);
 
+    httpd_uri_t autotune_get_uri = {
+        .uri = "/api/system/autotune",
```
duckaxe, line 1491: Please add both endpoints to the readme API section.

main/http_server/axe-os/src/app/components/autotune/autotune.component.ts
KillerInk: roger that

main/http_server/http_server.c
KillerInk: roger that

main/http_server/axe-os/src/app/components/autotune/autotune.component.html
KillerInk: so its duckaxe approved, i see this as resolved

kakulukia:
I have deployed that version on my Bitaxe 601 over a week ago and it performs much better than the base firmware version!

Well done!

0xf0xx0:
forgot a nit, there should be a toast when saving the autotuning settings

duckaxe:
> forgot a nit, there should be a toast when saving the autotuning settings

https://github.com/KillerInk/ESP-Miner/blob/1e308f5505ef848345deb1704b63b463712cc9ce/main/http_server/axe-os/src/app/components/autotune/autotune.component.ts#L170 ?

0xf0xx0:
> > forgot a nit, there should be a toast when saving the autotuning settings
> 
> 
> 
> https://github.com/KillerInk/ESP-Miner/blob/1e308f5505ef848345deb1704b63b463712cc9ce/main/http_server/axe-os/src/app/components/autotune/autotune.component.ts#L170 ?

it must not be getting called, ill take a closer look in a bit

KillerInk:
hmm what did i mess up. it get not called^^

kakulukia:
> hmm what did i mess up. it get not called^^

Toast is visible in the version I am testing on my bitaxe. Can't check which commit tho as I'm not at home for a few days, but revert a few commits and compare to the latest broken version. That should help finding that bug.

0xf0xx0:
finally got around to debugging, the reason is cause the connection is never closed on the axe side x3

using `httpd_resp_send_chunk(req, NULL, 0);` like the settings PATCH endpoint fixes it

main/http_server/axe-os/src/app/components/autotune/autotune.component.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    this.systemService.getInfo()
+      .pipe(this.loadingService.lockUIUntilComplete())
+      .subscribe({
+        next: (info) => {
+          // Update the slider config with dynamic minimum value if PID is active
+          this.updateSliderMinForPid(info);
+        },
+        error: () => {
+          this.toastr.error('Failed to load getInfog settings');
```
0xf0xx0: typo

0xf0xx0:
another thing ive encountered, adjusting the auto fan temp target above the autotuning target requires the user to also adjust the autotuning target. i think these should be linked when auto fan control is enabled, maybe with the slider in the autotuning settings hiding itself?

kakulukia:
That might explain things, but for me it looks like there are also other connections I have not yet figured out by just trying.

grafik
grafik
grafik

Sometimes it does exactly what i told it to but today its kinda unclear why it landed at its current target state.

Maybe it makes sense to move like two additional items (fan limit and autotune on/off) to the system settings and keep the rest hidden under "autotune advanced" so that for the felt mayority of cases moving thwo sliders at the system settings page is enough and the two separate target temp (asic) sliders that kinda corelated can be merged into one?

I guess most users probably wonder same as me about the rest of the autotune settings and probably should not alter any of those. This way the "kinda save sliders" are extracted which in INHO will be the better user experience in the end.

KillerInk:
> That might explain things, but for me it looks like there are also other connections I have not yet figured out by just trying.
> 
> grafik
> grafik
> grafik
> 
> Sometimes it does exactly what i told it to but today its kinda unclear why it landed at its current target state.
> 
> Maybe it makes sense to move like two additional items (fan limit and autotune on/off) to the system settings and keep the rest hidden under "autotune advanced" so that for the felt mayority of cases moving thwo sliders at the system settings page is enough and the two separate target temp (asic) sliders that kinda corelated can be merged into one?
> 
> I guess most users probably wonder same as me about the rest of the autotune settings and probably should not alter any of those. This way the "kinda save sliders" are extracted which in INHO will be the better user experience in the end.

Check asic temp. It's set to 50c and pid set to 64c. That cause pid not to rise fanspeed and autotune start to throttle because it goes above the 50+0.2overshoot

Think I understood now the problems that appear, and it need to get fixed with saving values on esp side. 
If pid target temp get saved and autotune is lower it need to automatic adjusted. The thing is you can use autotune with manual fanspeed . More tomorrow
Thx for feedback

kakulukia:
@KillerInk are you done with tha next version?

KillerInk:
@kakulukia yes its now forcing autotune if min fanspeed or temp is lower then from pid to pid values +1

kakulukia:
Going to test it later today. Thx!

main/http_server/http_server.c
```diff
@@ -779,8 +793,6 @@ static esp_err_t GET_system_info(httpd_req_t * req)
     cJSON_AddNumberToObject(root, "temptarget", nvs_config_get_u16(NVS_CONFIG_TEMP_TARGET, 60));
     cJSON_AddNumberToObject(root, "fanrpm", GLOBAL_STATE->POWER_MANAGEMENT_MODULE.fan_rpm);
 
-    cJSON_AddNumberToObject(root, "statsFrequency", nvs_config_get_u16(NVS_CONFIG_STATISTICS_FREQUENCY, 0));
```
0xf0xx0: accidental removal

kakulukia:
Hmmm .. i have just installed the new version, but i have to say that autotune feels like a totally independened plugin rather than part of the software now as max ASIC temp can only be set to something above the target temperature.

I understand where thats coming from as i was following this pull request, but i guess any other user would be confused about that limitation. 

I vote for adjusting the target temperature when max ASIC temp is set to lower values or as i have written above to put the most interesting sliders on the settings page and treat the rest as really advanced stuff hidden at the autune page.

@WantClue what is your take on how to handle all lthe additional settings?

kakulukia:
Another finding: The enhanced graph is broken :(

Somebody fixed the graph to be recorded in the background and not just live when visiting the dashboard page, BUT @KillerInk that destroyed your enhanced version of it. :(

kakulukia:
Oh, i just realized there is a new dropdown for even more values. I guess once autotune actually just works it makes no special sense to follow frequency and asic voltage (i guess those are the corresponding values - though they are not matching the settings page?).

main/http_server/http_server.c
KillerInk: not accidental^^ was intentional. i missunderstood what its doing

KillerInk:
> Hmmm .. i have just installed the new version, but i have to say that autotune feels like a totally independened plugin rather than part of the software now as max ASIC temp can only be set to something above the target temperature.
> 
> I understand where thats coming from as i was following this pull request, but i guess any other user would be confused about that limitation.

well the thing is, when you set the autotune max temperature below the pid target temp, autotune throttle down to 1001mv and 400mhz because pid do not increase the fanspeed because it do not hit the target temp. 
i think this behavior is more confusing and you also felt for it^^
with pid turned off and using manual fanspeed you can choose the full temp range.



> Another finding: The enhanced graph is broken :(
> 
> Somebody fixed the graph to be recorded in the background and not just live when visiting the dashboard page, BUT @KillerInk that destroyed your enhanced version of it. :(

yes thats intended https://github.com/bitaxeorg/ESP-Miner/pull/1152#issuecomment-3109864604
i think its made this way for mobile users. @terratec did a good job

autotune ignores the settings page. it increase voltage and frequency aslong its inside the fan, temp, power ,voltage and frequency limits. it can also run with manual fanspeed as written above.
from my view it is independent from the settings page

kakulukia:
So, after a week of testing it works pretty well and i find myself mostly just using the fan slider because in the end the noise level is all i care about after autotune takes care of handling all the rest.

Well done!

I would still vote for separating the settings between maybe "enabling autotune" + "max fan speed" and all the rest hidden under advanced, but in the end its just an oppinion. All i actually care about is: With autotune on there was not a single overheating event - without it i had quite a few.

main/tasks/power_management_task.h
```diff
@@ -10,7 +12,8 @@ typedef struct
     float chip_temp2_avg;
     float vr_temp;
     float voltage;
-    float frequency_value;
+    double core_voltage;
+    double frequency_value;
```
mutatrum: float (2x), no need for doubles.

main/thermal/auto_tune.h
```diff
@@ -0,0 +1,36 @@
+#ifndef AUTO_TUNE_H_
+#define AUTO_TUNE_H_
+#include <stdbool.h>
+#include <stdint.h>
+#include "global_state.h"
+
+typedef struct
+{
+    double power_limit;
```
mutatrum: Same here, try to avoid doubles unless it's absolutely necessary. There's no hardware double arithemtic on the ESP, so it's really, really slow.

main/nvs_config.c
```diff
@@ -195,6 +195,84 @@ void nvs_config_set_float(const char *key, float value)
...
+{
+    nvs_handle handle;
+    esp_err_t err;
+    err = nvs_open(NVS_CONFIG_NAMESPACE, NVS_READWRITE, &handle);
+    if (err != ESP_OK) {
+        ESP_LOGW(TAG, "Could not open nvs");
+        return;
+    }
+
+    err = nvs_set_blob(handle, key, &value, sizeof(bool));
```
mutatrum: booleans are currently done with u16, and write 0 or 1. For simplicity sake, it's probably better to stick to that scheme.

main/nvs_config.c
```diff
@@ -195,6 +195,84 @@ void nvs_config_set_float(const char *key, float value)
     nvs_config_set_string(key, str_value);
 }
 
+double nvs_config_get_double(const char * key, const double default_value)
```
mutatrum: Use floats, which are stored as strings. That makes it easier to supply a value from a factory firmware. With blob, there's no mechanism in place to put this in the config.cvs files when doing a factory flash.

main/tasks/power_management_task.h
KillerInk: the thing is with float we get as sample 600.00000013 instead of 600. and pid is also using double^^
that change with every write and triggers that the frequency or/and voltage get set on every cycle also if nothing changed.

KillerInk:
> When the NVS type changes, I'm not 100% sure how it will handle if you try to read a field of a different type (f.e. currently stored as blob, but read as string).

nothing happen, it fails to read and set default values. after flashing autotune disabled with default settings.

about floats. 
Unbenannt-1


and log spam is back^^
I (394315) auto_tune: Hashrate 1382.508015 Voltage 1089.915283 Frequency 553.701782
I (394323) power_management: set vcore voltage from 1089.815308mV to 1089.915283mV

mutatrum:
There's two things: the number of digits on the UI and in the logs is a formatting issue, that needs limiting with `%.2f` or how many decimals are needed.

If there's a comparison, take a look at `pll.c`. And floating point, both `float` as well as `double` needs a comparison with a delta:
```
#define EPSILON 0.0001f

if (fabs(curr_diff - min_diff) < EPSILON) {
```
So if the requested change is smaller than `EPSILON`, ignore it. That should solve these issues.

chillingbee:
Screenshot_20251029_060837_Brave

I got this from your latest build...

chillingbee:
The autotune page is back now but its not tuning...

chillingbee:
Yeah! Back in business 💪

chillingbee:
Just for information: its not showing the actual hashrate...
Screenshot_20251029_193017_Brave

mutatrum:
> Just for information: its not showing the actual hashrate... !

That's unrelated, that will be fixed with #1301

KillerInk:
Currently you get sometimes 0 hash rate when frequency change. But it continue hashing,

chillingbee:
Screenshot_20251103_065258_Brave
The graph is very "flat" now...

KillerInk:
Yes that's due the hash rate register. It shows the "real" hash rate. Befor it took the difficulty over time to calc the hash rate. But that's not caused by this pr

main/http_server/axe-os/src/app/layout/app.menu.component.ts
```diff
@@ -1,78 +1,47 @@
...
+          { label: 'Swarm', icon: 'pi pi-fw pi-sitemap', routerLink: ['swarm'] },
+          { label: 'Logs', icon: 'pi pi-fw pi-list', routerLink: ['logs'] },
+          { label: 'System', icon: 'pi pi-fw pi-wave-pulse', routerLink: ['system'] },
+          { separator: true },
+
+          { label: 'Pool', icon: 'pi pi-fw pi-server', routerLink: ['pool'] },
+          { label: 'Network', icon: 'pi pi-fw pi-wifi', routerLink: ['network'] },
+          { label: 'Theme', icon: 'pi pi-fw pi-palette', routerLink: ['design'] },
+          { label: 'Settings', icon: 'pi pi-fw pi-cog', routerLink: ['settings'] },
                     { label: 'Autotune', icon: 'pi pi-fw pi-sliders-h', routerLink: ['autotune'] },
```
duckaxe: Code indentation ;)

KillerInk: i see^^ there are more

main/http_server/http_server.c
duckaxe: ... and to `openapi.yaml`

main/http_server/openapi.yaml
```diff
@@ -853,3 +853,188 @@ paths:
...
+            example:
+              power_limit: 120.5
+              fan_limit: 80
+              max_voltage_asic: 1.3
+              max_frequency_asic: 2400
+              max_temp_asic: 75
+              auto_tune_hashrate: true
+              overshot_power_limit: 160
+              overshot_fanspeed: 90
+              max_temp_vr: 80
```
0xf0xx0: nit: nudge up to `components.schemas`, mayb named `AutotuneRequest`? see the other endpoints  
reasoning is it produces shorter codegen names and is easier to find

main/http_server/openapi.yaml
```diff
@@ -853,3 +853,188 @@ paths:
...
+                - max_temp_vr
+            example:
+              power_limit: 120.5
+              fan_limit: 80
+              max_voltage_asic: 1.3
+              max_frequency_asic: 2400
+              max_temp_asic: 75
+              auto_tune_hashrate: true
+              overshot_power_limit: 160
+              overshot_fanspeed: 90
```
0xf0xx0: nit: overshot -> overshoot

main/http_server/openapi.yaml
```diff
@@ -853,3 +853,188 @@ paths:
           description: Unauthorized - Client not in allowed network range
         '500':
           description: Internal server error
+
+  /api/system/autotune:
+    post:
```
0xf0xx0, line 973: minor nit: could the endpoint be PATCH like the axe config endpoint?

main/http_server/openapi.yaml
```diff
@@ -853,3 +853,188 @@ paths:
...
+                required:
+                  - power_limit
+                  - fan_limit
+                  - max_voltage_asic
+                  - max_frequency_asic
+                  - max_temp_asic
+                  - auto_tune_hashrate
+                  - overshot_power_limit
+                  - overshot_fanspeed
+                  - max_temp_vr
```
0xf0xx0: same nit here, named `AutotuneResponse`? just `Autotune`?

main/nvs_config.h
```diff
@@ -66,7 +66,20 @@ typedef enum {
...
+    NVS_CONFIG_KEY_FAN_LIMIT,
+    NVS_CONFIG_KEY_MAX_VOLTAGE_ASIC,
+    NVS_CONFIG_KEY_MAX_FREQUENCY_ASIC,
+    NVS_CONFIG_KEY_MAX_TEMP_ASIC,
+    NVS_CONFIG_KEY_AUTO_TUNE_ENABLE,
+    NVS_CONFIG_KEY_OVERSHOT_POWER_LIMIT,
+    NVS_CONFIG_KEY_OVERSHOT_FAN_LIMIT,
+    NVS_CONFIG_KEY_MAX_TEMP_VR,
+    NVS_ALL_CONFIG_COUNT
+
```
0xf0xx0, line 86: this feels weird but idk

main/http_server/axe-os/src/models/ISystemInfo.ts
```diff
@@ -67,6 +67,7 @@ export interface ISystemInfo {
     isUsingFallbackStratum: boolean,
     poolAddrFamily: number,
     frequency: number,
+    frequencySet: number,
```
0xf0xx0: whats this for?

KillerInk: Its for the settings to show the frequency from the config and not that one used by auto tune.

main/nvs_config.h
KillerInk: Its there that http server dont have to check not used keys for the info

main/http_server/axe-os/src/models/ISystemInfo.ts
KillerInk: grafik
without it it looks like this

KillerInk: grafik
with it like this

KillerInk:
> seems to not pay attention to when auto fan control is enabled, switched to manual at 90% and it started dropping the clock
> Image

yes its the old problem. i gonna workaround inside autotune

main/http_server/openapi.yaml
KillerInk: there are more^^

main/http_server/axe-os/src/models/ISystemInfo.ts
0xf0xx0: hmm, i think i prefer the furst as its easier to adjust the autotune if it gets confused. i got it into a state where it was holding a high voltage and low clock to reach the target temp instead of doing the opposite.

KillerInk: that still do work. i only show the saved frequency in nvs as it would be expected. when you set the frequency it gets applied.

0xf0xx0:
10m avg tuning is a lot more stable :D im lovin it

kakulukia:
Whats currently missing for this branch to be merged?

mutatrum:
> Whats currently missing for this branch to be merged?

Time to review. It's a really big change and it needs proper attention. You can help by also reviewing the code.

WantClue:
@KillerInk why closed ?

KillerInk:
I'm bored and dont think its desired. Good luck with the project.

kakulukia:
:`(

kakulukia:
That is very sad - does someone have backup copy of that branch? Mine is a bit too old - from November. :(

Maybe you, @0xf0xx0 ?

mutatrum:
You can still download the artifacts from the last build: https://github.com/bitaxeorg/ESP-Miner/actions/runs/20566490684/job/59065559551

Just make sure to unzip `esp-miner.bin` and `www.bin`.

We have a bunch of fairly big PRs lined up at the moment, but maybe someone else wants to pick up where KillerInk left it. It's opvious people liked the work.

kakulukia:
That helps a bit, but im more interested in the code as i was thinking about continuing the work of the lil Padawan who was not patient enough, but did a great work already!

0xf0xx0:
@kakulukia i still have it, pushed to my repo https://github.com/0xf0xx0/ESP-Miner/tree/master_autoclock  
3 months old, c279135  
idk why github is not notifying me when i get mentioned :\

kakulukia:
Thanks @0xf0xx0 !

0xf0xx0:
btw @kakulukia, are you in the osmu discord?

kakulukia:
nope - why are you asking @0xf0xx0 ?

0xf0xx0:
@kakulukia itd make it easier to discuss your revival uwu

kakulukia:
You can find me there now .. pleasee DM me to get in contact.
grafik

Issue #1070, opened 2025-06-26
Remove GZIP compression for WOFF2 files


duckaxe:
I had already worked on it https://github.com/WantClue/ESP-Miner-WantClue/pull/1. The ESP part is still missing. It would be great to finish this one.

PR #1161, opened 2025-07-24
Remove GZIP handling for WOFF2 files

duckaxe:
I grabbed my old branch and adjusted it slightly...

> WOFF2 files are already compressed, so applying additional gzip or deflate compression to them typically has little to no effect. This is because the WOFF2 file format internally compresses the font data, making further compression redundant.

Since WOFF2 files are already compressed, it makes no sense to compress them again and have the browser decompress them. This PR adds an exception for gzip compression during the build process and on the web server side.

### After changes

**No gzip compression for .woff2**
Screenshot 2025-07-24 at 20 20 11

**No Content-Encoding:gzip for .woff2**
Screenshot 2025-07-24 at 20 07 19

**Tested on Bitaxe**

Closes https://github.com/bitaxeorg/ESP-Miner/issues/1070

### DIFF
--- a/main/http_server/http_server.c
+++ b/main/http_server/http_server.c
@@ -2,6 +2,7 @@
 #include <fcntl.h>
 #include <string.h>
 #include <sys/param.h>
+#include <sys/stat.h>
 
 #include "freertos/FreeRTOS.h"
 #include "freertos/event_groups.h"
@@ -355,10 +356,16 @@ static esp_err_t rest_api_common_handler(httpd_req_t * req)
     return ESP_OK;
 }
 
+static bool file_exists(const char *path) {
+    struct stat buffer;
+    return (stat(path, &buffer) == 0);
+}
+
 /* Send HTTP response with the contents of the requested file */
 static esp_err_t rest_common_get_handler(httpd_req_t * req)
 {
     char filepath[FILE_PATH_MAX];
+    char gz_file[FILE_PATH_MAX];
     uint8_t filePathLength = sizeof(filepath);
 
     rest_server_context_t * rest_context = (rest_server_context_t *) req->user_ctx;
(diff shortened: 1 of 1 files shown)


main/http_server/axe-os/only-gzip.js
```diff
@@ -16,8 +16,8 @@ function processDirectory(dirPath) {
                 if (stats.isDirectory()) {
                     // If it's a directory, process it recursively
                     processDirectory(filePath);
-                } else if (!file.endsWith('.gz')) {
-                    // If it's a file and doesn't end with .gz, unlink it
+                } else if (!['.gz', '.woff2'].some(ext => file.endsWith(ext))) {
```
mutatrum: This check can be smarter. If there's a `.gz` file, delete the non-compressed version. This way we don't need exceptions here for every extension.

main/http_server/http_server.c
```diff
@@ -369,7 +369,12 @@ static esp_err_t rest_common_get_handler(httpd_req_t * req)
...
     }
     set_content_type_from_file(req, filepath);
-    strcat(filepath, ".gz");
+
+    bool is_woff2 = CHECK_FILE_EXTENSION(filepath, ".woff2");
+
+    if (!is_woff2) {
+        strcat(filepath, ".gz");
+    }
     int fd = open(filepath, O_RDONLY, 0);
```
mutatrum: Same smartness here: see if a `.gz` exists, append extension and add `Content-Encoding` header, otherwise, don't.

duckaxe:
> With generalised handling, we can also do this for the pdf. That also doesn't compress.

```
gzipper: File bitcoin.pdf has been compressed. 
Algorithm: GZIP 
Size: 179.97 KB -> 174.4 KB 
Time: 35.251251ms
```

Not much but compressed.

main/http_server/axe-os/only-gzip.js
duckaxe: Ok, refactored. Removed indents.

main/http_server/http_server.c
duckaxe: I cannot look for `.gz` because the `filepath` contains the raw file path **without** the .gz extension (like `main.2840eb6448fa06d2.js`). 

I renamed the variable from `is_woff2` to `ignore_gzip_handling` for more flexibility, so we can add more file extensions (like a whitelist) to this var.

Or do you have another idea?

duckaxe: > see if a .gz exists

Or you mean to check the .gz file exists? When yes, then we will have an `open` call more. I'm not sure if it's worth it.

mutatrum: `open` on a non-existant file is cheap.

duckaxe: Done.

main/http_server/http_server.c
```diff
@@ -369,8 +376,14 @@ static esp_err_t rest_common_get_handler(httpd_req_t * req)
         strlcat(filepath, req->uri, filePathLength);
     }
     set_content_type_from_file(req, filepath);
-    strcat(filepath, ".gz");
-    int fd = open(filepath, O_RDONLY, 0);
+
+    strlcpy(gz_file, filepath, filePathLength);
+    strlcat(gz_file, ".gz", filePathLength);
+
+    bool serve_gz = file_exists(gz_file);
```
mutatrum, line 383: 👍

mutatrum (approved):
LGTM, didn't test, only looked at code.

WantClue (approved):

PR #1162, opened 2025-07-24
Fix statistics deactivation without restart

terratec:
Steps to reproduce the issue:
- enable statistics
- wait for the API endpoints until at least 1 data point has been recorded
- disable statistics without restart
- the API endpoints show 1 data point with garbage data
- the dashboard is unhappy (endless loading and/or firefox complains about an unresponsive script)

Solution:
First step was to remove the guard `|| (0 == statsFrequency)` in `StatisticsNextNodePtr statisticData(StatisticsNodePtr nodeIn, StatisticsNodePtr dataOut)`. The function should always return data if input and output pointers are valid.
The next step was to remove all data with `void clearStatisticData()`.


WantClue:
please add description on you fix.
Looking through it rn

WantClue:
I cannot reproduce this issue

PR #1172, opened 2025-07-27
feat: test both temps gt

WantClue:
adds temp2 for asic2 to the webui
missing is the api info endpoint  (for now)

@duckAxe would be nice if you take a second look over this

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -204,7 +204,12 @@
 
                 <div class="mb-3">
                     <h6 class="flex justify-content-between mb-1">
-                        ASIC Temperature 1
+                        <ng-container *ngIf="info.temp2 > -1; else singleTemp">
+                            ASIC Temperature 1
+                        </ng-container>
+                        <ng-template #singleTemp>
+                            ASIC Temperature
+                        </ng-template>
                         <span>
                             <ng-container *ngIf="info.temp > 0; else noTemp">
                                 {{info.temp}} &deg;C
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -1,3 +1,5 @@
+<ng-template #noTemp>--</ng-template>
+
 <ng-container *ngIf="info$ | async as info">
     <!-- Temp warning alert -->
     <p-message *ngIf="info.overheat_mode" severity="error" styleClass="w-full mb-4 py-4 border-round-xl"
@@ -204,7 +206,6 @@
                     <h6 class="flex justify-content-between mb-1">
                         ASIC Temperature 1
                         <span>
-                            <ng-template #noTemp>--</ng-template>
                             <ng-container *ngIf="info.temp > 0; else noTemp">
                                 {{info.temp}} &deg;C
                             </ng-container>
@@ -223,8 +224,7 @@
                     <h6 class="flex justify-content-between mb-1">
                         ASIC Temperature 2
                         <span>
-                            <ng-template #noTemp2>--</ng-template>
-                            <ng-container *ngIf="info.temp2 > 0; else noTemp2">
+                            <ng-container *ngIf="info.temp2 > 0; else noTemp">
                                 {{info.temp2}} &deg;C
                             </ng-container>
                         </span>
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -202,7 +202,7 @@
 
                 <div class="mb-3">
                     <h6 class="flex justify-content-between mb-1">
-                        ASIC Temperature
+                        ASIC Temperature 1
                         <span>
                             <ng-template #noTemp>--</ng-template>
                             <ng-container *ngIf="info.temp > 0; else noTemp">
@@ -215,6 +215,25 @@
                     </p-progressBar>
 
                     <strong class="text-red-500" *ngIf="info.temp >= 70">
+                        Danger: High Temperature
+                    </strong>
+                </div>
+
+                <div class="mb-3" *ngIf="info.temp2 > -1">
+                    <h6 class="flex justify-content-between mb-1">
+                        ASIC Temperature 2
+                        <span>
+                            <ng-template #noTemp2>--</ng-template>
+                            <ng-container *ngIf="info.temp2 > 0; else noTemp2">
+                                {{info.temp2}} &deg;C
+                            </ng-container>
+                        </span>
+                    </h6>
+                    <p-progressBar [value]="(info.temp2 / maxTemp) * 100" [styleClass]="'p-progressbar--thin'">
+                        <ng-template pTemplate="content" let-value></ng-template>
+                    </p-progressBar>
+
+                    <strong class="text-red-500" *ngIf="info.temp2 >= 70">
                         Danger: High Temperature
                     </strong>
                 </div>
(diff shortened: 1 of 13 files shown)


WantClue:
image

WantClue:
WIP still finalizing

main/http_server/axe-os/src/app/components/home/home.component.html
```diff
@@ -202,9 +204,8 @@ <h4 class="text-center">Heat</h4>
 
                 <div class="mb-3">
                     <h6 class="flex justify-content-between mb-1">
-                        ASIC Temperature
+                        ASIC Temperature 1
```
johnny9: Can we do something like 
```
<h6 class="flex justify-content-between mb-1">
  <span *ngIf="info.temp2 > -1; else single">
    ASIC Temperature 1
  </span>
  <ng-template #single>ASIC Temperature</ng-template>
  …
</h6>
```

To keep the original name if only one sensor is available?

main/tasks/power_management_task.c
```diff
@@ -83,6 +83,15 @@ void POWER_MANAGEMENT_task(void * pvParameters)
...
         power_management->fan_rpm = Thermal_get_fan_speed(&GLOBAL_STATE->DEVICE_CONFIG);
         power_management->chip_temp_avg = Thermal_get_chip_temp(GLOBAL_STATE);
+        
+        // Get both temperatures for EMC2103
+        if (GLOBAL_STATE->DEVICE_CONFIG.EMC2103) {
+            EMC2103_temps_t temps = Thermal_get_chip_temps(GLOBAL_STATE);
+            power_management->chip_temp_avg = temps.temp1;
+            power_management->chip_temp2_avg = temps.temp2;
+        } else {
+            power_management->chip_temp2_avg = -1; // Invalid for non-EMC2103 devices
```
johnny9: I think overloading the value -1 to mean not available isn't great. Its better to have an explicit bool for if its supported, or a num_temp_sensors variable that has how many sensors and then return a list of values.

main/http_server/axe-os/src/app/components/home/home.component.html
WantClue: sure that can be done

main/tasks/power_management_task.c
WantClue: ACK will be changed

WantClue: @johnny9 I think with the most recent changes this should be resolved. Adding an boolean is not necessary IMO

main/tasks/power_management_task.c
```diff
@@ -83,6 +83,15 @@ void POWER_MANAGEMENT_task(void * pvParameters)
 
         power_management->fan_rpm = Thermal_get_fan_speed(&GLOBAL_STATE->DEVICE_CONFIG);
         power_management->chip_temp_avg = Thermal_get_chip_temp(GLOBAL_STATE);
+        
+        // Get both temperatures for EMC2103
+        if (GLOBAL_STATE->DEVICE_CONFIG.EMC2103) {
```
mutatrum: This logic should be inside `thermal.c`. the cleanest way would be to define number of temp sensors in `device_config` and have the method in thermal return an array of temps. Now, this case causes exceptions throughout the flow.

main/http_server/axe-os/src/app/components/home/home.component.html
```diff
@@ -202,9 +204,13 @@ <h4 class="text-center">Heat</h4>
 
                 <div class="mb-3">
                     <h6 class="flex justify-content-between mb-1">
-                        ASIC Temperature
+                        <ng-container *ngIf="info.temp2 > 0; else singleTemp">
+                            ASIC Temperature 1
```
mutatrum, line 208: One header with multiple bars might be cleaner.

main/http_server/axe-os/src/app/services/system.service.ts
```diff
@@ -28,6 +28,7 @@ export class SystemService {
         voltage: 5208.75,
         current: 2237.5,
         temp: 60,
+        temp2: 60,
```
mutatrum, line 31: Array of temp

main/thermal/thermal.h
```diff
@@ -13,5 +13,6 @@ esp_err_t Thermal_set_fan_percent(DeviceConfig * DEVICE_CONFIG, float percent);
 uint16_t Thermal_get_fan_speed(DeviceConfig * DEVICE_CONFIG);
 
 float Thermal_get_chip_temp(GlobalState * GLOBAL_STATE);
+EMC2103_temps_t Thermal_get_chip_temps(GlobalState * GLOBAL_STATE);
```
mutatrum: This method leaks hardware configuration.

mutatrum (changes requested):
Added some comments, I think it can be a lot cleaner and future resistant to make the temp readings an array.

Hope it makes sense, review was done on phone.

main/http_server/axe-os/src/app/components/home/home.component.html
mutatrum: If it's and array, this can loop over values, also futurproofing the frontend.

WantClue:
> Added some comments, I think it can be a lot cleaner and future resistant to make the temp readings an array.
> 
> Hope it makes sense, review was done on phone.
> 
> Some more thoughts:
> 
> * How does fan control work with 2 temperatures? Should it take the max of both? For the GT it's one fan for both ASICs IINM.
> * If we extend it to an array of temps, we could also add the VR temperature to this. And open the door to multiple fan controllers as well, that are governed by different temperature sensors.

I think it makes no sense for this PR to add an array for storing it, I do see the benefit but this would be out of scope of this PR and would add complexity to the UI and we would need to rewrite certain logics in the UI to handle that. I would take this for a future PR and leave this PR soley for enabling the second reading

main/http_server/axe-os/src/app/services/system.service.ts
WantClue: Not in this PR

main/thermal/thermal.h
WantClue: fixed in the latest commit

WantClue:
> Added some comments, I think it can be a lot cleaner and future resistant to make the temp readings an array.
> 
> Hope it makes sense, review was done on phone.
> 
> Some more thoughts:
> 
> * How does fan control work with 2 temperatures? Should it take the max of both? For the GT it's one fan for both ASICs IINM.
> * If we extend it to an array of temps, we could also add the VR temperature to this. And open the door to multiple fan controllers as well, that are governed by different temperature sensors.

ye the fan control probably should work with the avg of both temps for the gt, might need a bit of rework

main/tasks/power_management_task.c
WantClue: resolved

WantClue:
@mutatrum added overheat protection for the second chip as well, and changed the decision logic over to thermal where it should belong.  please check

mutatrum (approved):
Much better, LGTM.

mutatrum:
> ye the fan control probably should work with the avg of both temps for the gt, might need a bit of rework

I was thinking about this, maybe max of both would be better. If one asic stops hashing for whatever reason, the avg. can be tricky.

WantClue:
> > ye the fan control probably should work with the avg of both temps for the gt, might need a bit of rework
> 
> 
> 
> I was thinking about this, maybe max of both would be better. If one asic stops hashing for whatever reason, the avg. can be tricky.

Yes this probably also falls under the system monitoring that skot had in mind a year ago. We would need to create a major monitoring task that keeps track of everything. If one thing goes bad we want to stop.

PR #1175, opened 2025-07-29
Introduction Tooltip Components

duckaxe:
### Current state
Tooltips do not work on mobile devices. To solve this issue, I created two new components. Also to make it easier to integrate the tooltip icon in our code.

### 1. Icon with a tooltip `<tooltip-icon />`
```
<tooltip-icon 
    tooltip="Name of the device on your local network" 
/>
```
This is useful when a tooltip icon needs to be placed inside the copy as a single inline element.

### 2. Text + Icon with a tooltip `<tooltip-text-icon />`
```
<tooltip-text-icon
    text="Hostname"
    tooltip="Name of the device on your local network"
/>
```
Icon with a tooltip that is attached to the text (or rather, to the last word so that the icon does not wrap to a new line). This component make it easier for developers to add help icon with tooltip to form labels and headlines without worrying about the necessary HTML markup. The preferred way.

### Layout
I researched the best practices for displaying tooltips on smartphones. As result, on icon tap the tooltip will be displayed at the bottom of the screen. This allows us to display a lot of text while remaining user-friendly and readable. The user can close the tooltip/layer by taping on the close button or outside the layer.

To demonstrate how it works, I added help icons with tooltips to the dashboard (shares rejection, see screens below) and the network view (hostname label). In general, we should inject more hints and integrate help icons in AxeOS. We can do that in a separate PR. This will reduce the number of support requests.

**Desktop (tooltip on hover)**
Desktop

**Mobile (tooltip on tap)**

https://github.com/user-attachments/assets/fab18c15-0383-4c3d-94fb-382b838afacb

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -345,7 +345,10 @@
 
                 <div class="flex flex-nowrap">
                     <div class="col-fixed-width text-500">
-                        <span>URL <i class="pi pi-info-circle text-xs px-1" pTooltip="{{ getPoolProtocolType(info) }}" tooltipPosition="top"></i></span>
+                        <tooltip-text-icon
+                            text="URL"
+                            tooltip="{{ getPoolProtocolType(info) }}"
+                        />
                     </div>
                     <div>
                         <span>
@@ -354,9 +357,12 @@
                     </div>
                 </div>
 
-                <div class="flex flex-nowrap ellipsis-magic overflow-hidden" pTooltip="{{activePoolUser}}" tooltipPosition="top">
+                <div class="flex flex-nowrap ellipsis-magic overflow-hidden">
                     <div class="col-fixed-width text-500">
-                        <span>User</span>
+                        <tooltip-text-icon
+                            text="User"
+                            tooltip="{{ activePoolUser }}"
+                        />
                     </div>
 
                     <span class="start" sensitive-data>


main/http_server/axe-os/src/app/components/tooltip-text-icon/tooltip-text-icon.component.ts
```diff
@@ -0,0 +1,33 @@
+import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+
+@Component({
+  selector: 'tooltip-text-icon',
+  templateUrl: './tooltip-text-icon.component.html',
+})
+export class TooltipTextIconComponent implements OnChanges {
+  @Input() tooltip: string | null = '';
+  @Input() text: string | null = '';
+  @Input() split: boolean = true;
```
mutatrum, line 10: What is split doing?

duckaxe: By default, the tooltip text is split. The last word of the tooltip is placed in a `white-space-nowrap` container together with the icon so that the icon is wrapped to a new line together with this word. There are edge cases where this is not desired. This is what split = false is for.

mutatrum:
Are the files in `tooltip-icon` used at all?

duckaxe:
> Are the files in `tooltip-icon` used at all?

[Internally](https://github.com/duckaxe/ESP-Miner-Bitaxe/blob/649e7763b8c876a2fb60630764c7c89dd8a5dde7/main/http_server/axe-os/src/app/components/tooltip-text-icon/tooltip-text-icon.component.html) and maybe in the future.

PR #1177, opened 2025-08-02
Send valid JSON-RPC 2.0 requests

jirijakes:
This PR adds `jsonrpc` member to all messages to comply with JSON-RPC 2.0 [specification](https://www.jsonrpc.org/specification#request_object):

> The Request object has the following members:
>     **jsonrpc**: A String specifying the version of the JSON-RPC protocol. MUST be exactly "2.0".
>     …

### DIFF
--- a/components/stratum/stratum_api.c
+++ b/components/stratum/stratum_api.c
@@ -368,7 +368,7 @@ int STRATUM_V1_subscribe(int socket, int send_uid, const char * model)
     char subscribe_msg[BUFFER_SIZE];
     const esp_app_desc_t *app_desc = esp_app_get_description();
     const char *version = app_desc->version;	
-    sprintf(subscribe_msg, "{\"id\": %d, \"method\": \"mining.subscribe\", \"params\": [\"bitaxe/%s/%s\"]}\n", send_uid, model, version);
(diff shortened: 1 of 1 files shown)


mutatrum:
I'm not sure the Stratum API follows the json-rpc spec?

jirijakes:
> I'm not sure the Stratum API follows the json-rpc spec?

The widely linked specification at https://reference.cash/mining/stratum-protocol claims so. But I am not sure whether that is generally accepted or not.

mutatrum:
I've never seen that one. Not claiming that it's not a valid source, but I only know https://en.bitcoin.it/wiki/Stratum_mining_protocol. I also don't know what in the new BIP-40/BIP-41 is proposed?

jirijakes:
> I only know https://en.bitcoin.it/wiki/Stratum_mining_protocol

Seems like the page behind "Original example on slush's pool site" link in this document also says JSON-RPC 2.0: https://web.archive.org/web/20150307191254/http://mining.bitcoin.cz/stratum-mining#developers

[BIP40 (draft)](https://github.com/bitcoin/bips/pull/1557) says both JSON-RPC 1.0 and 2.0 are acceptable: https://github.com/ben221199/bips/blob/add-bip40/bip-0040.mediawiki#specification-data-format . But IIUC that is not for mining. I think BIP41 draft is not yet written.

ben221199:
The BIP41 draft indeed is not written yet. I will then investigate if both JSON-RPC 1.0 or 2.0 are accepted.

PR #1178, opened 2025-08-02
Bitaxe accessory port

### LINKED ISSUE #756: Bitaxe Accessory Port (BAP) Protocol
BAP protocol is 115200 baud (default). 
format is standard NMEA sentences (in ASCII). Connected accessories can "ask" for any parameter available over the web API. 
accessories can also "subscribe" to receive regular notifications on selected parameters.

WantClue:
This finally introduces the BAP to the Bitaxe. Giving it the option to control and read out values from the Bitaxe via UART.

The BAP supports a subscription based system, where a client device can subscribe to specific statistics that will be send over UART every 5 seconds (this might be reduced in the future).
Furthermore does it enable clients to send requests for specific information from the Bitaxe and a SET for setting values like the fan speed, wifi credentials.

closes #756 

WIP:
- [x] removing all the debug logging of BAP
- [x] add different start up init messages for AP mode and connected mode
- [x] readme

### DIFF
--- a/main/bap/bap_uart.c
+++ b/main/bap/bap_uart.c
@@ -113,19 +113,6 @@ void BAP_send_ap_message(GlobalState *state) {
     BAP_send_message(BAP_CMD_CMD, "mode", "ap_mode");
 }
 
-void BAP_ap_mode_task(void *pvParameters) {
-    GlobalState *state = (GlobalState *)pvParameters;
-    
-    ESP_LOGI(TAG, "BAP AP mode task started - sending AP messages every 5 seconds");
-    
-    while (1) {
-        BAP_send_ap_message(state);
-        
-        vTaskDelay(pdMS_TO_TICKS(5000));
-    }
-    
-    vTaskDelete(NULL);
-}
 
 static void uart_receive_task(void *pvParameters) {
     uint8_t *data = (uint8_t *) malloc(BAP_BUF_SIZE);
--- a/main/bap/bap_uart.h
+++ b/main/bap/bap_uart.h
@@ -59,11 +59,6 @@ void BAP_send_init_message(GlobalState *state);
  */
 void BAP_send_ap_message(GlobalState *state);
 
-/**
- * @brief AP mode task that sends periodic AP messages
- * @param pvParameters Global state pointer
- */
-void BAP_ap_mode_task(void *pvParameters);
 
 /**
  * @brief Start UART receive task


main/main.c
```diff
@@ -86,6 +87,15 @@ void app_main(void)
 
     SERIAL_init();
 
+    // Initialize BAP interface
+    esp_err_t bap_ret = BAP_init(&GLOBAL_STATE);
```
mutatrum: Should this be behind a feature toggle? E.g. enable BAP somewhere in config or AxeOS?

mutatrum:
Only first high level look at the code, needs more eyes. Documentation is missing, so that makes it a bit tricky to review ATM. Not sure what you landed on from the issue discussion.

When looking at the ESP docs on UART, I saw this: https://github.com/espressif/esp-idf/tree/v5.5/examples/peripherals/uart/nmea0183_parser, which might be a nice reference to look at, especially the rx pattern interrupt method. Probably outside this PR, though.

main/main.c
WantClue: This is atm the only bap entry, there will be two different ones depending on the state of axeos. I don't know if a toggle in the ui would be helpful or the right approach ideally you want to control the Bitaxe with the bap without any ui interaction

eandersson: I agree that having a feature flag here seems useful. It could just be an nvs config that defaults to on.

eandersson: It could help in the future to narrow down if issues (e.g. memory leaks) are related to this feature as well.

skot:
Since BAP data is opt-in, is there any downside to have it always on?

WantClue:
> Since BAP data is opt-in, is there any downside to have it always on?

atm it's always on, but never sends out anything if the bap_client is not sending a request or subscription. so bap needs to be activated essentially by the client. This saves resources IMO. 

I will finalize the readme for BAP today

main/bap/bap_uart.c
```diff
@@ -0,0 +1,298 @@
...
+    } else {
+        //ESP_LOGI(TAG, "Init message sent: %s", init_message);
+    }
+}
+
+void BAP_send_ap_message(GlobalState *state) {
+    BAP_send_message(BAP_CMD_CMD, "mode", "ap_mode");
+}
+
+void BAP_ap_mode_task(void *pvParameters) {
```
eandersson: Is this used anywhere?

WantClue: It's been send if the device is in ap mode.
L271 bap_subscription.c

main/main.c
Notmyprob487-png: What  "(memory leaks)" could you use?

PR #1004, opened 2025-06-04
Generate sdkconfig.defaults

mutatrum:
Only non-default values are now stored in `sdkconfig.defaults`. This also fixes the issue of using the wrong board after checkout or a full clean. When something changes, `idf.py save-defconfig` needs to be run again. This should not change anything in the build.

PR #1182, opened 2025-08-08
add a shell script to prepare the build env after an esp-idf update

0xf0xx0:
requested by mutatrum, will likely be superseded once #1004 is merged but for now it'll work uwu


mutatrum:
This is indeed obsolete with #1004, that sets the board version to S3 by default.

WantClue:
Merged #1004 so we can close this @0xf0xx0 ?

0xf0xx0:
yep, this can be closed uwu

PR #1064, opened 2025-06-23
Extranonce subscribe toggle and suggested difficulty pool configuration

mutatrum:
Added per pool settings for extranonce subscribe and suggested difficulty.

Added cards for primary and fallback pool configuration

`mining.set_extranonce` messages are not handled yet. Nicehash wants KYC.

Fixes #193

image

PR #1187, opened 2025-08-11
Switch pool on dashboard 

duckaxe:
The dashboard now includes a feature that allows you to switch to the fallback pool (and vice versa). Ideal if you want to switch quickly to the fallback pool. Stored on NVS.

Colors
* Green = Primary Pool
* Orange = Fallback Pool

**Video / Mobile**

https://github.com/user-attachments/assets/8db11261-b437-4b3a-aed6-34f90ac9f743

**Desktop / Primary Pool**
Primary

**Desktop / Fallback Pool**
Fallback

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,5 +1,5 @@
 # Stage 1: Base image with ESP-IDF and common dependencies
-FROM espressif/idf:v5.5 AS base
+FROM espressif/idf:v5.5.1 AS base
 
 ARG DEBIAN_FRONTEND=nointeractive
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
     - name: esp-idf build
       uses: espressif/esp-idf-ci-action@v1
       with:
-        esp_idf_version: v5.5
+        esp_idf_version: v5.5.1
         target: esp32s3
         command: GITHUB_ACTIONS="true" idf.py build
         path: '.'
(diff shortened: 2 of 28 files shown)


mutatrum:
What happens when pool is swapped and the fallback pool is not available?

duckaxe:
> What happens when pool is swapped and the fallback pool is not available?

Then it swaps automagical to the primary pool. I just tested it.

mutatrum:
Not sure about this, seems confusing as the naming doesn't change. A way simpler option would be to have a switch button on the pool config, where the two pool configs just swap places.

duckaxe:
That's right. It would be a simple option, but not future-proof. If [additional fallback pools were added later](https://github.com/bitaxeorg/ESP-Miner/pull/1064#issuecomment-3005703408), you would have to rethink and redesign this function. With my suggestion the amount of pools wouldn't matter because any amount could be added to the dropdown and just selected by the user. No UI changes would be necessary.

It's just an idea that we don't have to merge ;)

mutatrum:
With pools > 2 the UX can be done with reorder buttons (move up/move down), so it can be solved front-end only.

image

duckaxe:
Possible, but more complicated if you just want to use a specific pool. In your example, I first have to reorder. In my example, I just select.

duckaxe:
The drawback of my solution is that it requires a firmware change. Swapping field values is front-end only. For simplicity we can do the field values swap as you suggested.

mutatrum:
> The drawback of my solution is that it requires a firmware change. Swapping field values is front-end only. For simplicity we can do the field values swap as you suggested.

With more pools, using a selector to switch to a non-primary pool leaves the rest of priority order undefined, or at least, difficult to solve in UI. By flipping them in configuration, this behaviour is clear. It also fits nicely in the rest of the UX with a notification that a restart is needed and a manual restart. The device shouldn't restart implicitly.

And secondary: getting the fallback code working properly was quite tricky.

WantClue (approved):
LGTM

PR #1189, opened 2025-08-13
Clean up module naming (extract from #1098)

mutatrum:


eandersson (approved):

mutatrum:
Don't merge as it's conflicting with #1146, want to wait on the outcome of that one.

PR #1193, opened 2025-08-15
Quicklink service: Fix JS error when using local pool

duckaxe:
**Current state**
An error appears on the dashboard as soon as a local pool is defined as the Stratum host.

```
TypeError: Cannot read properties of null (reading '1')
    at main.84746aa7a21cd575.js:1:616665
    at String.replace (<anonymous>)
    at t.getQuickLink ()
```

**Solution**
Add a match check to see if quick links are being used.


WantClue (approved):
code LGTM, will be merged and we need to keep an eye on the quicklinks on the next beta

PR #1170, opened 2025-07-26
Add Pool Difficulty to Dashboard

duckaxe:
This PR adds pool difficulty to the Pool card.

**Desktop**
Screenshot 2025-07-27 at 07 32 36

**Mobile**
Screenshot 2025-07-27 at 07 32 54

PR #1200, opened 2025-08-20
Show block header, scriptsig and network difficulty

### LINKED ISSUE #1011: Show block header information in Axe-OS
Seeing https://x.com/boerst/status/1930754454451847375, it would make a lot of sense to show block header information on Axe-OS somewhere. Maybe a full clone of [stratum.work](stratum.work) is overkill, but at least show the miner tag (not sure what it's called) and possibly the merkle tree buildup would be interesting and educational.

### LINKED ISSUE #1166: add BIP-34 block height parsing to esp-miner and display on AxeOS
[BIP-34](https://github.com/bitcoin/bips/blob/master/bip-0034.mediawiki) has been activated for over a decade, so it's probably safe to say it's supported by all pools/stratum servers.

Part of this change adds the current Bitcoin block height into the coinbase transaction of blocks. Mining pools will send this as a part of block template in the `mining.notify` stratum message. We should be able to parse this out and keep track of the current block height for display on the AxeOS dashboard.

Of course we'll need proper error handling in for the inevitable situation that a stratum server doesn't send it  😅

mutatrum:
Replaced the Uptime card with Block Header:

image

Same info on the screen:

image

If people still want the Uptime card, that can be restored. But IMO another row of cards is a bit much, or we have to condense them further.

Fixes #1011
Fixes #1166


skot:
maybe add the payout coinbase address? And if this matches the one configured as the stratum user? (This would only work for solo mining)

mutatrum:
> maybe add the payout coinbase address? And if this matches the one configured as the stratum user? (This would only work for solo mining)

I'm adding network difficulty ATM. Coinbase address(es) are also possible, but need a bech32 encoder for that. Will look into that later.

skot:
> > maybe add the payout coinbase address? And if this matches the one configured as the stratum user? (This would only work for solo mining)
> 
> I'm adding network difficulty ATM. Coinbase address(es) are also possible, but need a bech32 encoder for that. Will look into that later.

I seem to remember the NerdQAxe fimware has (or had) payout address parsing? Maybe it was in PyMiner?

duckaxe:
I would prefer to keep the uptime card and rename it to something generic like `Runtime`. This will be a card where we collect secondary information during miner runtime.

Miner's uptime and [max temperature](https://github.com/bitaxeorg/ESP-Miner/pull/888) can be placed here. `Block height` and `Miner tag` as additional rows.

duckaxe:
https://github.com/bitaxeorg/ESP-Miner/pull/1170/commits/2584842623c004d0effae0f6e7bf0d0ad3a4fd68 can be a draft/idea for that what I mean.

main/http_server/axe-os/src/models/ISystemInfo.ts
```diff
@@ -63,6 +63,9 @@ export interface ISystemInfo {
     boardtemp1?: number,
     boardtemp2?: number,
     overheat_mode: number,
-    power_fault?: string
-    overclockEnabled?: number
+    power_fault?: string,
+    overclockEnabled?: number,
+
+    blockHeight?: number,
```
duckaxe, line 69: Make sense to add both variables to mock data at system.service.ts

main/http_server/http_server.c
```diff
@@ -724,6 +724,12 @@ static esp_err_t GET_system_info(httpd_req_t * req)
         cJSON_AddStringToObject(root, "power_fault", VCORE_get_fault_string(GLOBAL_STATE));
     }
 
+    if (GLOBAL_STATE->block_height > 0) {
+        cJSON_AddNumberToObject(root, "blockHeight", GLOBAL_STATE->block_height);
+        cJSON_AddStringToObject(root, "minerTag", GLOBAL_STATE->miner_tag);
+        cJSON_AddStringToObject(root, "networkDifficulty", GLOBAL_STATE->network_diff_string);
```
duckaxe, line 730: We should deliver the value as number and format it in AxeOS. This way, we don't have to convert it later, as we did with [bestDiff](https://github.com/bitaxeorg/ESP-Miner/pull/1202).

0xf0xx0:
this fails to parse the tag from public-pool and pogolo because the code checks in coinbase part 2 and the tag is in part 1.

mutatrum:
> this fails to parse the tag from public-pool and pogolo because the code checks in coinbase part 2 and the tag is in part 1.

Should be working now. It parses both parts and replaces non-printable characters with a `.`

Made the font monospace, but the baseline doesn't align properly with the rest of the font. Needs css magic from :duck::axe:

0xf0xx0:
works :D i think its pretty cool with the unprintable chars being periods 
properly-parsed pogolo tag

minor nit, `Block Header` instead of `Block header`?

mutatrum:
> works :D i think its pretty cool with the unprintable chars being periods 

Thanks for testing with pogolo! Replace with periods is similar to how the genesis block is often depicted:

image

> minor nit, `Block Header` instead of `Block header`?

Done.

main/http_server/axe-os/src/app/components/home/home.component.html
```diff
@@ -348,19 +348,19 @@ <h5>Block header</h5>
...
                     </div>
-                    <div class="code">
-                        {{ info.minerTag }}
+                    <div>
+                        {{ info.networkDifficulty }}
                     </div>
                 </div>
 
-                <div class="flex flex-nowrap">
+                <div class="flex flex-nowrap code-item">
```
duckaxe: Pls use `align-items-baseline`

duckaxe:
My 2 cents:
1. https://github.com/mutatrum/ESP-Miner/pull/3
2. https://github.com/bitaxeorg/ESP-Miner/pull/1200#discussion_r2296540075

main/http_server/axe-os/src/app/services/system.service.ts
```diff
@@ -83,7 +83,11 @@ export class SystemService {
 
         boardtemp1: 30,
         boardtemp2: 40,
-        overheat_mode: 0
+        overheat_mode: 0,
+
+        blockHeight: 811111,
+        scriptsig: "..%..h..,H...ckpool.eu/solo.ckpool.org/",
+        networkDiffString: "25.3T",
```
duckaxe: `networkDifficulty`

mutatrum:
> My 2 cents:
> 
> 1. [Truncate Scriptsig mutatrum/ESP-Miner#3](https://github.com/mutatrum/ESP-Miner/pull/3)

The `line-break: anywhere;` is preferable IMO, as we don't know which part of the scriptsig is informative or not. This is the smallest I could get it:

image

> 2. [Show block header, scriptsig and network difficulty #1200 (comment)](https://github.com/bitaxeorg/ESP-Miner/pull/1200#discussion_r2296540075)

Depends which one is merged earlier. If I convert this to a number now, this PR also needs the formatter and the scope leaks. If 1202 is merged first, I'll adjust this one, of this one is merged first, 1202 has to be extended with this new field, but then both PRs are separated.

0xf0xx0:
if we're going to keep making small breaking changes we might as well fix everything in one go. my two cents are we should merge #1202 as a stopgap, then refactor the api and do the frontend in the process. theres no reason to have the api rework thruplicated.

duckaxe:
> The line-break: anywhere; is preferable IMO, as we don't know which part of the scriptsig is informative or not. This is the smallest I could get it:

I would keep it consistent. With the pool user, we don't know which part is informative/important, so we shorten it anyway. The full scriptsig can be seen in the tooltip. Up to you.

main/http_server/axe-os/src/app/components/home/home.component.html
```diff
@@ -324,11 +324,47 @@ <h5>Pool ({{ activePoolLabel }})</h5>
             </div>
         </div>
 
+        <!--
```
duckaxe: I would remove the unused code. If we need a new card, we'll create a new one. No dead code.

duckaxe (changes requested):

main/http_server/axe-os/src/app/components/home/home.component.html
mutatrum: Done

WantClue:
LGTM, love it

Issue #1157, opened 2025-07-22
Swarm - Best Diff. sorting incorrect


Travetown:
In the swarm overview, the “Best Diff.” column is sorted incorrectly under certain circumstances. 
In my case, the Bitaxe with 1.01 G is rated as the smallest in descending order. There is no conversion to a common base when sorting.

<img width="294" height="196" alt="Image" src="https://github.com/user-attachments/assets/526fd6b3-679a-4d02-9b30-a00d1b105ebc" />

duckaxe:
For devs: I would prefer that ESP return a number instead of a pre-formatted string. Formatting should happen in AxeOS. This will also work with sorting.

PR #1202, opened 2025-08-21
Change best(Session)Diff type from string to number

duckaxe:
To make it easier to work with the `best(Session)Diff` value in AxeOS, the `info` endpoint should output `bestDiff` and `bestSessionDiff` as numbers. AxeOS handles formatting (calc & sort is also much easier).

The swarm component has been adapted so that a number is used instead of a string. There are no visual changes.

This is a breaking change because the API endpoint now returns a number instead of a formatted string.

Fixes https://github.com/bitaxeorg/ESP-Miner/issues/1157

### DIFF
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
@@ -1,15 +1,17 @@
 import { HttpClient } from '@angular/common/http';
-import { Component, OnDestroy, OnInit, ViewChild } from '@angular/core';
+import { Component, OnDestroy, OnInit, ViewChild, HostListener } from '@angular/core';
 import { FormBuilder, FormGroup, Validators, FormControl } from '@angular/forms';
 import { ToastrService } from 'ngx-toastr';
-import { forkJoin, catchError, from, map, mergeMap, of, take, timeout, toArray, Observable } from 'rxjs';
+import { forkJoin, catchError, from, map, mergeMap, of, take, timeout, toArray, Observable, Subscription } from 'rxjs';
 import { LocalStorageService } from 'src/app/local-storage.service';
+import { LayoutService } from "../../layout/service/app.layout.service";
 import { ModalComponent } from '../modal/modal.component';
 import { ISystemInfo } from 'src/models/ISystemInfo';
 
 const SWARM_DATA = 'SWARM_DATA';
 const SWARM_REFRESH_TIME = 'SWARM_REFRESH_TIME';
 const SWARM_SORTING = 'SWARM_SORTING';
+const SWARM_GRID_VIEW = 'SWARM_GRID_VIEW';
 
 type SwarmDevice = { IP: string; ASICModel: string; deviceModel: string; swarmColor: string; asicCount: number; [key: string]: any };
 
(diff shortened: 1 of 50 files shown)


mutatrum:
NACK in current state, this will break swarm if people don't upgrade all their devices in one go. Swarm needs to be able to parse the strings when they query an older version.

0xf0xx0:
ive done this in #667 as well uwu

duckaxe:
Will add legacy code. Converted to draft.

duckaxe:
Added fallback code for devices that return best (session) diff as a string. Ready for review.

duckaxe:
@shufps maybe you can adopt this PR for Nerdaxe devices.

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -359,4 +341,41 @@ export class SwarmComponent implements OnInit, OnDestroy {
...
+
+  private parseSuffixString(input: string): number {
+    const regex = /^\s*([\d]+(?:\.\d+)?)\s*([kMGTPE]?)\s*$/i;
+    const match = input.match(regex);
+
+    if (!match) {
+      throw new Error(`Invalid suffix string: "${input}"`);
+    }
+
+    const value = parseFloat(match[1]);
```
mutatrum: No need to do regex, parseFloat parses as far as it can, so you can just chuck in the whole input string.
Then you can check on the last character of the string, to determine the multiplier.

main/http_server/axe-os/src/app/pipes/diff-suffix.pipe.ts
```diff
@@ -0,0 +1,33 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'diffSuffix',
+  pure: true
+})
+export class DiffSuffixPipe implements PipeTransform {
```
mutatrum, line 7: Take a look at 1217, there I merged all suffix pipes. Might makes sense to go into that direction here as well. The only difference between the different pipes is whatever comes after it.

duckaxe: 1217 or you mean https://github.com/bitaxeorg/ESP-Miner/pull/1236 ?

duckaxe: This PR is a month old and came before yours. Suggestion: Merging my PR as it is. Then we'll create an extra PR and clean up all the places where we use pipes, also https://github.com/bitaxeorg/ESP-Miner/pull/955 is affected.

mutatrum: Ok, makes sense.

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
duckaxe: Done.

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -140,12 +141,12 @@ export class SwarmComponent implements OnInit, OnDestroy {
   private getAllDeviceInfo(ips: string[], errorHandler: (error: any, ip: string) => Observable<SwarmDevice[] | null>, fetchAsic: boolean = true) {
     return from(ips).pipe(
       mergeMap(IP => forkJoin({
-        info: this.httpClient.get(`http://${IP}/api/system/info`),
+        info: this.httpClient.get<ISystemInfo>(`http://${IP}/api/system/info`),
```
mutatrum: Swarm should not use ISystemInfo, it's <any> by default. We query other versions, so no guarantee fields are there. This is loose typing by design.

mutatrum: Not sure if that breaks anything in runtime, btw.

duckaxe: You're right. Changed to `any`.

WantClue (approved):
ACK, please create a new issue for removing the fallback code as soon as @shufps moves over to number as well

duckaxe:
@WantClue Follow up ticket created.

PR #1203, opened 2025-08-21
Initial BitaxeNaja support

skot:
WIP


main/device_config.h
```diff
@@ -98,6 +104,7 @@ static const FamilyConfig FAMILY_HEX         = { .id = HEX,         .name = "Hex
 static const FamilyConfig FAMILY_SUPRA       = { .id = SUPRA,       .name = "Supra",      .asic = ASIC_BM1368, .asic_count = 1, .max_power = 40, .power_offset = 5,  .nominal_voltage = 5,  .swarm_color = "blue",   };
 static const FamilyConfig FAMILY_GAMMA       = { .id = GAMMA,       .name = "Gamma",      .asic = ASIC_BM1370, .asic_count = 1, .max_power = 40, .power_offset = 5,  .nominal_voltage = 5,  .swarm_color = "green",  };
 static const FamilyConfig FAMILY_GAMMA_TURBO = { .id = GAMMA_TURBO, .name = "GammaTurbo", .asic = ASIC_BM1370, .asic_count = 2, .max_power = 60, .power_offset = 10, .nominal_voltage = 12, .swarm_color = "cyan",   };
+static const FamilyConfig FAMILY_NAJA        = { .id = NAJA,        .name = "Naja",       .asic = ASIC_BM1340, .asic_count = 2, .max_power = 60, .power_offset = 10, .nominal_voltage = 12, .swarm_color = "magenta",};
```
mutatrum, line 107: NajaTurbo? Or will there not be a single chip version?

skot: in my mind "turbo" just means faster than the normal. In this case that would be a 3 or 4 chip version. If we did a single BM1340 version that could be the "lite"

PR #1208, opened 2025-08-27
feat: share icon for accepted and rejected

WantClue:
adds the symbol for share accepted and share rejected to the screen logic

closes #1123

PR #1209, opened 2025-08-27
Add shares accepted, rejected and work received notification to screen

### LINKED ISSUE #1123: OLED display icon for share submitted
The new dot on the OLED display when a share is submitted is awesome [#975]. I like the subtle feedback.

I think we should use a more descriptive icon to show this. Maybe a "up arrow"? And then we could show a "down arrow" when new work is received (mining.notify)?

mutatrum:
I've taken @WantClue's #1208 and integrated the arrow characters I already had in an earlier attempt to make a PR. Getting the notification to stick a little longer is tricky, as that would need separate timers for each case, making the code even more complicated than it already is.

Fixes #1123

### DIFF
--- a/main/bap/bap_subscription.c
+++ b/main/bap/bap_subscription.c
@@ -189,7 +189,7 @@ void BAP_send_subscription_update(GlobalState *state) {
                     case BAP_PARAM_SHARES:
                         {
                             char shares_ar_str[64];
-                            snprintf(shares_ar_str, sizeof(shares_ar_str), "%ld/%ld", state->SYSTEM_MODULE.shares_accepted, state->SYSTEM_MODULE.shares_rejected);
+                            snprintf(shares_ar_str, sizeof(shares_ar_str), "%lld/%lld", state->SYSTEM_MODULE.shares_accepted, state->SYSTEM_MODULE.shares_rejected);
                             BAP_send_message_with_queue(BAP_CMD_RES, "shares", shares_ar_str);
                             
                         }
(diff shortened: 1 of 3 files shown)


WantClue (approved):

mutatrum:
For future reference, how to draw a character:

In the current font, all characters are 6x8, so start with a binary grid of that size, with 1 for white and 0 for black. Example arrow:
```
001000
011100
111110
001000
001000
001000
001000
000000
```
Concatenate them all:
```
001000011100111110001000001000001000001000000000
```
Convert to hex with a [converter](https://www.rapidtables.com/convert/number/binary-to-hex.html?x=001000011100111110001000001000001000001000000000):
```
21CF88208200
```
If less than 12 characters, pad it with zeroes.

Split in bytes:
```
0x21, 0xCF, 0x88, 0x20, 0x82, 0x00,
```
and you have your new character.

main/bap/bap_subscription.c
```diff
@@ -189,7 +189,7 @@ void BAP_send_subscription_update(GlobalState *state) {
                     case BAP_PARAM_SHARES:
                         {
                             char shares_ar_str[64];
-                            snprintf(shares_ar_str, sizeof(shares_ar_str), "%lld/%lld", state->SYSTEM_MODULE.shares_accepted, state->SYSTEM_MODULE.shares_rejected);
+                            snprintf(shares_ar_str, sizeof(shares_ar_str), "%ld/%ld", state->SYSTEM_MODULE.shares_accepted, state->SYSTEM_MODULE.shares_rejected);
```
WantClue: it should not be needed to change this, would prefer to keep 8bytes

PR #1213, opened 2025-08-28
Block found screen sometimes doesn't show

mutatrum:


ghost:
This PR resolves the issues I was seeing, in fact it's working better than before.

PR #1223, opened 2025-09-07
Rearrange swarm layout

terratec:
- Move the swarm table up (is used regularly)
- Move the swarm controls down (are used only once or sometimes)


duckaxe:
Implemented in https://github.com/bitaxeorg/ESP-Miner/pull/1231 , thanks for the inspiration.

WantClue (approved):

PR #1229, opened 2025-09-11
Move dashboard messages from template to component

duckaxe:
This PR extracts all dashboard messages into the home component to improve performance and increase code reusability.


main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -360,4 +367,22 @@ export class HomeComponent implements OnInit, OnDestroy {
 
     return this.calculateAverage(efficiencies);
   }
+
+  public handleSystemMessages(info: ISystemInfo) {
+    const updateMessage = (condition: boolean, severity: Message['severity'], text: string) => {
```
mutatrum: I'm confused by the `severity: Message['severity']`. What does the Message struct reference doing here?

E.g. why isn't it `severity: string`?

Isn't this just a function with 3 arguments?

duckaxe: `Message['severity']` is the type for `severity` defined @ https://github.com/bitaxeorg/ESP-Miner/pull/1229/files#diff-d79dc922fc506333fd03dd11010371ffe33ec626d6c59ff57ffe27873475148eR14
Of course the type can be set to `string` but is stricter to set it to the know values such as `'error' | 'warn' | 'success' | 'info'`.

mutatrum: Ok, it's a construct I didn't know. Bit confusing :laughing:

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -468,6 +475,23 @@ export class HomeComponent implements OnInit, OnDestroy {
...
+      const exists = this.messages.some(msg => msg.text === text);
+      if (condition && !exists) {
+        this.messages.push({ severity, text });
+      } else if (!condition && exists) {
+        this.messages = this.messages.filter(msg => msg.text !== text);
+      }
+    };
+
+    updateMessage(info.overheat_mode === 1, 'error', 'Device has overheated - See settings');
+    updateMessage(!!info.power_fault, 'error', `${info.power_fault} Check your Power Supply.`);
```
mutatrum: This is tricky, the message is dynamic. So if `power_fault` changes, the old message is never removed. Same with the version below, if that changes, the old one is never removed.

Will it break if you clear all messages and just put in the active messages on each call?

duckaxe: > the old message is never removed

Yes, an already existing entry will be removed by this line:
https://github.com/bitaxeorg/ESP-Miner/blob/0d02b4fab8225652636d842512bb84da73209a2f/main/http_server/axe-os/src/app/components/home/home.component.ts#L484C1-L485C1

I tested this several times with the fallback pool message by switching the primary pool off and on. The message for the fallback pool appeared, then disappeared again as the primary pool became accessible again. Sure, you can empty the array. But then the messages will visually jump because they disappear briefly and then reappear.

mutatrum: If the substituted values change, the line is not removed. F.e. if the value of `info.power_fault` changes, it'll break.

More concretely, I have a PR in the works that reads the actual TPS546 power faults into that `power_fault` string, so it's more dynamic than it is now.

Alternatively, you could add an enum into the Message struct, with one value for each message type and test on that, instead of the string.

duckaxe: Added `Message Type` as message identifier.

Issue #1228, opened 2025-09-10
Add support for bitaxeUltra 207 external temp diode


skot:
All of the BM1366-based bitaxeUltra are setup to use the EMC2101 internal temperature sensor because the BM1366 doesn't have a die temp diode. 

In 207 I added a external diode Q1 - MMBT3904 connected to the EMC2101 placed near the ASIC. 

<img width="684" height="225" alt="Image" src="https://github.com/user-attachments/assets/7fd31c44-870b-4133-ace5-480951d695aa" />

To support this in esp-miner we need a 207-conditional part of the EMC2101 code to switch back to external temp diode (like is done for Max, Supra and Gamma). 

This will also need some temp calibrating (ideality and/or beta compensation). As-is it will still work just fine, but it will be using the EMC2101 internal temp sensor, which probably isn't as close to the ASIC as it was for previous bitaxeUltras

ghost:
Doing some tests here for this.

** **

<img width="1223" height="779" alt="Image" src="https://github.com/user-attachments/assets/1a11a9f1-795d-49be-b795-73c04943979c" />

** **

<img width="774" height="657" alt="Image" src="https://github.com/user-attachments/assets/fd101e0d-287e-408f-ab8b-8966c58e219a" />

skot:
the EMC2101 has been moved farther away from the ASIC in 207, so using the EMC2101-internal temp sensor will probably read a little cooler. 

<img width="706" height="657" alt="Image" src="https://github.com/user-attachments/assets/feac6a96-d79b-4e59-8460-d2c8f8dfe85b" />

PR #1230, opened 2025-09-11
207 external temp

STSMiner1:
Tests here show this working for the Ultra 207.

https://github.com/bitaxeorg/ESP-Miner/issues/1228

** **
image

### DIFF
--- a/components/stratum/include/utils.h
+++ b/components/stratum/include/utils.h
@@ -34,6 +34,10 @@ void prettyHex(unsigned char *buf, int len);
 
 uint32_t flip32(uint32_t val);
 
+double networkDifficulty(uint32_t nBits);
+
+void suffixString(uint64_t val, char * buf, size_t bufsiz, int sigdigits);
+
 #define STRATUM_DEFAULT_VERSION_MASK 0x1fffe000
 
 #endif // STRATUM_UTILS_H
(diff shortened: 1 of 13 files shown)


skot:
I don't have HW to test, but the changes LGTM

mutatrum:
Setting `emc_internal_temp` to `false` shouldn't be needed, that is the default value.

ghost:
> Setting `emc_internal_temp` to `false` shouldn't be needed, that is the default value.

Without that I was not getting the reading with the decimal point value and it was also the same value as the VR at times.

I explicitly set `.emc_internal_temp = false` to guarantee the external diode is used.

** ** 

image

** ** 

The fan is also more active when the ASIC temp changes.

mutatrum:
> > Setting `emc_internal_temp` to `false` shouldn't be needed, that is the default value.
> 
> Without that I was not getting the reading with the decimal point value and it was also the same value as the VR at times.

Can you double-check this? As if that would be true, none of the Gammas should work. When I added the `DeviceConfig` struct, I double-checked how it's initialised and what the default values are. For `bool` this is false.

Not saying there's no issue, but this is a no operation code change, therefore the cause must be something else.

WantClue:
@STSMiner1 is this still being worked on ?

ghost:
> @STSMiner1 is this still being worked on ?

It's not required for the Ultra 207, I checked with @mutatrum with this when we was testing things, this PR was set to draft in case anything changes at a later date.

Closing this PR.

PR #1232, opened 2025-09-14
Fix color representation for Nerdaxe devices

duckaxe:
Nerdaxe devices provide a hex color like `#e7cf00` as `swarmColor`. However, the current code cannot handle this because it expects a string like `orange` to build a corresponding CSS class `[ngClass]="'text-' + axe.swarmColor + '-500'"`. 

This PR solves the problem by outputting the color directly, regardless of whether it is a string or HEX value, instead of using the CSS class.

**Before**
Before

**After**
After

### DIFF
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.html
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.html
@@ -121,7 +121,7 @@
             <tr>
                 <td>
                     <a
-                       [ngClass]="'text-'+axe.swarmColor+'-500'"
+                       [ngStyle]="{color: axe.swarmColor}"
                        [href]="'http://'+axe.IP"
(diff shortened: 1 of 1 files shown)


0xf0xx0 (approved):

0xf0xx0:
interesting how the ultra changes colors x3

mutatrum:
Pinging @shufps to join here. It would be good if we are in line for these cross-implementation things.

duckaxe:
> interesting how the ultra changes colors x3

Because `text-purple-500` != `color: purple`

WantClue (approved):

WantClue:
@duckaxe please resolve this conflict so we can merge

duckaxe:
@WantClue done

PR #1234, opened 2025-09-15
*NO NOT MERGE* Set PSRAM type to Quad

### LINKED ISSUE #826: Make firmware releases for other PSRAM variants
Unfortunately, the ESP32-S3 cannot auto-detect if the PSRAM on board is Quad or Octal. There have been 3 instances of people installing ESP's with different PSRAM settings on board, and they end up with a boot loop. They all succeeded in creating their own firmware, but it might be a small effort to have two subfolders in the build directory where images for these variants can be prepared. It would only have to be Quad and maybe for models without PSRAM. 

AFAIK there are no other ESP types that causes boot issues. There are also models with smaller flash size, but AFAIK these only trigger a warning in the logs, they don't cause a boot loop.

mutatrum:
This is a ad-hoc way of generating a `esp-miner.bin` file for devices which have Quad PSRAM. 

See #826


mutatrum:
It works, with plenty of free heap left:

image

WantClue:
I think we abandoned other psram variants ...

mutatrum:
> I think we abandoned other psram variants ...

We currently detect if there's no PSRAM on board, so we have minimal support. However, that won't help anymore as it just crashes shortly after as it runs out of memory before it finished booting. A possible low-effort solution is #1239, which disables non-essential things, so at least the firmware boots, but that's fairly tricky as we would always need to test if that's still functional. If it does work, it helps in troubleshooting, as these out of memory crashes are difficult to analyse. It needs either a detailed picture of the ESP to determine the model (which is hard for some users) or capture of boot logs before the crash (even harder). Alternatively, we could stop the boot if no PSRAM is detected and show a proper error message on the display, instead of just crashing into a boot-loop.

In the case someone has a Quad PSRAM (2MB versions), it's still possible to run the full firmware, as we've only just started to use the PSRAM. This PR was (is) a placeholder just for generating those binaries. This helped a few people already, which is a win. Of course there's #826, but that's more involved.

IMO a combination would be nice: proper error message if no PSRAM is there, and a Quad PSRAM support firmware. Or hopefully https://github.com/espressif/esp-idf/issues/13343 gets solved at some point.

mutatrum:
Same as #1239, no 2MB PSRAM devices have been spotted in quite a while.

PR #1235, opened 2025-09-15
Added blockFound to the info API endpoint (and dashboard)

duckaxe:
This PR adds `blockFound` to the `info` API endpoint. At the same time, a condition has been added to the dashboard that triggers a notification as soon as a block is found. Feel free to suggest a nicer message text.

```
{
    ...
	"statsFrequency":	0,
	"blockFound":	0,
	"blockHeight":	914841,
	"networkDifficulty":	"136.04 T"
}
```

**Desktop Notification**
Screenshot 2025-09-15 at 20 07 17

### DIFF
--- a/main/http_server/http_server.c
+++ b/main/http_server/http_server.c
@@ -720,6 +720,8 @@ static esp_err_t GET_system_info(httpd_req_t * req)
 
     cJSON_AddNumberToObject(root, "statsFrequency", nvs_config_get_u16(NVS_CONFIG_STATISTICS_FREQUENCY, 0));
 
+    cJSON_AddNumberToObject(root, "blockFound", GLOBAL_STATE->SYSTEM_MODULE.FOUND_BLOCK);
+
     if (GLOBAL_STATE->SYSTEM_MODULE.power_fault > 0) {
         cJSON_AddStringToObject(root, "power_fault", VCORE_get_fault_string(GLOBAL_STATE));
     }
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -22,6 +22,10 @@
         text="Firmware ({{info.version}}) and AxeOS ({{info.axeOSVersion}}) versions do not match. Please make sure to update both www.bin and esp-miner.bin.">
     </p-message>
 
+    <p-message *ngIf="info.blockFound" severity="success" styleClass="w-full mb-4 py-4 border-round-xl"
+        text="Today is your lucky day! Your device has found a block 🎉">
+    </p-message>
+
     <div class="grid">
         <div class="col-12 md:col-6 xl:col-3">
             <div class="card mb-0">
--- a/main/http_server/axe-os/src/app/services/system.service.ts
+++ b/main/http_server/axe-os/src/app/services/system.service.ts
@@ -88,6 +88,8 @@ export class SystemService {
         blockHeight: 811111,
         scriptsig: "..%..h..,H...ckpool.eu/solo.ckpool.org/",
         networkDifficulty: "25.3T",
+
+        blockFound: 0,
       }
     ).pipe(delay(1000));
   }
--- a/main/http_server/axe-os/src/models/ISystemInfo.ts
+++ b/main/http_server/axe-os/src/models/ISystemInfo.ts
@@ -69,4 +69,6 @@ export interface ISystemInfo {
     blockHeight?: number,
     scriptsig?: string,
     networkDifficulty?: string,
+
+    blockFound: number,
 }
--- a/main/http_server/openapi.yaml
+++ b/main/http_server/openapi.yaml
@@ -148,6 +148,7 @@ components:
         - blockHeight
         - scriptsig
         - networkDifficulty
+        - blockFound
       properties:
         ASICModel:
           type: string
@@ -331,6 +332,9 @@ components:
         networkDifficulty:
           type: string
           description: Current network difficulty in human-readable format
+        blockFound:
+          type: number
+          description: Whether a block was found (0=no, 1=yes)
 
     Settings:
       type: object


mutatrum:
Love it, but that notification should be WAAAAY bigger :laughing: 

And it would be cool to add block height, solved difficulty and network difficulty. Although we don't persist these.

skot:
this is so cool! I might suggest; "Today is your lucky day! Your device found block 914841  🎉"

skot:
maybe a surprise theme change too 😄

mutatrum:
https://confettipage.com/

skot:
> https://confettipage.com/

my mind is blown.. this is CaaS; https://confettipage.com/en/pricing

mutatrum:
> > https://confettipage.com/
> 
> my mind is blown.. this is CaaS; https://confettipage.com/en/pricing

Oh wow, didn't expect that. There's free js versions that do that, didn't expect the audacity to wrap that into a monetization scheme.

main/http_server/http_server.c
```diff
@@ -720,6 +720,8 @@ static esp_err_t GET_system_info(httpd_req_t * req)
 
     cJSON_AddNumberToObject(root, "statsFrequency", nvs_config_get_u16(NVS_CONFIG_STATISTICS_FREQUENCY, 0));
 
+    cJSON_AddNumberToObject(root, "blockFound", GLOBAL_STATE->SYSTEM_MODULE.FOUND_BLOCK);
```
mutatrum: blockFound or foundBlock? It's already a bit muddy in the code, maybe align them?

duckaxe: I would keep `blockFound` since we already have a key called `blockHeight`. I would rename FOUND_BLOCK to BLOCK_FOUND. Agreed?

mutatrum: Maybe lower case `block_found`, it's the only one shouting?

duckaxe:
> And it would be cool to add block height, solved difficulty and network difficulty. Although we don't persist these.

> this is so cool! I might suggest; "Today is your lucky day! Your device found block 914841  🎉"

That is why I did not include block information in the message. Not every user is constantly looking at the screen. Consider the following scenario: AxeOS is open in a browser window, and a block is found. An hour later, the user looks at the screen and is happy to see it. However, the block information is no longer correct because it has been updated several times.

ghost:
Lets see if you account for this...

"Device finds a block with this difficulty, user does not restart device, two days later device finds another block but this time with a lower difficulty value (yes, I know 1 in a billion chance ... but)"

Right now ... block found does not inform the user that it's found another block with a lower difficulty value.

main/http_server/http_server.c
duckaxe: Yes, only this. Renamed.

duckaxe:
* Added notification to the browser tab title
* Added Confetti effect to dashboard (no third-party scripts)

Video attached to PR description.

skot:
> That is why I did not include block information in the message. Not every user is constantly looking at the screen. Consider the following scenario: AxeOS is open in a browser window, and a block is found. An hour later, the user looks at the screen and is happy to see it. However, the block information is no longer correct because it has been updated several times.

perhaps the esp-miner API could provide the block height of the found block(s)? It could be as simple as;
```
{
    ...
	"statsFrequency":	0,
	"blockFound":	[914840, 914841, 914842],
	"blockHeight":	914845,
	"networkDifficulty":	"136.04 T"
}
```

mutatrum:
Awesome, confetti embedded with 100 lines of code!

Storing all found blocks will cause issues with the shitcoin miners, they find infinite worthless blocks.

The `found_block` is already an integer, it's trivial to use that to store the block. Although it needs to go through bm_job, as the current block height is from the last incoming `mining.notify`, which could be higher than the winning share submit.

skot:
> Awesome, confetti embedded with 100 lines of code!
> 
> Storing all found blocks will cause issues with the shitcoin miners, they find infinite worthless blocks.
> 
> The `found_block` is already an integer, it's trivial to use that to store the block. Although it needs to go through bm_job, as the current block height is from the last incoming `mining.notify`, which could be higher than the winning share submit.

Those damn shitcoin miners 😆 So maybe the esp-miner API just reports the block height of the most recent block found? This is all just in esp-miner RAM for now, right?

duckaxe:
[BlockFound.patch](https://github.com/user-attachments/files/22380523/BlockFound.patch)

How about storing the found block in the `block_found` formerly `FOUND_BLOCK` variable? Instead of using 0/1, assign the block height. It's an easy solution for storing the founded block height.

mutatrum:
> [BlockFound.patch](https://github.com/user-attachments/files/22380523/BlockFound.patch)
> 
> How about storing the found block in the `block_found` formerly `FOUND_BLOCK` variable? Instead of using 0/1, assign the block height. It's an easy solution for storing the founded block height.

See https://github.com/bitaxeorg/ESP-Miner/pull/1235#issuecomment-3300286499

It needs a bit more to get it accurate. Last thing we would want is to have a screenshot of a nice banner with confetti of a Bitaxe finding a block and it shows wrong block height.

duckaxe:
> See [#1235 (comment)](https://github.com/bitaxeorg/ESP-Miner/pull/1235#issuecomment-3300286499)

Oh, missed your comment. Alright.

duckaxe:
I would merge this pull request as it is. Once we have a good solution for storing block heights, we will create a new PR.

main/http_server/http_server.c
```diff
@@ -767,6 +767,8 @@ static esp_err_t GET_system_info(httpd_req_t * req)
 
     cJSON_AddNumberToObject(root, "statsFrequency", nvs_config_get_u16(NVS_CONFIG_STATISTICS_FREQUENCY, 0));
 
+    cJSON_AddNumberToObject(root, "blockFound", GLOBAL_STATE->SYSTEM_MODULE.block_found);
```
0xf0xx0, line 770: non-blocking nit: `AddBoolToObject`?

PR #1231, opened 2025-09-14
Swarm Facelift

duckaxe:
This PR introduces an enhanced Swarm View. It is more compact, clear, and mobile friendly. Based on the design of the other views.

Inspired by https://github.com/bitaxeorg/ESP-Miner/pull/1223, the swarm table moves up and the button bar moves down.

All functionalities remain unchanged. The changes only affect the template and stylings.

**Desktop**
Desktop

**Mobile**
Mobile

PR #1236, opened 2025-09-16
Add scoreboard :trophy: 

mutatrum:
This PR adds a scoreboard screen that shows the top 20 shares:

image

And an `api/system/scoreboard` api call:

```
[
  {
    "difficulty": 956725.8468403114,
    "job_id": "684934fc00047bfd",
    "extranonce2": "0600000000000000",
    "ntime": 1758023248,
    "nonce": "F298047E",
    "version_bits": "040BC000"
  },
  {
    "difficulty": 942190.512488009,
    "job_id": "684934fc00047bc7",
    "extranonce2": "0e00000000000000",
    "ntime": 1758021685,
    "nonce": "5B7D04AA",
    "version_bits": "0122A000"
  },
  {
    "difficulty": 699177.2079349027,
    "job_id": "684934fc00047b90",
    "extranonce2": "1f00000000000000",
    "ntime": 1758020167,
    "nonce": "8B1F05EA",
    "version_bits": "04CA2000"
  },
```

Currently it captures all the fields from `mining.submit`, but others can be added. Not sure if adding the full header and resulting hash should be added.

### DIFF
--- a/main/http_server/openapi.yaml
+++ b/main/http_server/openapi.yaml
@@ -580,6 +580,41 @@ components:
             items:
               type: number
 
+    SystemScoreboardEntry:
+      type: object
+      required:
+        - difficulty
+        - job_id
+        - extranonce2
+        - ntime
+        - nonce
+        - version_bits
+      properties:
+        rank:
+          type: number
+          description: Rank of the share in the scoreboard
+        since:
+          type: number
+          description: Time since share found in seconds
+        difficulty:
+          type: number
+          description: Difficulty of the share
+        job_id:
+          type: string
+          description: Job ID of the share
+        extranonce2:
+          type: string
+          description: Extranonce2 of the share
+        ntime:
+          type: number
+          description: Ntime of the share
+        nonce:
+          type: string
+          description: Nonce of the share
+        version_bits:
+          type: string
+          description: Version bits of the share
+
     Settings:
       type: object
       properties:
@@ -843,34 +878,7 @@ paths:
               schema:
                 type: array
                 items:
-                  title: SystemScoreboardEntry
-                  type: object
-                  required:
-                    - difficulty
-                    - job_id
-                    - extranonce2
-                    - ntime
-                    - nonce
-                    - version_bits
-                  properties:
-                    difficulty:
-                      type: number
-                      description: Difficulty of the share
-                    job_id:
-                      type: string
-                      description: Job ID of the share
-                    extranonce2:
-                      type: string
-                      description: Extranonce2 of the share
-                    ntime:
-                      type: number
-                      description: Ntime of the share
-                    nonce:
-                      type: string
-                      description: Nonce of the share
-                    version_bits:
-                      type: string
-                      description: Version bits of the share
+                  $ref: '#/components/schemas/SystemScoreboardEntry'
 
   /api/system/pause:
     post:
--- a/main/http_server/axe-os/src/app/components/network-edit/network.edit.component.ts
+++ b/main/http_server/axe-os/src/app/components/network-edit/network.edit.component.ts
@@ -7,7 +7,7 @@ import { BehaviorSubject, Observable } from 'rxjs';
 import { DialogService } from 'src/app/services/dialog.service';
 import { LoadingService } from 'src/app/services/loading.service';
 import { SystemApiService } from 'src/app/services/system.service';
-import { WifiNetwork } from 'src/app/generated';
+import { WifiNetwork } from 'src/app/generated/models';
 
 @Component({
   selector: 'app-network-edit',


mutatrum:
Thanks to @duckaxe for combing over my NG stuff :handshake:

skot:
> Next step would be to show it somewhere on AxeOS and to persist this table in NVS. Adding the current all time best to the table would be a nice-to-have, even though we don't have the other fields.

As with all flash we need to be careful to not erase from NVS at high frequency as this can wear out the flash. I think keeping the high score list in NVS is a great idea. Just need to be aware that we only get something like 100,000 erase cycles on a ESP32 flash sector before it will wear out and be unreliable. An erase every minute would be bad (~ 2 months MTBF), an erase every hour (~ 11 years MTBF) would probably be fine. The esp-idf NVS library does some wear leveling afaik, but I wouldn't want to rely on that.

generally speaking, just _changing_ a value in flash involves a erase cycle on the whole sector.

mutatrum:
> > Next step would be to show it somewhere on AxeOS and to persist this table in NVS. Adding the current all time best to the table would be a nice-to-have, even though we don't have the other fields.
> 
> As with all flash we need to be careful to not erase from NVS at high frequency as this can wear out the flash. I think keeping the high score list in NVS is a great idea. Just need to be aware that we only get something like 100,000 erase cycles on a ESP32 flash sector before it will wear out and be unreliable. An erase every minute would be bad (~ 2 months MTBF), an erase every hour (~ 11 years MTBF) would probably be fine. The esp-idf NVS library does some wear leveling afaik, but I wouldn't want to rely on that.
> 
> generally speaking, just _changing_ a value in flash involves a erase cycle on the whole sector.

Apologies for the long reply, it was a deeper rabbit hole than I expected.

I did some reading, and initially bad news: the current NVS partition is only 24kb, which means 6 pages of 4kb. However, the write pattern is more subtle, and that's good news: each page is an append log, and writes are append onto the same page without an erase. Updates are also appended, and the old value is marked as deleted. Only when a page is full, it starts to use a next page. If all but the last empty page is used, a full page is recycled, where still valid key/values are moved to a new page and it's actually flash erased. [link](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/storage/nvs_flash.html#pages-and-entries)

We are currently at ~33% usage of NVS, Added some statistics:
```
I (1104) nvs_device: Used entries: 249
I (1104) nvs_device: Free entries: 507
I (1104) nvs_device: Available entries: 381
I (1107) nvs_device: Total entries: 756
```
The difference between free and available is old values that are marked as deleted, and will at some time be garbage collected page. Each entry is 32 bytes: small values (8, 16 or 32 bit) are a single entry, blobs and strings can be more than one entry. Blobs have 2 entries overhead and string have 1 entry overhead.

I also looked at repartitioning, but that's a tricky avenue. It has been done, but makes f.e. downgrades very challenging. This PR doesn't warrant that.

So, not sure here. It's possible to add 20 NVS string entries with all values concatenated per line. This would be around 80 NVS entries. IMO this is a reasonable trade-off for write granularity. Having the scoreboard persistent is making it a lot more fun. More ways of celebrating high difficulties and to have an idea when they happened and how rare a high difficulty is gives a lot of insight in the process.

mutatrum:
Besides the NVS storage aspect, are there other fields that should be added or replaced?

skot:
This does feel a little bit outside of the practical use of the ESP32's limited NVS. I could image the high scoreboard being a good thing to punt to the client dashboard to keep track of.. Or maybe just ESP32 RAM?

mutatrum:
lol I added it, as I wanted to test if it would work. With 20 entries as strings, I get the following NVS statistics:
```
I (1116) nvs_device: Used entries: 329
I (1116) nvs_device: Free entries: 427
I (1116) nvs_device: Available entries: 301
I (1119) nvs_device: Total entries: 756
```
Exactly 80 entries are added, the math works out. It uses about 10% of NVS storage, it's now at 43%.

The first few minutes have quite a few updates, but as the difficulty rises quickly, the number of updates will drop really fast. A minimum threshold before saving an entry would reduce that.

> This does feel a little bit outside of the practical use of the ESP32's limited NVS. I could image the high scoreboard being a good thing to punt to the client dashboard to keep track of.. Or maybe just ESP32 RAM?

I agree, it's pushing it a bit. But as I mentioned above, at the same time it's also a really informative addition, IMO. And having the scoreboard survive a reboot is chefs kiss :laughing:

To not paint ourselves into the corner, I'll experiment a bit with adding a second NVS partition at the end of the partition table. I can't find a conclusive answer from the documentation if this is allowed/possible or not.

main/http_server/http_server.c
```diff
@@ -1119,6 +1170,14 @@ esp_err_t start_rest_server(void * pvParameters)
     };
     httpd_register_uri_handler(server, &system_statistics_dashboard_get_uri);
 
+    httpd_uri_t scoreboard_get_uri = {
+        .uri = "/api/system/scoreboard",
```
duckaxe, line 1438: Add the new endpoint to the Readme API section.

duckaxe: ... and to `openapi.yml`

mutatrum: Added, thx!

duckaxe:
Some improvements https://github.com/mutatrum/ESP-Miner/pull/4

main/http_server/axe-os/src/app/components/scoreboard/scoreboard.component.html
```diff
@@ -6,25 +6,62 @@ <h2>Scoreboard</h2>
...
+                        {
+                            name: 'nonce',
+                            label: 'Nonce'
+                        },
+                        {
+                            name: 'version_bits',
+                            label: 'Version Bits'
+                        }
+                    ]">
+                        <div class="flex align-items-center cursor-pointer select-none h-full" (click)="sortBy(field.name)">
```
duckaxe: It would be better and more consistent if you based your change on the new swarm design, not the old one. https://github.com/bitaxeorg/ESP-Miner/pull/1231

duckaxe: This applies not only to this line :)

mutatrum: Done, however, the table  columns jump around when switching sort column. I tried several things (invisible icon and span placeholder), but can't seem to fix it.

main/http_server/axe-os/src/app/components/scoreboard/scoreboard.component.html
```diff
@@ -0,0 +1,49 @@
...
+                                    'pi-sort-down-fill': sortField === field.name && sortDirection === 'desc'
+                                }"></i>
+                            </div>
+                        </th>
+                        <th></th>
+                    </tr>
+                </thead>
+                <tbody>
+                    <tr *ngFor="let entry of scoreboard">
+                        <td text-center>
```
duckaxe: ~~`text-center`~~
If you like to center column content: `<td class="text-center">`

mutatrum: Removed it, was ugly as the header was not centered. This is fiddly territory so it's good enough as it is without center.

main/http_server/axe-os/src/app/components/scoreboard/scoreboard.component.html
duckaxe: In case you missed my pr https://github.com/mutatrum/ESP-Miner/pull/5

mutatrum: I indeed missed that, thanks!

duckaxe:
A little fix for smaller screens https://github.com/mutatrum/ESP-Miner/pull/6

Travetown:
Wouldn't it be good if the highest difficulty level to date could be adopted? Of course, not all the information is available anymore. Perhaps it could be replaced with ‘*’ or something else. But then the list would be complete.  At least in first place.

main/http_server/axe-os/src/app/components/scoreboard/scoreboard.component.html
```diff
@@ -0,0 +1,48 @@
...
+                        <td>
+                            <i class="pi pi-trophy" *ngIf="entry.rank < 3" [ngStyle]="{
+                                'color': entry.rank === 0 ? 'gold' : entry.rank === 1 ? 'silver' : 'peru'
+                            }"></i>
+                            <span *ngIf="entry.rank >= 3" class="text-500 text-xs">{{ entry.rank + 1 }}</span>
+                        </td>
+                        <td pTooltip="{{ entry.difficulty }}" tooltipPosition="top" class="text-900">{{ entry.difficulty | suffix }}</td>
+                        <td class="code">{{ entry.job_id }}</td>
+                        <td class="code">{{ entry.extranonce2 }}</td>
+                        <td pTooltip="{{ entry.since | dateAgo: {intervals: 2} }}" tooltipPosition="top">{{ entry.ntime * 1000 | date:'yyyy-MM-dd HH:mm:ss' }}</td>
```
0xf0xx0: nit: add " ago" to the tooltip

mutatrum:
> Wouldn't it be good if the highest difficulty level to date could be adopted? Of course, not all the information is available anymore. Perhaps it could be replaced with ‘*’ or something else. But then the list would be complete. At least in first place.

I though about that, but think that's more grief than necessary. I'm running my Gamma now for ~ 9 months, so to get a new higher best, it would take on average another 9 months. (I didn't calculate, but it has a 5G best difficulty, no idea what the odds are for that.) To get that score flushed out of the top 20, it would need 20 higher scores, which would probably take longer than the lifetime of the device. Then for the rest of the lifetime of the device, there will be a single line item that can't be sorted and shows empty fields.

Travetown:
> I though about that, but think that's more grief than necessary. I'm running my Gamma now for ~ 9 months, so to get a new higher best, it would take on average another 9 months. (I didn't calculate, but it has a 5G best difficulty, no idea what the odds are for that.) To get that score flushed out of the top 20, it would need 20 higher scores, which would probably take longer than the lifetime of the device. Then for the rest of the lifetime of the device, there will be a single line item that can't be sorted and shows empty fields.

I agree with you. However, as is easy to see, people are very focused on the interest/pursuit of high difficulty. 
It will be interesting to see how the absence of the current highest difficulty level is received. I suspect there will be many questions/complaints. 
Since most people are probably not very interested in the data displayed (except for difficulty /date), or are unable to interpret it, it would not be a great loss not to display it.

seepv:
A Summary would be helpful:

```
The scoreboard difficulty distribution is:
0 T-Diff | 3 G-Diff and 17 M-Diff 
```

Travetown:
A top 20 of all worker (swarm) would also be nice.

main/http_server/axe-os/src/app/components/swarm/swarm.component.html
```diff
@@ -6,7 +6,7 @@ <h2>Swarm</h2>
         <div class="grid">
             <div class="col-6 xl:col-3" *ngFor="let metric of [
                 { label: 'Total Devices', value: swarm.length },
-                { label: 'Total Hashrate', value: totals.hashRate * 1000000000 | hashSuffix },
+                { label: 'Total Hashrate', value: totals.hashRate * 1000000000 | hashSuffix},
```
duckaxe: If you remove the space, do so at the beginning and end of all entries.

mutatrum: It was a revert commit to get `swarm.component.html` out of this PR. I missed a newline, so now it's no longer in this PR.

Whitespace fixes will come on a PR that does something on the file.

mutatrum:
> tested, my main concern is def the nvs though could the scoreboard be updated in ram and written periodically?

After a week or two there's maybe one update a day or so, it's good for decades.

Travetown:
After a little over a month with a total of 5 Bitaxe Gamma with scoreboard:
The lowest difficulty is now between 50 and 75M, and new entries have become correspondingly rare.
There have been no problems with the scoreboard and I think it's great.

duckaxe (approved):

Travetown:
Another quick update.
After about 10 weeks, # 20 is now around 150M for most Bitaxe. This means that on average, a new share can be expected in 5 days (at 1.5 Th/s).
Unfortunately, this PR seems to have fallen out of focus somewhat.
Perhaps it will be tested again with the current master.  😇

mutatrum:
> Another quick update. After about 10 weeks, # 20 is now around 150M for most Bitaxe. This means that on average, a new share can be expected in 5 days (at 1.5 Th/s). Unfortunately, this PR seems to have fallen out of focus somewhat. Perhaps it will be tested again with the current master. 😇

It won't be forgotten, it just takes time to mature :wink:

main/tasks/scoreboard.c
```diff
@@ -0,0 +1,91 @@
...
+void scoreboard_init(Scoreboard *scoreboard) {
+    scoreboard->count = 0;
+    scoreboard->mutex = xSemaphoreCreateMutex();
+    if (scoreboard->mutex == NULL) {
+        ESP_LOGE(TAG, "Failed to create mutex");
+    }
+
+    for (int i = 0; i < MAX_SCOREBOARD; i++) {
+        char *entry_str = nvs_config_get_string_indexed(NVS_CONFIG_SCOREBOARD, i);
+        if (entry_str[0] == '\0') {
```
WantClue: If nvs_config_get_string_indexed returns NULL (e.g., on allocation failure), dereferencing entry_str[0] on line 17 will crash. Add a NULL check before accessing the string.

if (entry_str == NULL) {
            ESP_LOGW(TAG, "Failed to get scoreboard entry %d from NVS", i);
           break;
       }

main/tasks/scoreboard.c
```diff
@@ -0,0 +1,91 @@
...
+#include "esp_log.h"
+#include <stdio.h>
+
+static const char * TAG = "scoreboard";
+
+void scoreboard_init(Scoreboard *scoreboard) {
+    scoreboard->count = 0;
+    scoreboard->mutex = xSemaphoreCreateMutex();
+    if (scoreboard->mutex == NULL) {
+        ESP_LOGE(TAG, "Failed to create mutex");
```
WantClue, line 13: you're missing a mutex creation exit on failure, add "return; "

main/nvs_config.c
```diff
@@ -297,38 +346,57 @@ esp_err_t nvs_config_init(void)
...
     // Skip if invalid, wrong type, or value unchanged
-    if (!setting || setting->type != TYPE_STR || (setting->value.str && strcmp(setting->value.str, value) == 0)) return;
+    if (!setting || setting->type != TYPE_STR || (setting->value[0].str && strcmp(setting->value[0].str, value) == 0)) return;
 
     ConfigUpdate update = { .key = key, .type = TYPE_STR, .value.str = strdup(value) };
     if (!update.value.str) return;
     xQueueSend(nvs_save_queue, &update, portMAX_DELAY);
 }
 
+void nvs_config_set_string_indexed(NvsConfigKey key, int index, const char *value)
```
WantClue, line 390: The function accesses setting->value[index] without validating that index is within [0, setting->array_size).

-    if (!setting || setting->type != TYPE_STR || (setting->value[index].str && strcmp(setting->value[index].str, value) == 0)) return;
+    if (!setting || setting->type != TYPE_STR || index < 0 || index >= get_array_size(setting)) return;
+    if (setting->value[index].str && strcmp(setting->value[index].str, value) == 0) return;

main/nvs_config.c
```diff
@@ -297,38 +346,57 @@ esp_err_t nvs_config_init(void)
 char *nvs_config_get_string(NvsConfigKey key)
 {
     Settings *setting = nvs_config_get_settings(key);
-    if (!setting || setting->type != TYPE_STR) {
+    if (!setting || setting->type != TYPE_STR || setting->array_size > 1) {
```
WantClue: If setting is NULL (invalid key), the condition !setting is true, but the code still accesses setting->nvs_key_name in the error log, causing undefined behavior.

remove the setting->nvs_key_name and replace it with the "key" which is always available

main/http_server/http_server.c
```diff
@@ -1075,6 +1075,57 @@ static esp_err_t GET_system_statistics(httpd_req_t * req)
...
+            cJSON_AddStringToObject(entry, "nonce", nonce_str);
+            cJSON_AddStringToObject(entry, "version_bits", version_bits_str);
+
+            cJSON_AddItemToArray(root, entry);
+        }
+        xSemaphoreGive(scoreboard.mutex);
+    } else {
+        ESP_LOGE(TAG, "Failed to take mutex for JSON conversion");
+        cJSON_Delete(root);
+        return ESP_FAIL;
```
WantClue: Returning ESP_FAIL without sending an HTTP response leaves the client hanging. Other handlers in this file send an error response before returning.

return gracefully with an ESP_OK and send error like 500

main/http_server/http_server.c
```diff
@@ -1075,6 +1075,57 @@ static esp_err_t GET_system_statistics(httpd_req_t * req)
...
+
+    httpd_resp_set_type(req, "application/json");
+
+    // Set CORS headers
+    if (set_cors_headers(req) != ESP_OK) {
+        httpd_resp_send_500(req);
+        return ESP_OK;
+    }
+
+    Scoreboard scoreboard = GLOBAL_STATE->SYSTEM_MODULE.scoreboard;
```
WantClue: Copying the Scoreboard struct by value happens outside the mutex protection

main/tasks/scoreboard.h
```diff
@@ -0,0 +1,33 @@
...
+} ScoreboardEntry;
+
+typedef struct {
+    ScoreboardEntry entries[MAX_SCOREBOARD];
+    int count;
+    SemaphoreHandle_t mutex;
+} Scoreboard;
+
+void scoreboard_init(Scoreboard *scoreboard);
+void scoreboard_add(Scoreboard *scoreboard, double difficulty, const char *job_id, const char *extranonce2, uint32_t ntime, uint32_t nonce, uint32_t version_bits);
```
WantClue: both are void you should use esp_err_t to indicate any errors or failures such as mutex creation failure

main/http_server/axe-os/src/app/components/scoreboard/scoreboard.component.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { ScoreboardComponent } from './scoreboard.component';
+
+describe('SystemComponent', () => {
```
WantClue: uhm this should be ScoreboardComponent

main/http_server/axe-os/src/app/components/scoreboard/scoreboard.component.ts
```diff
@@ -0,0 +1,89 @@
...
+    });
+    this.refresh$.next();
+  }
+
+  private sortData(data: ISystemScoreboardEntry[], field: string, direction: 'asc' | 'desc'): ISystemScoreboardEntry[] {
+    if (!data || !field) {
+      return data;
+    }
+
+    return data.sort((a, b) => {
```
WantClue: we might want to use return [...data].sort((a, b) => { otherwise we violate the immutability principles.

main/http_server/axe-os/src/app/pipes/date-ago.pipe.ts
```diff
@@ -35,12 +35,11 @@ export class DateAgoPipe implements PipeTransform {
           if (result) result += args?.short ? ' ' : ', ';
           result += counter + (args?.short ? i[0] : ' ' + i + (counter > 1 ? 's' : ''));
           seconds -= intervals[i] * counter
-          shownIntervals++;
         }
+        if (result) shownIntervals++;
```
WantClue, line 39: Moving the increment outside the if (counter > 0) block causes shownIntervals to increment on every loop iteration once result becomes truthy, even when that iteration adds nothing (counter = 0)

main/http_server/axe-os/src/app/services/system.service.ts
```diff
@@ -195,6 +196,35 @@ export class SystemService {
     });
   }
 
+  public getScoreboard(uri: string = ''): Observable<ISystemScoreboardEntry[]> {
+    if (environment.production) {
+      return this.httpClient.get(`${uri}/api/system/scoreboard`) as Observable<ISystemScoreboardEntry[]>;
```
WantClue: you're missing the timeout pipe, it's on all other GET methods add a timeout 5000

main/http_server/openapi.yaml
```diff
@@ -768,6 +768,53 @@ paths:
...
+                items:
+                  type: object
+                    required:
+                      - difficulty
+                      - job_id
+                      - extranonce2
+                      - ntime
+                      - nonce
+                      - version_bits
+                    properties:
```
WantClue: The required and properties keywords are over-indented. In OpenAPI YAML, these should be siblings of type: object, not nested under it.

main/http_server/axe-os/src/app/pipes/date-ago.pipe.ts
mutatrum: Yes, it skips zero value elements. Instead of `1h 0m` you would only get `1h`.

main/http_server/http_server.c
mutatrum: Ok, moved inside mutex and copy by reference. No need to copy over the whole thing if it's guarded as well.

main/tasks/scoreboard.c
mutatrum: Added. However, not sure what to do in cases like this, with allocation failures it's going to crash and burn anyways.

main/nvs_config.c
mutatrum: key is the enum, so that won't print the name. Made a separate null check.

mutatrum:
> only nits are `SystemScoreboardEntry` should be moved up to `components.schemas`, and there should be some dupe protection in case the axe is looping and happens to get a high score (somehow i did this) image

How did you do this? Not even sure how to do the duplicate protection either. Is it also submitting the share multiple times to the pool?

0xf0xx0:
my pool fell over while i was gone for a day x3 so it was never submitted, just looping old work.

mutatrum:
> my pool fell over while i was gone for a day x3 so it was never submitted, just looping old work.

I think that would be covered by #911

Travetown:
I would be very curious to know why this PR is not being accepted. 
Four months of maturation does not seem to be enough.

I found it quite exciting to observe the development of the difficulties over several months. But now I will probably delete it with the new firmware. 

Bildschirmfoto 2026-02-22 um 18 06 55

mutatrum:
> I would be very curious to know why this PR is not being accepted. Four months of maturation does not seem to be enough.
> 
> I found it quite exciting to observe the development of the difficulties over several months. But now I will probably delete it with the new firmware.

It'll happen, just need patience. There were several big PRs going in each version already, and this is not a high-prio item.

Edit: Oops, wrong button. Didn't mean to close the PR.

PR #1237, opened 2025-09-16
Destroy subscriptions when leaving page

mutatrum:
Some pages didn't destroy subscriptions when switching to another page. On returning to the page, a new subscription was added, accumulating more and more. Fixed by setting `refCount: true` on `shareReplay`.

### DIFF
--- a/main/http_server/axe-os/src/app/components/system/system.component.ts
+++ b/main/http_server/axe-os/src/app/components/system/system.component.ts
@@ -36,11 +36,11 @@ export class SystemComponent implements OnInit, OnDestroy {
   ) {
     this.info$ = timer(0, 5000).pipe(
       switchMap(() => this.systemService.getInfo()),
-      shareReplay(1)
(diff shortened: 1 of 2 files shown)


duckaxe:
> If Angular's `async pipe` uses the observable in the template, Angular automatically unsubscribes when the component is destroyed.

In both files we use `| async` to load `info$`.

mutatrum:
> By default [shareReplay](https://rxjs.dev/api/index/function/shareReplay) will use refCount of false, meaning that it will not unsubscribe the source when the reference counter drops to zero, i.e. the inner [ReplaySubject](https://rxjs.dev/api/index/class/ReplaySubject) will not be unsubscribed (and potentially run for ever).

Try it out by looking at the polling frequency after switching between home and system several times.

Not sure if we could also solve it by not using `shareReply`?

duckaxe:
That was new to me. Then let's go with your solution.

PR #1239, opened 2025-09-17
Low memory mode

mutatrum:
Skip statistics and bap when no PSRAM detected, adds warnings.

Runs with ~30kb free memory. Without this change, everything starts crashing and AxeOS is not reachable if PSRAM is not there or not configured correctly.

image

image

See #1234 and #826


duckaxe:
@mutatrum Can you pls resolve conflicts so I can push some frontend improvements.

mutatrum:
Haven't seen devices without PSRAM or 2MB PSRAM for quite a while now, closing.

PR #1133, opened 2025-07-05
Cross-Origin Bearer API Authentication

korbin:
This PR:

- Adds mDNS service advertisement for bitaxe hostnames
- Suggests a suffixed hostname (based on the station MAC) for new Bitaxe setups
- Adds an NVS field, constant-time comparison functions, and corresponding frontend form fields for an API Secret token
- Adds Swarm support for scanning and adding cross-origin Bitaxes locally
  - Proxies Swarm local network scanning through the currently-connected backend Bitaxe to avoid CORS issues
- Reworks and simplifies all CORS logic:
  - All cross-origin requests must present an `Authorization: Bearer` header as well as the `X-Requested-With` header (see [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#employing-custom-request-headers-for-ajaxapi))
  - We no longer attempt to nor need determine which network the requester belongs to for cross-origin request safety.
- Exposes current IP/subnet on the info endpoint

This has a number of advantages:

- Hostname-based now access works in many standard scenarios: port forwarding, nginx, Cloudflare tunnels, VPNs, Tailscale, etc.
- Swarm continues to work fairly seamlessly
- Cross-origin API requests are now authenticated (this can easily be adapted to make *all* requests authenticated.)
- Centralized UIs (just HTML pages) can be hosted (think GitHub Pages) and safely access local user Bitaxes (optional)

Possible interesting follow-ons:

- Bitaxe discovery via mDNS
  -  Tokens, services, etc. can be discovered on many more networks automatically without port scanning the whole subnet.

PR #1240, opened 2025-09-18
feat: Add mDNS support for network discovery and dynamic hostname updates

### LINKED ISSUE #233: Add support for mDNS to have pretty dashboard URL
ESP-IDF apparently [supports mDNS](https://docs.espressif.com/projects/esp-idf/en/v5.2.2/esp32s3/api-reference/protocols/mdns.html) which would potentially allow us to have the the AxeOS dashboard have a URL like `bitaxeAB02.local` instead of the IP address.

I suggest using the last 4 hex digits of the WiFi MAC address (like we do for the setup AP SSID) so that there are not problems with multiple Bitaxe on the same network. Of course if there is a better way to do this, that would be awesome too.

### LINKED ISSUE #1118: Error - while update esp-miner.bin and www.bin - Unauthorized
**Describe the bug**
Updates cannot be installed; they run up to 100% and then report an error.
When updating both files, I receive the following error message after the update:

[www.bin](http://www.bin/) -> Error - Http failure response for http://bitaxe/api/system/OTAWWW: 401 Unauthorized
esp-miner.bin -> Unauthorized

This is basically a copy-paste of issue #671. I can't tell if this is a regression or if the bug wasn't actually fixed.

**To Reproduce**

- access the web interface http://bitaxe
- install previous version:  v2.8.1
- download Latest Release: v2.9.0
- start Updates
- error message appears and update is unsuccessful

**Expected behavior**
The update should succeed. 

 - Bitaxe HW version: Gamma 601
 - Bitaxe HW vendor: bitcoinmerch
 - ESP-Miner FW version: v2.8.1
 - Hash Frequency: 625
 - Voltage: 5

**Additional context**
This happens using the url: `http://bitaxe`
If I use `http://bitaxe.lan` it works okay. I have my main router configured to use lan as the local domain.
I know you can use the ip address as a workaround but for not-technical-savvy people this can be frustrating. Using `http://bitaxe` is convenient for new users.

camalolo:
### Summary
This PR implements comprehensive mDNS (Multicast DNS) support for the ESP-Miner device, enabling seamless network discovery and dynamic hostname management. The implementation allows devices to be discovered automatically on local networks using Bonjour/Avahi-compatible services.

#### 🔍 **Network Discovery**
- **mDNS Service Registration**: Device automatically registers as "ESP-Miner Web Server" with HTTP service on port 80
- **Bonjour/Avahi Compatibility**: Device is discoverable using standard network discovery tools
- **Zero-Configuration Networking**: No manual IP address configuration needed for local access

#### 🌐 **Dynamic Hostname Management**
- **Automatic mDNS Hostname Setup**: Hostname is set during WiFi initialization
- **Dynamic Updates**: Hostname changes are reflected in mDNS immediately when updated via web interface
- **Hostname Normalization**: Prevents duplicate ".local.local" suffixes by stripping existing ".local" before setting

#### 🛡️ **Enhanced Security & Usability**
- **Simplified CORS Logic**: Streamlined network access validation for private IP ranges
- **Graceful Degradation**: mDNS failures don't prevent device operation, only log warnings
- **Comprehensive Logging**: Detailed status messages for troubleshooting

#### Files Modified:
1. **`components/connect/CMakeLists.txt`** - Added mDNS component dependency
2. **`components/connect/connect.c`** - Core mDNS initialization and management
3. **`components/connect/include/connect.h`** - Public API for hostname updates
4. **`main/http_server/http_server.c`** - Dynamic hostname updates and CORS improvements
5. **`main/idf_component.yml`** - Added espressif/mdns dependency

#### Key Functions Added:
- `mdns_init()` - Initializes mDNS service
- `mdns_hostname_set()` - Sets device hostname for discovery
- `mdns_service_add()` - Registers HTTP service
- `update_mdns_hostname()` - Dynamic hostname updates
- `normalize_hostname()` - Prevents hostname conflicts

#### Network Discovery:
```bash
# Discover device using avahi-browse
avahi-browse _http._tcp

# Access via mDNS hostname
curl http://esp-miner-XXXX.local
```

#### Web Interface Updates:
- Hostname changes in the web UI now immediately update mDNS
- Device remains discoverable with new hostname without restart

### Benefits

✅ **Improved User Experience**: No need to remember IP addresses  
✅ **Network-Friendly**: Works with existing Bonjour/Avahi infrastructure  
✅ **Dynamic Configuration**: Hostname changes take effect immediately  
✅ **Backward Compatible**: Existing functionality remains unchanged  
✅ **Robust Error Handling**: Graceful degradation on mDNS failures  

### Testing Considerations

- Verify device appears in network discovery tools (avahi-browse, Bonjour Browser)
- Test hostname updates via web interface
- Confirm HTTP access works via `.local` hostname
- Test behavior when mDNS initialization fails
- Validate CORS still works for private network access

### Dependencies Added
- `espressif/mdns: ^1.8.0` - Official ESP-IDF mDNS component

This implementation provides a solid foundation for network discovery and makes the ESP-Miner much more user-friendly in home and enterprise networks.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -20,7 +20,7 @@ jobs:
         with:
           esp_idf_version: v5.5.3
           target: esp32s3
-          command: idf.py add-dependency "espressif/mdns^1.8.0" && GITHUB_ACTIONS="true" idf.py build
+          command: GITHUB_ACTIONS="true" idf.py build
           path: 'test-ci'
 
       - name: Run tests
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -20,7 +20,7 @@ jobs:
         with:
           esp_idf_version: v5.5.3
           target: esp32s3
-          command: GITHUB_ACTIONS="true" idf.py build
+          command: idf.py add-dependency "espressif/mdns^1.8.0" && GITHUB_ACTIONS="true" idf.py build
           path: 'test-ci'
 
       - name: Run tests
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -65,6 +65,7 @@ static wifi_ap_record_t ap_info[MAX_AP_COUNT];
 static int s_retry_num = 0;
 static int clients_connected_to_ap = 0;
 static bool mdns_initialized = false;
+static bool mdns_init_in_progress = false;
 
 static const char *get_wifi_reason_string(int reason);
 static void wifi_softap_on(void);
@@ -107,8 +108,28 @@ esp_err_t wifi_apply_hostname(const char *hostname)
     return ESP_OK;
 }
 
+static void initialize_mdns_if_needed(GlobalState *GLOBAL_STATE);
 static char* generate_unique_hostname(const char *base);
 static char* check_and_resolve_hostname_conflict(const char *hostname, const char *current_ip);
+
+static void mdns_init_task(void *pvParameters) {
+    GlobalState *GLOBAL_STATE = (GlobalState *)pvParameters;
+    initialize_mdns_if_needed(GLOBAL_STATE);
+    mdns_init_in_progress = false;
+    vTaskDelete(NULL);
+}
+
+static void spawn_mdns_init_if_needed(GlobalState *GLOBAL_STATE) {
+    if (mdns_initialized || mdns_init_in_progress) {
+        return;
+    }
+    mdns_init_in_progress = true;
+    BaseType_t ret = xTaskCreate(mdns_init_task, "mdns_init", 4096, GLOBAL_STATE, 5, NULL);
+    if (ret != pdPASS) {
+        ESP_LOGW(TAG, "Failed to create mDNS init task");
+        mdns_init_in_progress = false;
+    }
+}
 
 static void initialize_mdns_if_needed(GlobalState *GLOBAL_STATE) {
     if (mdns_initialized) {
@@ -412,7 +433,7 @@ static void event_handler(void * arg, esp_event_base_t event_base, int32_t event
             ESP_LOGE(TAG, "Failed to create IPv6 link-local address: %s", esp_err_to_name(ipv6_err));
         }
 
-        initialize_mdns_if_needed(GLOBAL_STATE);
+        spawn_mdns_init_if_needed(GLOBAL_STATE);
     }
 
     if (event_base == IP_EVENT && event_id == IP_EVENT_GOT_IP6) {
@@ -445,7 +466,7 @@ static void event_handler(void * arg, esp_event_base_t event_base, int32_t event
             ESP_LOGI(TAG, "IPv6 Address: %s", GLOBAL_STATE->SYSTEM_MODULE.ipv6_addr_str);
         }
 
-        initialize_mdns_if_needed(GLOBAL_STATE);
+        spawn_mdns_init_if_needed(GLOBAL_STATE);
     }
 }
 
@@ -518,7 +539,7 @@ static char* generate_unique_hostname(const char *base) {
 
 static char* check_and_resolve_hostname_conflict(const char *hostname, const char *current_ip) {
     mdns_result_t *results = NULL;
-    esp_err_t err = mdns_query_generic(hostname, NULL, NULL, MDNS_TYPE_A, MDNS_QUERY_MULTICAST, 3000, 1, &results);
+    esp_err_t err = mdns_query_generic(hostname, NULL, NULL, MDNS_TYPE_A, MDNS_QUERY_MULTICAST, 1000, 1, &results);
     if (err != ESP_OK || !results || !results->addr) {
         // No A record found, no conflict
         if (results) mdns_query_results_free(results);
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -2,10 +2,19 @@
 #include "esp_event.h"
 #include "esp_log.h"
 #include "esp_netif.h"
+#include "mdns.h"
+#include "esp_system.h"
 #include "esp_wifi.h"
+#include "freertos/FreeRTOS.h"
+#include "freertos/event_groups.h"
 #include "freertos/task.h"
 #include "freertos/timers.h"
 #include "lwip/err.h"
+#include "lwip/lwip_napt.h"
+#include "lwip/sys.h"
+#include "lwip/sockets.h"
+#include "esp_netif.h"
+#include "nvs_flash.h"
 #include "esp_wifi_types_generic.h"
 
 #include "connect.h"
@@ -55,6 +64,7 @@ static uint16_t ap_number = 0;
 static wifi_ap_record_t ap_info[MAX_AP_COUNT];
 static int s_retry_num = 0;
 static int clients_connected_to_ap = 0;
+static bool mdns_initialized = false;
 
 static const char *get_wifi_reason_string(int reason);
 static void wifi_softap_on(void);
@@ -94,6 +104,109 @@ esp_err_t wifi_apply_hostname(const char *hostname)
         return err;
     }
 
+    return ESP_OK;
+}
+
+static char* generate_unique_hostname(const char *base);
+static char* check_and_resolve_hostname_conflict(const char *hostname, const char *current_ip);
+
+static void initialize_mdns_if_needed(GlobalState *GLOBAL_STATE) {
+    if (mdns_initialized) {
+        return;
+    }
+
+    char * hostname = nvs_config_get_string(NVS_CONFIG_HOSTNAME);
+    if (hostname == NULL) {
+        ESP_LOGW(TAG, "Hostname not configured, skipping mDNS setup");
+        return;
+    }
+    esp_err_t err = mdns_init();
+    if (err != ESP_OK) {
+        ESP_LOGW(TAG, "mDNS/Avahi initialization failed: %s", esp_err_to_name(err));
+        ESP_LOGW(TAG, "Device will not be discoverable via mDNS/Bonjour/Avahi");
+    } else {
+        ESP_LOGI(TAG, "mDNS/Avahi initialized successfully - device discoverable on network");
+
+        /* Get current IP */
+        esp_netif_ip_info_t ip_info;
+        esp_netif_get_ip_info(esp_netif_get_handle_from_ifkey("WIFI_STA_DEF"), &ip_info);
+        char current_ip[16];
+        snprintf(current_ip, sizeof(current_ip), IPSTR, IP2STR(&ip_info.ip));
+
+        /* Check for hostname conflicts */
+        char *final_hostname = check_and_resolve_hostname_conflict(hostname, current_ip);
+
+        /* Set mDNS hostname */
+        err = mdns_hostname_set(final_hostname);
+        if (err != ESP_OK) {
+            ESP_LOGW(TAG, "mDNS hostname setup failed: %s", esp_err_to_name(err));
+            ESP_LOGW(TAG, "Device hostname not set for mDNS discovery");
+        } else {
+            ESP_LOGI(TAG, "mDNS hostname set to: %s.local", final_hostname);
+            ESP_LOGI(TAG, "Access device at: http://%s.local", final_hostname);
+        }
+
+        free(final_hostname);
+
+        /* Set mDNS instance name */
+        uint8_t mac[6];
+        esp_wifi_get_mac(WIFI_IF_STA, mac);
+        char mac_suffix[6];
+        snprintf(mac_suffix, sizeof(mac_suffix), "%02X%02X", mac[4], mac[5]);
+
+        char instance_name[64];
+        snprintf(instance_name, sizeof(instance_name), "Bitaxe %s %s (%s)",
+                 GLOBAL_STATE->DEVICE_CONFIG.family.name,
+                 GLOBAL_STATE->DEVICE_CONFIG.board_version,
+                 mac_suffix);
+        
+        /* Add HTTP service */
+        err = mdns_service_add(instance_name, "_http", "_tcp", 80, NULL, 0);
+        if (err != ESP_OK) {
+            ESP_LOGW(TAG, "mDNS HTTP service registration failed: %s", esp_err_to_name(err));
+            ESP_LOGW(TAG, "HTTP service not advertised via mDNS");
+        } else {
+            ESP_LOGI(TAG, "mDNS HTTP service registered: _http._tcp port 80");
+            ESP_LOGI(TAG, "Discover with: avahi-browse _http._tcp");
+            ESP_LOGI(TAG, "mDNS instance: %s", instance_name);
+        }
+
+        ESP_LOGI(TAG, "mDNS/Avahi setup complete - device ready for network discovery");
+        mdns_initialized = true;
+    }
+    free(hostname);
+}
+
+esp_err_t update_mdns_hostname(const char *new_hostname) {
+    if (new_hostname == NULL || strlen(new_hostname) == 0) {
+        ESP_LOGW(TAG, "Invalid hostname provided for mDNS update");
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    /* Get current IP for conflict checking */
+    esp_netif_ip_info_t ip_info;
+    esp_netif_get_ip_info(esp_netif_get_handle_from_ifkey("WIFI_STA_DEF"), &ip_info);
+    char current_ip[16];
+    snprintf(current_ip, sizeof(current_ip), IPSTR, IP2STR(&ip_info.ip));
+
+    /* Check for hostname conflicts and resolve if needed */
+    char *resolved_hostname = check_and_resolve_hostname_conflict(new_hostname, current_ip);
+
+    /* If the hostname was resolved to a different one, update NVS */
+    if (strcmp(resolved_hostname, new_hostname) != 0) {
+        nvs_config_set_string(NVS_CONFIG_HOSTNAME, resolved_hostname);
+        ESP_LOGI(TAG, "Hostname conflict resolved, updated NVS to: %s", resolved_hostname);
+    }
+
+    esp_err_t err = mdns_hostname_set(resolved_hostname);
+    if (err != ESP_OK) {
+        ESP_LOGW(TAG, "Failed to update mDNS hostname to: %s, error: %s", resolved_hostname, esp_err_to_name(err));
+        free(resolved_hostname);
+        return err;
+    }
+
+    ESP_LOGI(TAG, "mDNS hostname updated to: %s", resolved_hostname);
+    free(resolved_hostname);
     return ESP_OK;
 }
 
@@ -292,6 +405,8 @@ static void event_handler(void * arg, esp_event_base_t event_base, int32_t event
         if (ipv6_err != ESP_OK) {
             ESP_LOGE(TAG, "Failed to create IPv6 link-local address: %s", esp_err_to_name(ipv6_err));
         }
+
+        initialize_mdns_if_needed(GLOBAL_STATE);
     }
 
     if (event_base == IP_EVENT && event_id == IP_EVENT_GOT_IP6) {
@@ -323,6 +438,8 @@ static void event_handler(void * arg, esp_event_base_t event_base, int32_t event
             GLOBAL_STATE->SYSTEM_MODULE.ipv6_addr_str[sizeof(GLOBAL_STATE->SYSTEM_MODULE.ipv6_addr_str) - 1] = '\0';
             ESP_LOGI(TAG, "IPv6 Address: %s", GLOBAL_STATE->SYSTEM_MODULE.ipv6_addr_str);
         }
+
+        initialize_mdns_if_needed(GLOBAL_STATE);
     }
 }
 
@@ -378,6 +495,49 @@ static void wifi_softap_off(void)
     if (is_wifi_operation_allowed(err)) {
         ESP_ERROR_CHECK(err);
     }
+}
+
+
+
+static char* generate_unique_hostname(const char *base) {
+    uint8_t mac[6];
+    esp_wifi_get_mac(WIFI_IF_STA, mac);
+    char suffix[6];
+    snprintf(suffix, sizeof(suffix), "-%02x%02x", mac[4], mac[5]);
+    char *new_hostname = malloc(strlen(base) + strlen(suffix) + 1);
+    strcpy(new_hostname, base);
+    strcat(new_hostname, suffix);
+    return new_hostname;
+}
+
+static char* check_and_resolve_hostname_conflict(const char *hostname, const char *current_ip) {
+    mdns_result_t *results = NULL;
+    esp_err_t err = mdns_query_generic(hostname, NULL, NULL, MDNS_TYPE_A, MDNS_QUERY_MULTICAST, 3000, 1, &results);
+    if (err != ESP_OK || !results || !results->addr) {
+        // No A record found, no conflict
+        if (results) mdns_query_results_free(results);
+        return strdup(hostname);
+    }
+
+    mdns_ip_addr_t *a = results->addr;
+    char ip_str[INET6_ADDRSTRLEN];
+    if (a->addr.type == IPADDR_TYPE_V4) {
+        esp_ip4addr_ntoa(&a->addr.u_addr.ip4, ip_str, sizeof(ip_str));
+    } else {
+        inet_ntop(AF_INET6, &a->addr.u_addr.ip6, ip_str, sizeof(ip_str));
+    }
+
+    if (strcmp(ip_str, current_ip) != 0) {
+        // Different IP, conflict detected
+        char *new_hostname = generate_unique_hostname(hostname);
+        ESP_LOGI(TAG, "mDNS conflict detected for '%s' at %s, renaming to '%s'", hostname, ip_str, new_hostname);
+        nvs_config_set_string(NVS_CONFIG_HOSTNAME, new_hostname);
+        mdns_query_results_free(results);
+        return new_hostname;
+    }
+
+    mdns_query_results_free(results);
+    return strdup(hostname);
 }
 
 static void wifi_softap_on(void)
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -144,6 +144,8 @@ static void initialize_mdns_if_needed(GlobalState *GLOBAL_STATE) {
         } else {
             ESP_LOGI(TAG, "mDNS hostname set to: %s.local", final_hostname);
             ESP_LOGI(TAG, "Access device at: http://%s.local", final_hostname);
+            strlcpy(GLOBAL_STATE->SYSTEM_MODULE.mdns_hostname, final_hostname, sizeof(GLOBAL_STATE->SYSTEM_MODULE.mdns_hostname));
+            snprintf(GLOBAL_STATE->SYSTEM_MODULE.full_hostname, sizeof(GLOBAL_STATE->SYSTEM_MODULE.full_hostname), "%s.local", final_hostname);
         }
 
         free(final_hostname);
@@ -177,7 +179,7 @@ static void initialize_mdns_if_needed(GlobalState *GLOBAL_STATE) {
     free(hostname);
 }
 
-esp_err_t update_mdns_hostname(const char *new_hostname) {
+esp_err_t update_mdns_hostname(const char *new_hostname, GlobalState *GLOBAL_STATE) {
     if (new_hostname == NULL || strlen(new_hostname) == 0) {
         ESP_LOGW(TAG, "Invalid hostname provided for mDNS update");
         return ESP_ERR_INVALID_ARG;
@@ -206,6 +208,10 @@ esp_err_t update_mdns_hostname(const char *new_hostname) {
     }
 
     ESP_LOGI(TAG, "mDNS hostname updated to: %s", resolved_hostname);
+    if (GLOBAL_STATE != NULL) {
+        strlcpy(GLOBAL_STATE->SYSTEM_MODULE.mdns_hostname, resolved_hostname, sizeof(GLOBAL_STATE->SYSTEM_MODULE.mdns_hostname));
+        snprintf(GLOBAL_STATE->SYSTEM_MODULE.full_hostname, sizeof(GLOBAL_STATE->SYSTEM_MODULE.full_hostname), "%s.local", resolved_hostname);
+    }
     free(resolved_hostname);
     return ESP_OK;
 }
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -156,6 +156,11 @@ static void initialize_mdns_if_needed(GlobalState *GLOBAL_STATE) {
 
         /* Check for hostname conflicts */
         char *final_hostname = check_and_resolve_hostname_conflict(hostname, current_ip);
+        if (final_hostname == NULL) {
+            ESP_LOGE(TAG, "Failed to resolve hostname conflicts, skipping mDNS hostname setup");
+            free(hostname);
+            return;
+        }
 
         /* Set mDNS hostname */
         err = mdns_hostname_set(final_hostname);
@@ -214,6 +219,10 @@ esp_err_t update_mdns_hostname(const char *new_hostname, GlobalState *GLOBAL_STA
 
     /* Check for hostname conflicts and resolve if needed */
     char *resolved_hostname = check_and_resolve_hostname_conflict(new_hostname, current_ip);
+    if (resolved_hostname == NULL) {
+        ESP_LOGE(TAG, "Failed to resolve hostname conflicts");
+        return ESP_ERR_NO_MEM;
+    }
 
     /* If the hostname was resolved to a different one, update NVS */
     if (strcmp(resolved_hostname, new_hostname) != 0) {
@@ -532,6 +541,10 @@ static char* generate_unique_hostname(const char *base) {
     char suffix[6];
     snprintf(suffix, sizeof(suffix), "-%02x%02x", mac[4], mac[5]);
     char *new_hostname = malloc(strlen(base) + strlen(suffix) + 1);
+    if (new_hostname == NULL) {
+        ESP_LOGE(TAG, "Failed to allocate memory for unique hostname");
+        return NULL;
+    }
     strcpy(new_hostname, base);
     strcat(new_hostname, suffix);
     return new_hostname;
@@ -555,8 +568,12 @@ static char* check_and_resolve_hostname_conflict(const char *hostname, const cha
     }
 
     if (strcmp(ip_str, current_ip) != 0) {
-        // Different IP, conflict detected
         char *new_hostname = generate_unique_hostname(hostname);
+        if (new_hostname == NULL) {
+            ESP_LOGW(TAG, "mDNS conflict detected for '%s' but could not generate unique name, keeping original", hostname);
+            mdns_query_results_free(results);
+            return strdup(hostname);
+        }
         ESP_LOGI(TAG, "mDNS conflict detected for '%s' at %s, renaming to '%s'", hostname, ip_str, new_hostname);
         nvs_config_set_string(NVS_CONFIG_HOSTNAME, new_hostname);
         mdns_query_results_free(results);
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -20,6 +20,7 @@
 #include "connect.h"
 #include "global_state.h"
 #include "nvs_config.h"
+#include "esp_app_desc.h"
 
 // Maximum number of access points to scan
 #define MAX_AP_COUNT 20
@@ -206,6 +207,27 @@ static void initialize_mdns_if_needed(GlobalState *GLOBAL_STATE) {
                 ESP_LOGI(TAG, "mDNS AxeOS subtype registered: _axeos._sub._http._tcp");
                 ESP_LOGI(TAG, "Discover AxeOS devices with: avahi-browse _axeos._sub._http._tcp");
             }
+
+            /* Add TXT records for device identification */
+            err = mdns_service_txt_item_set_for_host(instance_name, "_http", "_tcp", NULL, "board", GLOBAL_STATE->DEVICE_CONFIG.board_version);
+            if (err != ESP_OK) ESP_LOGW(TAG, "mDNS TXT 'board' failed: %s", esp_err_to_name(err));
+            err = mdns_service_txt_item_set_for_host(instance_name, "_http", "_tcp", NULL, "family", GLOBAL_STATE->DEVICE_CONFIG.family.name);
+            if (err != ESP_OK) ESP_LOGW(TAG, "mDNS TXT 'family' failed: %s", esp_err_to_name(err));
+            err = mdns_service_txt_item_set_for_host(instance_name, "_http", "_tcp", NULL, "asic", GLOBAL_STATE->DEVICE_CONFIG.family.asic.name);
+            if (err != ESP_OK) ESP_LOGW(TAG, "mDNS TXT 'asic' failed: %s", esp_err_to_name(err));
+            char asic_count_str[4];
+            snprintf(asic_count_str, sizeof(asic_count_str), "%u", GLOBAL_STATE->DEVICE_CONFIG.family.asic_count);
+            err = mdns_service_txt_item_set_for_host(instance_name, "_http", "_tcp", NULL, "asic_count", asic_count_str);
+            if (err != ESP_OK) ESP_LOGW(TAG, "mDNS TXT 'asic_count' failed: %s", esp_err_to_name(err));
+            const esp_app_desc_t *app_desc = esp_app_get_description();
+            err = mdns_service_txt_item_set_for_host(instance_name, "_http", "_tcp", NULL, "fw_version", app_desc->version);
+            if (err != ESP_OK) ESP_LOGW(TAG, "mDNS TXT 'fw_version' failed: %s", esp_err_to_name(err));
+            ESP_LOGI(TAG, "mDNS TXT records added: board=%s, family=%s, asic=%s, asic_count=%s, fw_version=%s",
+                     GLOBAL_STATE->DEVICE_CONFIG.board_version,
+                     GLOBAL_STATE->DEVICE_CONFIG.family.name,
+                     GLOBAL_STATE->DEVICE_CONFIG.family.asic.name,
+                     asic_count_str,
+                     app_desc->version);
         }
 
         ESP_LOGI(TAG, "mDNS/Avahi setup complete - device ready for network discovery");
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -197,6 +197,15 @@ static void initialize_mdns_if_needed(GlobalState *GLOBAL_STATE) {
             ESP_LOGI(TAG, "mDNS HTTP service registered: _http._tcp port 80");
             ESP_LOGI(TAG, "Discover with: avahi-browse _http._tcp");
             ESP_LOGI(TAG, "mDNS instance: %s", instance_name);
+
+            /* Add AxeOS subtype for DNS-SD discovery */
+            err = mdns_service_subtype_add_for_host(instance_name, "_http", "_tcp", NULL, "_axeos");
+            if (err != ESP_OK) {
+                ESP_LOGW(TAG, "mDNS AxeOS subtype registration failed: %s", esp_err_to_name(err));
+            } else {
+                ESP_LOGI(TAG, "mDNS AxeOS subtype registered: _axeos._sub._http._tcp");
+                ESP_LOGI(TAG, "Discover AxeOS devices with: avahi-browse _axeos._sub._http._tcp");
+            }
         }
 
         ESP_LOGI(TAG, "mDNS/Avahi setup complete - device ready for network discovery");
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -253,7 +253,11 @@ esp_err_t update_mdns_hostname(const char *new_hostname, GlobalState *GLOBAL_STA
         ip_info.ip.addr = 0;
     }
     char current_ip[16];
-    snprintf(current_ip, sizeof(current_ip), IPSTR, IP2STR(&ip_info.ip));
+    if (ip_info.ip.addr == 0) {
+        strlcpy(current_ip, "0.0.0.0", sizeof(current_ip));
+    } else {
+        snprintf(current_ip, sizeof(current_ip), IPSTR, IP2STR(&ip_info.ip));
+    }
 
     /* Check for hostname conflicts and resolve if needed */
     char *resolved_hostname = check_and_resolve_hostname_conflict(new_hostname, current_ip);
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -151,9 +151,17 @@ static void initialize_mdns_if_needed(GlobalState *GLOBAL_STATE) {
 
         /* Get current IP */
         esp_netif_ip_info_t ip_info;
-        esp_netif_get_ip_info(esp_netif_get_handle_from_ifkey("WIFI_STA_DEF"), &ip_info);
+        esp_netif_t *netif = esp_netif_get_handle_from_ifkey("WIFI_STA_DEF");
+        if (netif == NULL) {
+            netif = esp_netif_get_handle_from_ifkey("WIFI_AP_DEF");
+        }
         char current_ip[16];
-        snprintf(current_ip, sizeof(current_ip), IPSTR, IP2STR(&ip_info.ip));
+        if (netif == NULL || esp_netif_get_ip_info(netif, &ip_info) != ESP_OK || ip_info.ip.addr == 0) {
+            ESP_LOGW(TAG, "No active network interface for mDNS init, using 0.0.0.0");
+            strlcpy(current_ip, "0.0.0.0", sizeof(current_ip));
+        } else {
+            snprintf(current_ip, sizeof(current_ip), IPSTR, IP2STR(&ip_info.ip));
+        }
 
         /* Check for hostname conflicts */
         char *final_hostname = check_and_resolve_hostname_conflict(hostname, current_ip);
@@ -161,6 +169,12 @@ static void initialize_mdns_if_needed(GlobalState *GLOBAL_STATE) {
             ESP_LOGE(TAG, "Failed to resolve hostname conflicts, skipping mDNS hostname setup");
             free(hostname);
             return;
+        }
+
+        /* If conflict resolution changed the hostname, persist to NVS */
+        if (strcmp(final_hostname, hostname) != 0) {
+            nvs_config_set_string(NVS_CONFIG_HOSTNAME, final_hostname);
+            ESP_LOGI(TAG, "Hostname conflict resolved, updated NVS to: %s", final_hostname);
         }
 
         /* Set mDNS hostname */
@@ -617,7 +631,6 @@ static char* check_and_resolve_hostname_conflict(const char *hostname, const cha
             return strdup(hostname);
         }
         ESP_LOGI(TAG, "mDNS conflict detected for '%s' at %s, renaming to '%s'", hostname, ip_str, new_hostname);
-        nvs_config_set_string(NVS_CONFIG_HOSTNAME, new_hostname);
         mdns_query_results_free(results);
         return new_hostname;
     }
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -244,7 +244,14 @@ esp_err_t update_mdns_hostname(const char *new_hostname, GlobalState *GLOBAL_STA
 
     /* Get current IP for conflict checking */
     esp_netif_ip_info_t ip_info;
-    esp_netif_get_ip_info(esp_netif_get_handle_from_ifkey("WIFI_STA_DEF"), &ip_info);
+    esp_netif_t *netif = esp_netif_get_handle_from_ifkey("WIFI_STA_DEF");
+    if (netif == NULL) {
+        netif = esp_netif_get_handle_from_ifkey("WIFI_AP_DEF");
+    }
+    if (netif == NULL || esp_netif_get_ip_info(netif, &ip_info) != ESP_OK) {
+        ESP_LOGW(TAG, "No active network interface for mDNS hostname update");
+        ip_info.ip.addr = 0;
+    }
     char current_ip[16];
     snprintf(current_ip, sizeof(current_ip), IPSTR, IP2STR(&ip_info.ip));
 
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
@@ -542,8 +542,8 @@ private isIpAddress(value: string): boolean {
 
   get sortOptions() {
     return [
-      { label: 'Hostname (Z-A)', value: { sortField: 'hostname', sortDirection: 'desc' } },
-      { label: 'Hostname (A-Z)', value: { sortField: 'hostname', sortDirection: 'asc' } },
+      { label: 'Hostname', value: { sortField: 'hostname', sortDirection: 'desc' } },
+      { label: 'Hostname', value: { sortField: 'hostname', sortDirection: 'asc' } },
       { label: 'Address', value: { sortField: 'address', sortDirection: 'desc' } },
       { label: 'Address', value: { sortField: 'address', sortDirection: 'asc' } },
       { label: 'Hashrate', value: { sortField: 'hashRate', sortDirection: 'desc' } },
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
@@ -597,7 +597,17 @@ return this.swarm.filter(axe =>
     }
   }
 
-  isThisDevice(IP: string): boolean {
-    return IP === window.location.hostname || (this.currentDeviceIp !== null && IP === this.currentDeviceIp);
+  isThisDevice(device: SwarmDevice): boolean {
+    const hostname = window.location.hostname;
+    
+    if (device.address === hostname || device.connectionAddress === hostname) {
+      return true;
+    }
+    
+    if (this.currentDeviceIp !== null && device['ip'] === this.currentDeviceIp) {
+      return true;
+    }
+    
+    return false;
   }
 }
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
@@ -13,7 +13,16 @@ const SWARM_REFRESH_TIME = 'SWARM_REFRESH_TIME';
 const SWARM_SORTING = 'SWARM_SORTING';
 const SWARM_GRID_VIEW = 'SWARM_GRID_VIEW';
 
-type SwarmDevice = { IP: string; ASICModel: string; deviceModel: string; swarmColor: string; asicCount: number; [key: string]: any };
+type SwarmDevice = { 
+  address: string; 
+  ASICModel: string; 
+  deviceModel: string; 
+  swarmColor: string; 
+  asicCount: number; 
+  displayName?: string; 
+  connectionAddress?: string; 
+  [key: string]: any 
+};
 
 @Component({
   selector: 'app-swarm',
@@ -50,6 +59,8 @@ export class SwarmComponent implements OnInit, OnDestroy {
 
   public filterText = '';
 
+  public currentDeviceIp: string | null = null;
+
   @HostListener('document:keydown.esc', ['$event'])
   onEscKey() {
     if (this.filterText) {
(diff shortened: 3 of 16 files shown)


main/http_server/http_server.c
```diff
@@ -186,28 +174,17 @@ esp_err_t is_network_allowed(httpd_req_t * req)
...
-
-    // Convert to IPv4 string
+    // Convert to IPv4 string for logging
     inet_ntop(AF_INET, &request_ip_addr, ipstr, sizeof(ipstr));
+    ESP_LOGD(CORS_TAG, "Request from IP: %s", ipstr);
 
-    // Attempt to get the Origin header.
-    char origin[128];
-    uint32_t origin_ip_addr;
-    if (httpd_req_get_hdr_value_str(req, "Origin", origin, sizeof(origin)) == ESP_OK) {
```
mutatrum, line 310: What's the rationale of removing Origin header parsing?

camalolo: It allows accessing the web UI via the .local address (example bitaxe.local) and save the settings - without it, there's an authentication error when you try and save. Thank you for your time !

mutatrum: IIRC this is preventing CSRF from malicious websites. Allowing all requests from private networks is too permissive.

camalolo: I will have a look and check if I can do it better. Maybe I can allow requests only from the IP or the mDNS generated hostname. Thank you for your valuable feedback !

mutatrum: I think that was my initial idea as well, so only extend the current allowed requests with the mDNS hostname. Looking forward to this, having hostnames work would be a really nice fix!

mutatrum:
One other thing to consider if the swarm screen is working with mDNS hostnames.

main/http_server/http_server.c
camalolo: @mutatrum I added a check on the Origin header for the mDNS hostname. Tested it and it seemed to work in my environment.

However, regarding the check on the origin for access via IP address (http://192.168.1.x) I did not restore the previous approach, and here is my rationale : the Origin header can be changed trivially. However, the actual request source IP is not so easy to fake. So the approach of checking the source IP might actually be more secure than relying on the Origin header.

Didn't check the swarm page yet as I wanted to know what you think of that.

I might be completely wrong, so I particularly welcome your input.

mutatrum: The local IP is a verification that an publicly exposed Bitaxe doesn't just accept requests from the internet. Of course, here, the Origin header is useless, as anyone can set it.

However, that's not the attack vector. A malicious website can - from one of your browser tabs - easily try each local IP and see if there's a Bitaxe device there, and change the mining credentials that way. This has been proven to work and has been seen in the wild. In this case, it comes from a local IP. However, the `Origin` header cannot be changed, javascript can't do that from within the browser sandbox.

At least, this is as far as I understood it, although I think I'm missing some details here. There's a write-up: https://snotra.uk/axeos-csrf-vulnerability.html

camalolo: Makes sense ! I'll update the commit with both the original IP check and the mDNS support. Thank you !

camalolo:
Regarding http_server.c and CORS : I believe I restored the original checks, while adding a block that also allows requests with Origin set to the mDNS host.

.github/workflows/unittest.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         with:
           esp_idf_version: v5.5
           target: esp32s3
-          command: GITHUB_ACTIONS="true" idf.py build
+          command: idf.py add-dependency "espressif/mdns^1.8.0" ; GITHUB_ACTIONS="true" idf.py build
```
mutatrum: Why is this needed? The other dependencies are picked up automatically?

camalolo: I admit I don't know why - the CI task failed on GitHub without it :

HINT: The component 'espressif__mdns' could not be found. This could be because: component name was misspelled, the component was not added to the build, the component has been moved to the IDF component manager, the component has been removed and refactored into some other component or the component may not be supported by the selected target.

https://github.com/bitaxeorg/ESP-Miner/actions/runs/17818969270/job/50662335727

mutatrum: Apparently that's what you have to do. Unsure why it's different, seems it's not part of the normal component repo? See https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/mdns.html

mutatrum:
This was also part of #1133, but it's good it's a separate API. Code looks good but haven't tested it yet, I'll give it a spin tomorrow.

camalolo:
> One other thing to consider if the swarm screen is working with mDNS hostnames.

Preliminary work shows it does not. Other bitaxes with no support for mDNS hostnames reject the requests to /api/system/info as they violate CORS when sent from a swarm page that runs on a domain such as http://bitaxe.local. So far the best idea that might work could be to proxy the requests thru the backend of the current device, but I have concerns of an automatic scan for hundreds of hosts thru the backend.

mutatrum:
> > One other thing to consider if the swarm screen is working with mDNS hostnames.
> 
> Preliminary work shows it does not. Other bitaxes with no support for mDNS hostnames reject the requests to /api/system/info as they violate CORS when sent from a swarm page that runs on a domain such as http://bitaxe.local. So far the best idea that might work could be to proxy the requests thru the backend of the current device, but I have concerns of an automatic scan for hundreds of hosts thru the backend.

No, swarm should stay front-end. 

So is this correct?

| swarm device | client device | result |
| --- | --- | --- |
| ip | ip | works |
| ip | hostname | works |
| hostname | ip | CORS |
| hostname | hostname | works |

If this is the case, then I guess it's fine for now? Does mean that NerdAxes need this at some point as well.

Unfortunately I only have a single device at the moment, so I can't test this.

camalolo:
> > > One other thing to consider if the swarm screen is working with mDNS hostnames.
> > 
> > 
> > Preliminary work shows it does not. Other bitaxes with no support for mDNS hostnames reject the requests to /api/system/info as they violate CORS when sent from a swarm page that runs on a domain such as http://bitaxe.local. So far the best idea that might work could be to proxy the requests thru the backend of the current device, but I have concerns of an automatic scan for hundreds of hosts thru the backend.
> 
> No, swarm should stay front-end.
> 
> So is this correct?
> 
> swarm device	client device	result
> ip	ip	works
> ip	hostname	works
> hostname	ip	CORS
> hostname	hostname	works
> If this is the case, then I guess it's fine for now? Does mean that NerdAxes need this at some point as well.
> 
> Unfortunately I only have a single device at the moment, so I can't test this.

I'll try again implementing it on the swarm page, maybe my first attempt was incorrect.

The swarm page accessible at http://bitaxe.local/#/swarm would try to connect to for example http://192.168.1.10/api/system/info, but because the Origin header is http://bitaxe.local/, the device at 192.168.1.10 replied with a 401 (because it runs a firmware without mDNS support, so it did not understand the bitaxe.local Origin header).

If you think that shouldn't be an issue, then I'll try again to implement it and give feedback if I succeed.

camalolo:
> > > > One other thing to consider if the swarm screen is working with mDNS hostnames.
> > > 
> > > 
> > > Preliminary work shows it does not. Other bitaxes with no support for mDNS hostnames reject the requests to /api/system/info as they violate CORS when sent from a swarm page that runs on a domain such as http://bitaxe.local. So far the best idea that might work could be to proxy the requests thru the backend of the current device, but I have concerns of an automatic scan for hundreds of hosts thru the backend.
> > 
> > 
> > No, swarm should stay front-end.
> > So is this correct?
> > swarm device	client device	result
> > ip	ip	works
> > ip	hostname	works
> > hostname	ip	CORS
> > hostname	hostname	works
> > If this is the case, then I guess it's fine for now? Does mean that NerdAxes need this at some point as well.
> > Unfortunately I only have a single device at the moment, so I can't test this.
> 
> I'll try again implementing it on the swarm page, maybe my first attempt was incorrect.
> 
> The swarm page accessible at http://bitaxe.local/#/swarm would try to connect to for example http://192.168.1.10/api/system/info, but because the Origin header is http://bitaxe.local/, the device at 192.168.1.10 replied with a 401 (because it runs a firmware without mDNS support, so it did not understand the bitaxe.local Origin header).
> 
> If you think that shouldn't be an issue, then I'll try again to implement it and give feedback if I succeed.

Got the same issue after starting from scratch. Basically, if the other bitaxe(s) on the LAN don't have the mDNS patch, they cannot be accessed because of the URL on the bitaxe you are viewing the swarm page on.

### The Specific Issue
1. **Your current page**: Served from `http://espressif.local`
2. **Target device**: `http://192.168.1.114` 
3. **Browser security**: Blocks the XMLHttpRequest because `espressif.local` ≠ `192.168.1.114`
4. **Chrome console shows**: Access to XMLHttpRequest at 'http://192.168.1.114/api/system/info' from origin 'http://espressif.local' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

### Why It Happens
- The browser sends a preflight OPTIONS request to `192.168.1.114`
- The ESP device responds without CORS headers
- Browser blocks the actual request and shows: `"No 'Access-Control-Allow-Origin' header is present"`

### The Root Cause
The ESP device firmware doesn't include the required CORS response headers:
- `Access-Control-Allow-Origin`
- `Access-Control-Allow-Methods` 
- `Access-Control-Allow-Headers`

This is a server-side issue where the older ESP device is not able to allow cross-origin requests from your web interface

So basically : using the mDNS feature and accessing the swarm page via the nice URL breaks the swarm page as it cannot connect to "older" bitaxes. That might be a problem until those devices receive a firmware upgrade.

Maybe you know a better way ?

mutatrum:
Thanks for the extensive write-up. I think it's fine, we can't expect older devices to magically do-the-right-thing :tm: . For single users, this is a nice improvement as there's a lot of people who run into issues if their local setup already works for hostnames, but none of the calls (f.e. firmware updates) work. For swarm users, they can start using it whenever they have upgraded their devices.

mutatrum:
One addition (don't know if that's not already the case): if a mDNS device scans and find a Bitaxe device which also has mDNS, to query it on it's hostname instead of IP. Basically, at some point, we could get rid of the ip column in swarm, or migrate them to a single column.

camalolo:
Swarm Enhancement: Support for IPs and mDNS Hostnames
I've pushed a commit that updates the swarm management page to handle both IP addresses and mDNS hostnames (e.g., bitaxe.local). Here's what changed:

Frontend Changes (Angular):

Renamed all "IP" references to "address" for better clarity and flexibility.
Updated the manual add form to accept hostnames with validation (supports .local domains).
Enhanced sorting logic: IPs sort numerically, hostnames alphabetically, with IPs prioritized.
Added user-friendly error handling for older devices that may not be compatible.

Backend Changes :

Added the device's current IP address to the /api/system/info JSON response to enable subnet scanning when accessing via hostname. Required to be able to scan the LAN as the browser does not seem to be able to resolve the IP of the current page when accessing it over a .local domain

Key Benefits:

Users can now add swarm peers using either IP (e.g., 192.168.1.100) or hostname (e.g., bitaxe.local).
Network scanning works seamlessly regardless of access method.
Backward compatible with existing IP-based setups.

The build passes successfully, and all changes maintain existing functionality while expanding support. Ready for review and testing!

camalolo:
**NB** I wrote "backwards compatible" but in fact it's not really correct - older firmware block calling their API from a mDNS domain because of CORS

components/connect/connect.c
```diff
@@ -389,6 +390,45 @@ void wifi_init(void * pvParameters)
...
+            if (err != ESP_OK) {
+                ESP_LOGW(TAG, "mDNS hostname setup failed: %s", esp_err_to_name(err));
+                ESP_LOGW(TAG, "Device hostname not set for mDNS discovery");
+            } else {
+                ESP_LOGI(TAG, "mDNS hostname set to: %s.local", hostname);
+                ESP_LOGI(TAG, "Access device at: http://%s.local", hostname);
+            }
+
+            /* Set mDNS instance name */
+            err = mdns_instance_name_set("ESP-Miner Web Server");
```
mutatrum: Is this instance name exposed somewhere, or is it internal only? If exposed, maybe we can give it a bit nicer name. Maybe device model, f.e. `Bitaxe Gamma` from DeviceConfig?

mutatrum:
Testing it now. Opening swarm gave a ton of these messages:
image
I think that's not ok. It found itself, but had it's IP as address, not sure if that's desired:
image

It popped up in `avahi-browse` after boot, which is nice:
```
mutatrum@laptop:~$ avahi-browse _http._tcp
+ enx5c857e30fd32 IPv4 ESP32 WebServer                               Web Site             local
```
But would indeed make the description reflect the device.

mutatrum:
Changing settings while accessing it on hostname works, but it gave a log line:
```
W (331374) CORS: Invalid IP address: bitaxe-gamma
```

I didn't access with `hostname.local`, just `hostname`. Not sure what happened after that, i tried accessing on `hostname.localhost`, and that crashed the device. On restart, the watchdog was triggered:
```
I (6524) connect: IP Address: 192.168.1.20
I (6524) connect: Connected to SSID: *****
I (6525) wifi:mode : sta (d8:3b:da:6e:cd:cc)
I (6529) esp_netif_handlers: sta ip: 192.168.1.20, mask: 255.255.255.0, gw: 192.168.1.1
I (6536) connect: Configuration Access Point disabled
E (7448) task_wdt: Task watchdog got triggered. The following tasks/users did not reset the watchdog in time:
E (7448) task_wdt:  - IDLE0 (CPU 0)
E (7448) task_wdt: Tasks currently running:
E (7448) task_wdt: CPU 0: main
E (7448) task_wdt: CPU 1: IDLE1
E (7448) task_wdt: Print CPU 0 (current core) backtrace


Backtrace: 0x42052DB3:0x3FC9EEF0 0x40377309:0x3FC9EF20 0x4202C531:0x3FCC0060 0x42027D82:0x3FCC00A0 0x42027DCD:0x3FCC00D0 0x4203B64E:0x3FCC0100 0x4203B689:0x3FCC01C0 0x420DAEC5:0x3FCC02B0 0x42027450:0x3FCC02D0 0x4202A37A:0x3FCC0320 0x4202AB25:0x3FCC0360 0x4202B2CA:0x3FCC0390 0x420278AB:0x3FCC03C0 0x42027987:0x3FCC03F0 0x42019A78:0x3FCC0410 0x4200E769:0x3FCC0430 0x4200E347:0x3FCC0460 0x420E65EF:0x3FCC0490
--- 0x42052db3: task_wdt_timeout_handling at /home/mutatrum/esp/v5.5.1/esp-idf/components/esp_system/task_wdt/task_wdt.c:436
--- (inlined by) task_wdt_isr at /home/mutatrum/esp/v5.5.1/esp-idf/components/esp_system/task_wdt/task_wdt.c:509
--- 0x40377309: _xt_lowint1 at /home/mutatrum/esp/v5.5.1/esp-idf/components/xtensa/xtensa_vectors.S:1240
--- 0x4202c531: lv_inv_area at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_refr.c:277
--- 0x42027d82: lv_obj_invalidate_area at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_obj_pos.c:847
--- 0x42027dcd: lv_obj_invalidate_area at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_obj_pos.c:824
--- (inlined by) lv_obj_invalidate at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_obj_pos.c:863
```
I'll flash back to master, to see if it's something on my network. I'll test further this evening.

Addition: reflashed master and device boots again. Not sure what was happening here. Will continue.

camalolo:
Thanks for the feedback.

Yeah I also was hesitating about the "potential peer" warnings. It's an issue with CORS and XHR, on my LAN I have a bunch of devices that can be reached (TVs, routers, etc), but the swarm page cannot really get accurate HTTP status (like a 404 for most devices, or 401 for older bitaxes). Initially I wanted to warn the user they should update the firmware on the other bitaxe(s) but it would make sense they upgrade all of the devices more or less at the same time anyways. I think I can silently skip unresponsive devices without warning.

Agree with you about the mDNS instance name, using the device model is a good idea, will implement.

I will also try to play around with editing the hostname, to see if I can get the same crash as you did. Could it be the length of the hostname, or the "-" ?

mutatrum:
Note on the crash: I had `avahi-browse _http._tcp` running at that moment, not sure if that has anything to do with it?

camalolo:
> I (6524) connect: IP Address: 192.168.1.20
> I (6524) connect: Connected to SSID: *****
> I (6525) wifi:mode : sta (d8:3b:da:6e:cd:cc)
> I (6529) esp_netif_handlers: sta ip: 192.168.1.20, mask: 255.255.255.0, gw: 192.168.1.1
> I (6536) connect: Configuration Access Point disabled
> E (7448) task_wdt: Task watchdog got triggered. The following tasks/users did not reset the watchdog in time:
> E (7448) task_wdt:  - IDLE0 (CPU 0)
> E (7448) task_wdt: Tasks currently running:
> E (7448) task_wdt: CPU 0: main
> E (7448) task_wdt: CPU 1: IDLE1
> E (7448) task_wdt: Print CPU 0 (current core) backtrace
> 
> 
> Backtrace: 0x42052DB3:0x3FC9EEF0 0x40377309:0x3FC9EF20 0x4202C531:0x3FCC0060 0x42027D82:0x3FCC00A0 0x42027DCD:0x3FCC00D0 0x4203B64E:0x3FCC0100 0x4203B689:0x3FCC01C0 0x420DAEC5:0x3FCC02B0 0x42027450:0x3FCC02D0 0x4202A37A:0x3FCC0320 0x4202AB25:0x3FCC0360 0x4202B2CA:0x3FCC0390 0x420278AB:0x3FCC03C0 0x42027987:0x3FCC03F0 0x42019A78:0x3FCC0410 0x4200E769:0x3FCC0430 0x4200E347:0x3FCC0460 0x420E65EF:0x3FCC0490
> --- 0x42052db3: task_wdt_timeout_handling at /home/mutatrum/esp/v5.5.1/esp-idf/components/esp_system/task_wdt/task_wdt.c:436
> --- (inlined by) task_wdt_isr at /home/mutatrum/esp/v5.5.1/esp-idf/components/esp_system/task_wdt/task_wdt.c:509
> --- 0x40377309: _xt_lowint1 at /home/mutatrum/esp/v5.5.1/esp-idf/components/xtensa/xtensa_vectors.S:1240
> --- 0x4202c531: lv_inv_area at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_refr.c:277
> --- 0x42027d82: lv_obj_invalidate_area at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_obj_pos.c:847
> --- 0x42027dcd: lv_obj_invalidate_area at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_obj_pos.c:824
> --- (inlined by) lv_obj_invalidate at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_obj_pos.c:863

Based on the backtrace of the crash, it seems to be related to the screen animation, not the network. But please don't take my word for it, I am not super familiar with any of the code to be honest, and that part is definitely above my head :)

camalolo:
> > I (6524) connect: IP Address: 192.168.1.20
> > I (6524) connect: Connected to SSID: *****
> > I (6525) wifi:mode : sta (d8:3b:da:6e:cd:cc)
> > I (6529) esp_netif_handlers: sta ip: 192.168.1.20, mask: 255.255.255.0, gw: 192.168.1.1
> > I (6536) connect: Configuration Access Point disabled
> > E (7448) task_wdt: Task watchdog got triggered. The following tasks/users did not reset the watchdog in time:
> > E (7448) task_wdt:  - IDLE0 (CPU 0)
> > E (7448) task_wdt: Tasks currently running:
> > E (7448) task_wdt: CPU 0: main
> > E (7448) task_wdt: CPU 1: IDLE1
> > E (7448) task_wdt: Print CPU 0 (current core) backtrace
> > Backtrace: 0x42052DB3:0x3FC9EEF0 0x40377309:0x3FC9EF20 0x4202C531:0x3FCC0060 0x42027D82:0x3FCC00A0 0x42027DCD:0x3FCC00D0 0x4203B64E:0x3FCC0100 0x4203B689:0x3FCC01C0 0x420DAEC5:0x3FCC02B0 0x42027450:0x3FCC02D0 0x4202A37A:0x3FCC0320 0x4202AB25:0x3FCC0360 0x4202B2CA:0x3FCC0390 0x420278AB:0x3FCC03C0 0x42027987:0x3FCC03F0 0x42019A78:0x3FCC0410 0x4200E769:0x3FCC0430 0x4200E347:0x3FCC0460 0x420E65EF:0x3FCC0490
> > --- 0x42052db3: task_wdt_timeout_handling at /home/mutatrum/esp/v5.5.1/esp-idf/components/esp_system/task_wdt/task_wdt.c:436
> > --- (inlined by) task_wdt_isr at /home/mutatrum/esp/v5.5.1/esp-idf/components/esp_system/task_wdt/task_wdt.c:509
> > --- 0x40377309: _xt_lowint1 at /home/mutatrum/esp/v5.5.1/esp-idf/components/xtensa/xtensa_vectors.S:1240
> > --- 0x4202c531: lv_inv_area at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_refr.c:277
> > --- 0x42027d82: lv_obj_invalidate_area at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_obj_pos.c:847
> > --- 0x42027dcd: lv_obj_invalidate_area at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_obj_pos.c:824
> > --- (inlined by) lv_obj_invalidate at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_obj_pos.c:863
> 
> Based on the backtrace of the crash, it seems to be related to the screen animation, not the network. But please don't take my word for it, I am not super familiar with any of the code to be honest, and that part is definitely above my head :)

One possibility : the mDNS feature somehow freezes the CPU long enough for the display code to fail. Maybe I should make sure the mDNS code runs a bit later after boot and/or asynchronously

camalolo:
> > > I (6524) connect: IP Address: 192.168.1.20
> > > I (6524) connect: Connected to SSID: *****
> > > I (6525) wifi:mode : sta (d8:3b:da:6e:cd:cc)
> > > I (6529) esp_netif_handlers: sta ip: 192.168.1.20, mask: 255.255.255.0, gw: 192.168.1.1
> > > I (6536) connect: Configuration Access Point disabled
> > > E (7448) task_wdt: Task watchdog got triggered. The following tasks/users did not reset the watchdog in time:
> > > E (7448) task_wdt:  - IDLE0 (CPU 0)
> > > E (7448) task_wdt: Tasks currently running:
> > > E (7448) task_wdt: CPU 0: main
> > > E (7448) task_wdt: CPU 1: IDLE1
> > > E (7448) task_wdt: Print CPU 0 (current core) backtrace
> > > Backtrace: 0x42052DB3:0x3FC9EEF0 0x40377309:0x3FC9EF20 0x4202C531:0x3FCC0060 0x42027D82:0x3FCC00A0 0x42027DCD:0x3FCC00D0 0x4203B64E:0x3FCC0100 0x4203B689:0x3FCC01C0 0x420DAEC5:0x3FCC02B0 0x42027450:0x3FCC02D0 0x4202A37A:0x3FCC0320 0x4202AB25:0x3FCC0360 0x4202B2CA:0x3FCC0390 0x420278AB:0x3FCC03C0 0x42027987:0x3FCC03F0 0x42019A78:0x3FCC0410 0x4200E769:0x3FCC0430 0x4200E347:0x3FCC0460 0x420E65EF:0x3FCC0490
> > > --- 0x42052db3: task_wdt_timeout_handling at /home/mutatrum/esp/v5.5.1/esp-idf/components/esp_system/task_wdt/task_wdt.c:436
> > > --- (inlined by) task_wdt_isr at /home/mutatrum/esp/v5.5.1/esp-idf/components/esp_system/task_wdt/task_wdt.c:509
> > > --- 0x40377309: _xt_lowint1 at /home/mutatrum/esp/v5.5.1/esp-idf/components/xtensa/xtensa_vectors.S:1240
> > > --- 0x4202c531: lv_inv_area at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_refr.c:277
> > > --- 0x42027d82: lv_obj_invalidate_area at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_obj_pos.c:847
> > > --- 0x42027dcd: lv_obj_invalidate_area at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_obj_pos.c:824
> > > --- (inlined by) lv_obj_invalidate at /home/mutatrum/code-workspace/esp-miner/managed_components/lvgl__lvgl/src/core/lv_obj_pos.c:863
> > 
> > 
> > Based on the backtrace of the crash, it seems to be related to the screen animation, not the network. But please don't take my word for it, I am not super familiar with any of the code to be honest, and that part is definitely above my head :)
> 
> One possibility : the mDNS feature somehow freezes the CPU long enough for the display code to fail. Maybe I should make sure the mDNS code runs a bit later after boot and/or asynchronously

Trying this : moving the mDNS configuration to a later stage in the boot process

camalolo:
Added a new commit :

- Moved the mDNS setup after an IP has been received (tentative fix for possible timing issues in display animation)
- Made the service show up as Bitaxe <Model> <Board> (MAC first 4 digits) for easy identification

camalolo:
Commit added to the swarm page so it fails silently for irrelevant/old devices during automatic scans

WantClue:
@camalolo can you please resolve the conflicts ?

camalolo:
@WantClue hello ! Thanks for bringing this to my attention, I will have a look !

camalolo:
@WantClue 

I updated my branch. Here's a summary of what the commit does :

Core mDNS Implementation

- Added mDNS dependency: Integrated espressif/mdns^1.8.0 component via CMakeLists.txt and CI workflow updates, ensuring reliable network service discovery.
- mDNS initialization: In connect.c, added initialize_mdns() function that:
  - Sets device hostname (e.g., bitaxe.local) for network discovery
  - Registers HTTP service on port 80 with descriptive instance names (e.g., "Bitaxe Ultra 2.1 (ABCD)")
  - Handles IPv4/IPv6 connectivity events to ensure mDNS starts when network is available
  - Includes fallback logging for troubleshooting discovery issues
- Hostname management: Added update_mdns_hostname() for dynamic hostname updates and normalize_hostname() in HTTP server to prevent ".local.local" duplicates.

Swarm Component Refactoring (Address-Based Support)

- Unified address handling: Replaced IP-only logic with flexible "address" support (accepts both IPv4 addresses and .local hostnames)
- Enhanced network scanning:
  - Detects access method (direct IP vs. mDNS hostname) and adapts scanning strategy
  - Fetches server IP via API when accessed via hostname, then scans subnet
  - Improved duplicate detection using hostname/IP cross-checking
- UI updates: 
  - Form validation now accepts both IP patterns and .local hostnames
  - Table headers and links updated from "IP" to "Address"
  - Sorting logic enhanced to handle mixed IP/hostname sorting (IPs first, then alphabetical)
- API consistency: All HTTP calls, error handlers, and device references now use address property instead of IP

HTTP Server Enhancements

- CORS improvements: Extended is_network_allowed() to permit requests from mDNS hostnames (e.g., http://bitaxe.local), enabling cross-origin swarm management
- Hostname normalization: Prevents configuration issues by stripping .local suffixes when setting hostnames
Infrastructure Updates

Others

- Build system: Added mDNS component to connect module dependencies
- CI/CD: Updated unittest workflow to install mDNS dependency for testing
- Documentation: README updated with mDNS discovery instructions

I tested locally and it seemed to work

However the changes in master since my last push have been extensive and changed a lot of files I had modified, so I cannot emphasize enough, **please do test thoroughly if you plan to merge this** - I welcome any feedback and I'll fix problems if you are kind enough to report them. There's a real possibility I overlooked something and I'd be grateful for your time testing it.

mutatrum:
Thank you for rebasing. I did a brief test, with merged master, and found some small things:
 * I can access my device perfectly fine on hostname alone, without the `.local` suffix, as with any other local device on my network. In the code however, `.local` is added and shown on the dashboard and swarm. Somehow I can't really find where that is re-added if I try to remove it from the settings.
 * When clicking on a host in swarm, it's opened by IP address.
 * When updating the hostname, I got a warning in the logs: `W (506261) CORS: IP address string is too long: bitaxe-gamma-601`

mutatrum:
Ah, found it in https://github.com/bitaxeorg/ESP-Miner/pull/1240/changes#diff-b92a1982bb82df47ec29d7c8b425023f334ec820176646b59970905c61ebc5e4R869.

As `.local` is not a universal standard, maybe we should let people use the suffix they need, or no suffix at all. I used to have a router that had `.home` as suffix, and as said above, my current LAN works without suffix.

I tested with line 866-881 in `http_server.c` commented out, so it just exposes hostname to the API, and it seems to work fine. API works, it shows just hostname without suffix in the top left and on swarm.

camalolo:
Hello @mutatrum - thanks for taking the time to test and give feedback, I will have a look !

On my LAN I can also find hosts via the hostname only, but I think it's because of netbios resolution (so mostly windows PCs). You can see linux machines without a .local domain ? Any chance it's because you have samba installed with netbios enabled ?

It could also be your router that automatically adds them to its dns server.

As far as I understand, mDNS standards require .local : https://datatracker.ietf.org/doc/html/rfc6762

Also noted your other findings, will work on those as well

mutatrum:
On my Linux machine I can see almost anything without `.local`: other Linux machines, Windows laptop, android tv, phones, printer, sonos, routers.

So maybe the mDNS announcement should indeed have `.local` but it might be possible to not show that, so it's only an internal thing? Depending if the front-end is opened on `.local` or hostname only, its seamless either way.

I haven't tested this with multiple mDNS Bitaxes, will do that as well.

mutatrum:
Oh, by default `hostname` is set to `bitaxe` through the config. What happens if the hostname is already in use and there's a conflict? And same question for when changing the hostname to something that's already in use, how is that handled?

camalolo:
> Oh, by default `hostname` is set to `bitaxe` through the config. What happens if the hostname is already in use and there's a conflict? And same question for when changing the hostname to something that's already in use, how is that handled?

Excellent points, will try and solve these as well

camalolo:
Hello @mutatrum ! I updated the code to take into account your remarks. For now, the mDNS will still use the .local suffix, as I couldn't find any evidence that this is not the standard.

But I believe I handled possible conflicts (two devices called bitaxe for example), fixed the CORS problem, allowed browsing the device via a basic hostname (i.e "bitaxe") if that resolves on your LAN, and improved the flow when you change the hostname via the network settings page.

I tried to test all situations I could think of, however, as usual, a fresh pair of eyes might uncover issues I did not encounter or think of.

Thanks for taking the time to test it !

mutatrum:
Hi @camalolo, I'm working on a Early Access release build and would like to include this PR as well. Can you update the PR to latest in master?

camalolo:
Hello @mutatrum - glad to hear from you ! I will certainly update the PR to the latest master - please give me a few days !

camalolo:
Dear @mutatrum - hopefully the rebase is fine. I did test briefly on my side and it appears to compile and run as before. Please double check everything, as there were 50 commits to master since my last rebase ! Thanks for your time

mutatrum:
Small bug on swarm: if I access the device without `.local`, the swarm page doesn't recognize itself:

image

It should show a :star: with tooltip:
image

Not sure if these two ways of accessing the dashboard can be made transparent.

Will continue testing. I tried merging into `early-access` but that is a bit tricky still.

camalolo:
Oh great, a bug report :) Will fix asap !

mutatrum:
Thanks! Also, looking at the swarm screenshot, not sure if `Hostname` and `Address` should be 2 columns. When do they differ (apart from the suffix)?

camalolo:
> Small bug on swarm: if I access the device without `.local`, the swarm page doesn't recognize itself:

I believe it is now fixed in commit d8e7e3c596487cb413f6a4ce789fd97edc2b55d7 - please check !

camalolo:
> Thanks! Also, looking at the swarm screenshot, not sure if `Hostname` and `Address` should be 2 columns. When do they differ (apart from the suffix)?

Yeah it's true it feels redundant. One was showing the configured hostname and the other the mDNS name, but since the latter relies on the former, it looks like a repetition. Made it a bit better in 67a4049c448d46c9604f6cf895e788d57b9bb6cb :

image

main/http_server/http_server.c
```diff
@@ -830,7 +956,25 @@ static esp_err_t GET_system_info(httpd_req_t * req)
...
 
     char * ssid = nvs_config_get_string(NVS_CONFIG_WIFI_SSID);
-    char * hostname = nvs_config_get_string(NVS_CONFIG_HOSTNAME);
+    char * hostname_base = nvs_config_get_string(NVS_CONFIG_HOSTNAME);
+    char mdns_hostname[64] = {0};
+    char full_hostname[64];
+    
+    // Get the base hostname from mDNS if available
+    if (GLOBAL_STATE->SYSTEM_MODULE.is_connected) {
+        esp_err_t mdns_err = mdns_hostname_get(mdns_hostname);
```
mutatrum: The call to `mdns_hostname_get` shouldn't be here, it's called every invocation of the info endpoint. ideally it should get whatever it needs from `GLOBAL_STATE`.

camalolo: Thank you for the feedback. Please check commit 312ee31535d3e2af1d17ed2994c1909ce408e601 at your convenience !

main/http_server/http_server.c
```diff
@@ -884,11 +1021,16 @@ static esp_err_t GET_system_info(httpd_req_t * req)
     cJSON_AddFloatToObject(root, "frequency", frequency);
     cJSON_AddStringToObject(root, "ssid", ssid);
     cJSON_AddStringToObject(root, "macAddr", formattedMac);
-    cJSON_AddStringToObject(root, "hostname", hostname);
+    cJSON_AddStringToObject(root, "hostname", hostname_base);
+    cJSON_AddStringToObject(root, "fullHostname", full_hostname);
+    if (strlen(mdns_hostname) > 0) {
+        cJSON_AddStringToObject(root, "mdnsHostname", mdns_hostname);
```
mutatrum, line 1068: So, regarding this comment on Discord (https://discord.com/channels/1091348375301013615/1272457672225984543/1482832332778897551):
> Updated to early access 2026-03. In swarm I can't navigate to my other Bitaxe, it's using the hostname instead of IP
> the other ones are a mix of versions early-access, 2.13.1, 2.13.0b2 all link to hostname instead of IP

In swarm, maybe it's possible to have the link only point to `mdnsHostname` if that's available, then fall back to `fullHostname` and if that's not available, back to IP? That way on swarm, at least the other devices dashboard can be opened from an mDNS swarm to pre-mDNS versions?

Not 100% of the sequency and which ones to use though. It's a subtle backwards compatibility thing, but I think it's fairly easy to remedy.

camalolo: Fixed in a32588a6a6da23fb510abe0132e0f83a132b1036

I didn't use the mdnsHostname field at all because that's the "bitaxe" part in "bitaxe.local" and I was concerned it might not resolve, so my commit only fallsback to the IP (technically the address at which the remote device was found)

But let me know if I am wrong and adding mdnsHostname as a fallback before the IP makes sense and I'll add it !

Thanks !

.github/workflows/unittest.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         with:
           esp_idf_version: v5.5.2
           target: esp32s3
-          command: GITHUB_ACTIONS="true" idf.py build
```
eandersson, line 23: Why is this needed? Feels like there must be a better way of handling this.

camalolo: @eandersson you might be right ! What do you think of this approach :

1. Create components/connect/idf_component.yml with:
      dependencies:
     espressif/mdns: "^1.8.0"
   2. Revert unittest.yml line 21 to: command: GITHUB_ACTIONS="true" idf.py build

camalolo: I tested it locally, it works. Will push a small commit for this in a bit. My apologies, I never actually tried to compile the test build on my end, and I misunderstood the proper solution for the error I got before from github (Failed to resolve component 'espressif__mdns' required by component 'connect': unknown name.).

Thank you for your time reviewing this !

camalolo: Fixed in c8d3d3f7c0d7ec227fa84683ce83740c08692e95

camalolo:
@mutatrum I merged with recent master changes ! Good luck

components/connect/connect.c
```diff
@@ -343,6 +465,49 @@ static void wifi_softap_off(void)
...
+    snprintf(suffix, sizeof(suffix), "-%02x%02x", mac[4], mac[5]);
+    char *new_hostname = malloc(strlen(base) + strlen(suffix) + 1);
+    strcpy(new_hostname, base);
+    strcat(new_hostname, suffix);
+    return new_hostname;
+}
+
+static char* check_and_resolve_hostname_conflict(const char *hostname, const char *current_ip) {
+    mdns_result_t *results = NULL;
+    esp_err_t err = mdns_query_generic(hostname, NULL, NULL, MDNS_TYPE_A, MDNS_QUERY_MULTICAST, 3000, 1, &results);
```
WantClue: check_and_resolve_hostname_conflict calls mdns_query_generic with a 3000ms timeout from within the WiFi event handler (via initialize_mdns_if_needed). This blocks the system event loop and may cause watchdog timeouts or missed events on slow networks.

Consider moving mDNS initialization to a separate task or using a shorter timeout with async retry.

components/connect/connect.c
```diff
@@ -343,6 +465,49 @@ static void wifi_softap_off(void)
     }
 }
 
+
+
+static char* generate_unique_hostname(const char *base) {
```
WantClue, line 500: generate_unique_hostname allocates memory but doesn't check if malloc succeeded. On memory exhaustion, subsequent string operations will dereference NULL.

WantClue:
```
+    if (new_hostname == NULL) {
+        ESP_LOGE(TAG, "Failed to allocate memory for hostname");
+        return NULL;
+    }
```

WantClue: and then in check_and_resolve_hostname_conflict
If the malloc fix is applied, callers need to handle NULL returns. Additionally, the NVS write here persists a hostname the user didn't explicitly set, dunno about this one yet

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -422,10 +542,10 @@ export class SwarmComponent implements OnInit, OnDestroy {
 
   get sortOptions() {
     return [
-      { label: 'Hostname', value: { sortField: 'hostname', sortDirection: 'desc' } },
-      { label: 'Hostname', value: { sortField: 'hostname', sortDirection: 'asc' } },
-      { label: 'IP', value: { sortField: 'IP', sortDirection: 'desc' } },
-      { label: 'IP', value: { sortField: 'IP', sortDirection: 'asc' } },
+      { label: 'Hostname (Z-A)', value: { sortField: 'hostname', sortDirection: 'desc' } },
```
mutatrum: No need for (Z-A)/(A-Z), there's a visual indicator for asc/dec.

components/connect/connect.c
camalolo: Thanks for your feedback, will fix ASAP

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
camalolo: Thank you 🙏 will fix

mutatrum:
I had a short test of mDNS this afternoon, and the swarm page was broken directly after the upgrade. It looked exactly like the problem in #1617. I also saw javascript errors in the console on null fields, but lost the stacktrace. I think it needs some null checks on the stored swarm information.

components/connect/connect.c
camalolo: Please see 4d869a8255dfd864af81d0a8f464a79b1834f6b5

components/connect/connect.c
camalolo: Please see 8ab859b7cbcf995cf4fb933b37bb136f06335f04

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
camalolo: Please see b4c1b03541aadc7571d693f136944476379db764

camalolo:
> I had a short test of mDNS this afternoon, and the swarm page was broken directly after the upgrade. It looked exactly like the problem in #1617. I also saw javascript errors in the console on null fields, but lost the stacktrace. I think it needs some null checks on the stored swarm information.

Check if 27c2b14c36f54e127ae38f5951d8fe039fb1f040 fixes it !

WantClue:
@camalolo can you please resolve the conflicts

camalolo:
> @camalolo can you please resolve the conflicts

Rebased ! Please check. Thank you !

mutatrum:
> > @camalolo can you please resolve the conflicts
> 
> Rebased ! Please check. Thank you !

That's fantastic! Will do one more round of testing after 2.14 is released. If all goes well and there are no issues, this should go into 2.15.

components/connect/connect.c
```diff
@@ -97,6 +108,144 @@ esp_err_t wifi_apply_hostname(const char *hostname)
...
+        char mac_suffix[6];
+        snprintf(mac_suffix, sizeof(mac_suffix), "%02X%02X", mac[4], mac[5]);
+
+        char instance_name[64];
+        snprintf(instance_name, sizeof(instance_name), "Bitaxe %s %s (%s)",
+                 GLOBAL_STATE->DEVICE_CONFIG.family.name,
+                 GLOBAL_STATE->DEVICE_CONFIG.board_version,
+                 mac_suffix);
+        
+        /* Add HTTP service */
```
frantisekbohacek, line 206: Hey, would you consider also adding a subservice  for the AxeOS api itself? This is a generic http, so it's not saying that this device also supports specific endpoints in a specific format. For this `mdns_service_subtype_add_for_host` may be used. AxeOS could for example advertise as `_axeos._sub._http._tcp`

Then it would be possible to find AxeOS devices on the local network easily through DNS-SD. Right now this would still need an iteration over all http services on the network.

components/connect/connect.c
```diff
@@ -97,6 +108,144 @@ esp_err_t wifi_apply_hostname(const char *hostname)
...
+        err = mdns_service_add(instance_name, "_http", "_tcp", 80, NULL, 0);
```
frantisekbohacek, line 207: I think this is a missed opportunity, have you considered adding TXT records that would tell you information about the device, such as the model, board, asic used, count of asics, firmware version?

components/connect/connect.c
```diff
@@ -97,6 +108,144 @@ esp_err_t wifi_apply_hostname(const char *hostname)
...
+        free(final_hostname);
+
+        /* Set mDNS instance name */
+        uint8_t mac[6];
+        esp_wifi_get_mac(WIFI_IF_STA, mac);
+        char mac_suffix[6];
+        snprintf(mac_suffix, sizeof(mac_suffix), "%02X%02X", mac[4], mac[5]);
+
+        char instance_name[64];
+        snprintf(instance_name, sizeof(instance_name), "Bitaxe %s %s (%s)",
```
frantisekbohacek, line 201: The instance name mentioned in readme.md does not correspond to this instance name.

camalolo: @frantisekbohacek added 3 commits to address your recommendations. Would you please have a look ? Thank you for your time !

frantisekbohacek: Hey, thanks for that, I like the changes, I see nothing problematic with them. Unfortunately I can't really help with testing this, though. So I hope that's fine with you.

frantisekbohacek:
```
> avahi-browse -rt _axeos._sub._http._tcp
+ wlp1s0 IPv6 Bitaxe Gamma 601 (1D1C)                       _http._tcp           local
+ wlp1s0 IPv4 Bitaxe Gamma 601 (1D1C)                       _http._tcp           local
= wlp1s0 IPv6 Bitaxe Gamma 601 (1D1C)                       _http._tcp           local
   hostname = [bitaxe.local]
   address = [10.53.222.86]
   port = [80]
   txt = ["board=601" "family=Gamma" "asic=BM1370" "asic_count=1" "fw_version=v2.14.0b4-14-g397eabb-dirty"]
= wlp1s0 IPv4 Bitaxe Gamma 601 (1D1C)                       _http._tcp           local
   hostname = [bitaxe.local]
   address = [10.53.222.86]
   port = [80]
   txt = ["board=601" "family=Gamma" "asic=BM1370" "asic_count=1" "fw_version=v2.14.0b4-14-g397eabb-dirty"]
```

change of plans, I could test this :) Works fine for me!

camalolo: Thanks for spending the time !

WantClue:
rereview please from @eandersson @mutatrum @WantClue

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -114,6 +120,39 @@ export class SwarmComponent implements OnInit, OnDestroy {
         }
       }
     }, 1000);
+
+    this.httpClient.get(`http://${window.location.hostname}/api/system/info`).subscribe({
+      next: (response: any) => {
+        this.currentDeviceIp = response.ip;
```
mutatrum: The `ip` field does not exist. It's either ipv4 or ipv6. This happens several times in this file and once in the html.

components/connect/connect.c
```diff
@@ -97,6 +109,174 @@ esp_err_t wifi_apply_hostname(const char *hostname)
...
+
+esp_err_t update_mdns_hostname(const char *new_hostname, GlobalState *GLOBAL_STATE) {
+    if (new_hostname == NULL || strlen(new_hostname) == 0) {
+        ESP_LOGW(TAG, "Invalid hostname provided for mDNS update");
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    /* Get current IP for conflict checking */
+    esp_netif_ip_info_t ip_info;
+    esp_netif_get_ip_info(esp_netif_get_handle_from_ifkey("WIFI_STA_DEF"), &ip_info);
```
mutatrum: Check ip_info on NULL, as this might fail on AP mode. If null, it should try on "WIFI_AP_DEF", but that could also still be NULL which would crash on line 249.

main/http_server/http_server.c
```diff
@@ -924,6 +1052,25 @@ static esp_err_t GET_system_info(httpd_req_t * req)
...
+        snprintf(full_hostname, sizeof(full_hostname), "%s.local", mdns_hostname);
+    } else {
+        snprintf(full_hostname, sizeof(full_hostname), "%s.local", hostname_base ? hostname_base : "unknown");
+    }
+
+    cJSON_AddStringToObject(root, "fullHostname", full_hostname);
+    if (strlen(mdns_hostname) > 0) {
+        cJSON_AddStringToObject(root, "mdnsHostname", mdns_hostname);
+    }
+    cJSON_AddStringToObject(root, "ipv6", GLOBAL_STATE->SYSTEM_MODULE.ipv6_addr_str);
```
mutatrum: The field `ipv6` is already added in system_api_json.c, system_api_add_config.

main/http_server/http_server.c
```diff
@@ -739,9 +849,27 @@ static esp_err_t PATCH_update_settings(httpd_req_t * req)
         }
     }
 
-    cJSON_Delete(root);
-    httpd_resp_send_chunk(req, NULL, 0);
-    return ESP_OK;
+    // Create response JSON
+    cJSON *response = cJSON_CreateObject();
```
mutatrum, line 853: Response schema is missing in `openapi.yaml`

main/http_server/http_server.c
```diff
@@ -328,9 +355,42 @@ esp_err_t is_network_allowed(httpd_req_t * req)
...
+    }
+
+    // Check if Origin header matches the avahi hostname    
+    if (httpd_req_get_hdr_value_len(req, "Origin") > 0) {
+        httpd_req_get_hdr_value_str(req, "Origin", origin, sizeof(origin));
+        ESP_LOGD(CORS_TAG, "Origin header: %s", origin);
+        char *hostname = nvs_config_get_string(NVS_CONFIG_HOSTNAME);
+        ESP_LOGD(CORS_TAG, "Configured hostname: %s", hostname);
+        char expected[256];
+        snprintf(expected, sizeof(expected), "http://%s.local", hostname);
```
mutatrum: This should support port number as well, for dev servers which don't run on port 80

mutatrum (commented):
Some review comments. Will run tests later this week. Exciting to get closer and closer and that we can finally get rid if access on IP addresses.

camalolo:
Thank you @mutatrum I appreciate you spending time on checking the code, I will fix and test more thoroughly

camalolo:
Hello @mutatrum 

All your comments were spot on

I rebased to get the latest master commits, and fixed the bugs you found in 483047854d0e712730ea42f9e33e3d4dfddf2cd8

I also performed a few automated tests to revalidate all the features in this pull request

1                        │  596bcef                 │ mDNS init &  _http._tcp  service       │ ✅
 2                        │  7918d9a                 │ Swarm self-identification (star icon)  │ ✅
 3                        │  5b7fb99                 │ Swarm device link + IP display         │ ✅
 4                        │  d0cd954                 │ Hostname in global state / API         │ ✅
 5                        │  861d707                 │ Fallback to IP for older devices       │ ✅
 6                        │  c678883                 │ mDNS dependency in component           │ ✅
 7                        │  251beec                 │ mDNS init in separate task             │ ✅
 8                        │  dea4684                 │ NULL checks in hostname conflict       │ ✅
 9                        │  8de80e3                 │ Clean sort labels                      │ ✅
 10                       │  666a5bd                 │ Cache invalidation + null safety       │ ✅
 11                       │  cdc1114                 │ Buffer overflow protection             │ ✅
 12                       │  4fe2955                 │ AxeOS subtype  _axeos._sub._http._tcp  │ ✅
 13                       │  a146eba                 │ DNS-SD TXT records                     │ ✅
 14                       │  bb57780                 │ Readme docs                            │ ✅
 15                       │  4830478                 │ CORS, OpenAPI, swarm validation        │ ✅

Thank you very much for your time

components/connect/connect.c
```diff
@@ -97,6 +109,181 @@ esp_err_t wifi_apply_hostname(const char *hostname)
...
+    esp_netif_t *netif = esp_netif_get_handle_from_ifkey("WIFI_STA_DEF");
+    if (netif == NULL) {
+        netif = esp_netif_get_handle_from_ifkey("WIFI_AP_DEF");
+    }
+    if (netif == NULL || esp_netif_get_ip_info(netif, &ip_info) != ESP_OK) {
+        ESP_LOGW(TAG, "No active network interface for mDNS hostname update");
+        ip_info.ip.addr = 0;
+    }
+    char current_ip[16];
+    snprintf(current_ip, sizeof(current_ip), IPSTR, IP2STR(&ip_info.ip));
```
mutatrum: Missing null check here, `ip_info` can still be 0. If this happens, maybe default `current_ip` to `0.0.0.0`?

mutatrum:
I've tested this with two devices, and noticed some issues with swarm:
 * For visual clarity, bare hostname (no `.local`) should be sufficient on the overview. However, the link backing it, should follow the current devices access method. So based on how the dashboard is opened, it should follow the access pattern. Either IP, hostname or hostname.local;
 * The scan doesn't work, as it's hitting against CORS. CORS should allow for hostnames without any `.` or the `.local` suffix. If I'm not mistaken, this should hold the CORS protection in place.

Last one surprises me, I though CORS was already fixed for hostname access, or am I testing wrong? But even if I manually try to add the other device, it's blocked by CORS.

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -67,7 +81,7 @@ export class SwarmComponent implements OnInit, OnDestroy {
   ) {
 
     this.form = this.fb.group({
-      manualAddIp: [null, [Validators.required, Validators.pattern('(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)')]]
+      manualAddAddress: [null, [Validators.required, Validators.pattern('^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)|[a-zA-Z0-9-]+(?:\\.local)?$')]]
```
mutatrum: This is a bit of an unwieldy regex. Maybe refactor this into a concrete method?
``` ts
import { AbstractControl, ValidationErrors } from '@angular/forms';
export function addressValidator(control: AbstractControl): ValidationErrors | null {
  const value = control.value;
  if (!value) return null;
  const parts = value.split('.');
  switch (parts.length) {
    case 1: // Bare hostname (e.g. "bitaxe")
      return null;
    case 2: // mDNS hostname (e.g. "bitaxe.local")
      if (parts[1].toLowerCase() === 'local') return null;
      break;
    case 4: // IP Address (e.g. "192.168.1.1")
      if (parts.every(part => !isNaN(Number(part)) && part.trim() !== '')) return null;
      break;
  }  return { invalidAddress: true };
}
```

camalolo:
Hello @mutatrum - added 4 commits to address your concerns, and did some basic testing. Please tell me if it all works fine for you

mutatrum:
> Hello @mutatrum - added 4 commits to address your concerns, and did some basic testing. Please tell me if it all works fine for you

image

Directly after scan, works like a charm! I have only one small nit, see below.

Also, I noticed this (unparsed JSON, instead of just the message), but not sure if this is from this PR or if that was already the case:

image

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -122,10 +178,44 @@ export class SwarmComponent implements OnInit, OnDestroy {
...
+  // Utility method to get the display name for a device
+  public getDeviceDisplayName(device: SwarmDevice): string {
+    return device.displayName || device.address;
+  }
+
+  // Utility method to get the link URL for a device
+  // Follows the current device's access method (IP, hostname.local, or bare hostname)
+  public getDeviceLink(device: SwarmDevice): string {
+    const currentHost = window.location.hostname;
+    const isIP = /^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$/.test(currentHost);
```
mutatrum: use `isIpAddress` from line 192?

mutatrum:
Maybe as a follow-up PR is to leverage the DNS-SD TXT records for discovery, instead of the IP range scan. But that's a bit of work as it needs to go through the backend.

mutatrum (approved):
LGTF, congrats 🚀  

Only one last nit 👍

bonifacio123:
Have tested this on a Duo and GT. From the CLI and I am able to ping using the host name (ping duo). Also tested access from the browser (http://duo) and the page loads right up. In Swarm I see the IP addresses for the Bitaxes - for QAxe's it's missing.

components/connect/connect.c
```diff
@@ -380,6 +575,57 @@ static void wifi_softap_off(void)
...
 }
 
+
+
+static char* generate_unique_hostname(const char *base) {
+    uint8_t mac[6];
+    esp_wifi_get_mac(WIFI_IF_STA, mac);
+    char suffix[6];
+    snprintf(suffix, sizeof(suffix), "-%02x%02x", mac[4], mac[5]);
+    char *new_hostname = malloc(strlen(base) + strlen(suffix) + 1);
```
WantClue, line 599: The generate_unique_hostname() function appends a 5-character MAC-based suffix (-XXYY) to the base hostname without checking the total length. If the base hostname is already at or near the DNS maximum of 63 characters, the resulting hostname will exceed this limit (up to 68 characters), causing mDNS registration failures and potential DNS resolution issues.

components/connect/connect.c
```diff
@@ -380,6 +575,57 @@ static void wifi_softap_off(void)
...
+
+    if (strcmp(ip_str, current_ip) != 0) {
+        char *new_hostname = generate_unique_hostname(hostname);
+        if (new_hostname == NULL) {
+            ESP_LOGW(TAG, "mDNS conflict detected for '%s' but could not generate unique name, keeping original", hostname);
+            mdns_query_results_free(results);
+            return strdup(hostname);
+        }
+        ESP_LOGI(TAG, "mDNS conflict detected for '%s' at %s, renaming to '%s'", hostname, ip_str, new_hostname);
+        nvs_config_set_string(NVS_CONFIG_HOSTNAME, new_hostname);
```
WantClue: The check_and_resolve_hostname_conflict() function directly writes the resolved hostname to NVS on line 620. However, the caller update_mdns_hostname() also writes to NVS on line 271 when it detects the hostname changed. This results in two consecutive NVS writes when a conflict is resolved, which is inefficient and increases flash wear.

main/http_server/http_server.c
```diff
@@ -880,6 +1052,24 @@ static esp_err_t GET_system_info(httpd_req_t * req)
...
+    char full_hostname[72]; // 64 (mdns_hostname max) + 6 (".local") + 2 (margin)
+    
+    if (mdns_hostname != NULL && strlen(mdns_hostname) > 0) {
+        snprintf(full_hostname, sizeof(full_hostname), "%s.local", mdns_hostname);
+    } else {
+        snprintf(full_hostname, sizeof(full_hostname), "%s.local", hostname_base ? hostname_base : "unknown");
+    }
+
+    cJSON_AddStringToObject(root, "fullHostname", full_hostname);
+    if (strlen(mdns_hostname) > 0) {
```
WantClue, line 1067: Line 1067 calls strlen(mdns_hostname) outside the null-check block that starts on line 1060. If mdns_hostname is NULL (which is possible per line 1057), this will cause a segmentation fault.

replace: if (strlen(mdns_hostname) > 0) {

with: if (mdns_hostname != NULL && strlen(mdns_hostname) > 0) {

WantClue (changes requested):

bonifacio123:
> > Yes, I'm able to connect to the QAxe's via the host name (http://qaxe4)
> 
> Can you open the dashboard of the Bitaxe with `hostname.local`? And if you do that, will the QAxe dashboard also open on `qaxe4.local`?

Yes, both Bitaxe and QAxe open with .local; i.e.   http://duo.local

bonifacio123:
> > > Yes, I'm able to connect to the QAxe's via the host name (http://qaxe4)
> > 
> > 
> > Can you open the dashboard of the Bitaxe with `hostname.local`? And if you do that, will the QAxe dashboard also open on `qaxe4.local`?
> 
> Yes, both Bitaxe and QAxe open with .local; i.e. http://duo.local

Clarification, .local is required. My router dhcp had the same host names, duo and qaxe6, I changed the host names in the miners to something new and rebooted. They only respond to   http://duo.local  and  http://qaxe6.local

camalolo:
Thanks a lot for the extensive feedback. Added 2 commits to address some concerns.

Some things might not be a problem though :

1. connect.c#L585 https://github.com/bitaxeorg/ESP-Miner/blob/8b995f9/components/connect/connect.c#L585 — Hostname length overflow: The NVS config caps hostname at 32 chars (https://github.com/bitaxeorg/ESP-Miner/blob/8b995f9/main/nvs_config.c#L53). The suffix is 5 chars ( -XXYY ). Total max = 37, well under the DNS 63-char
limit.

2. https://github.com/bitaxeorg/ESP-Miner/blob/8b995f9/main/http_server/http_server.c#L1067 — NULL deref on  mdns_hostname :  mdns_hostname  is declared as  char mdns_hostname[64]  in the  GlobalState  struct (https://github.com/bitaxeorg/ESP-Miner/blob/8b995f9/main/global_state.h#L116) — a fixed array, not a pointer. It can never be NULL. The  strlen()  call is always safe.

Let me know if I am mistaken, happy to fix those if needed

mutatrum:
> Have tested this on a Duo and GT. From the CLI and I am able to ping using the host name (ping duo). Also tested access from the browser (http://duo) and the page loads right up. In Swarm I see the IP addresses for the Bitaxes - for QAxe's it's missing.

Opened issue #1769 for the missing IP from QAxe. I deleted comments on this to keep this about mDNS. @bonifacio123 if you want to add anythere there, feel free.

WantClue:
> Thanks a lot for the extensive feedback. Added 2 commits to address some concerns.
> 
> Some things might not be a problem though :
> 
> 1. connect.c#L585 https://github.com/bitaxeorg/ESP-Miner/blob/8b995f9/components/connect/connect.c#L585 — Hostname length overflow: The NVS config caps hostname at 32 chars (https://github.com/bitaxeorg/ESP-Miner/blob/8b995f9/main/nvs_config.c#L53). The suffix is 5 chars ( -XXYY ). Total max = 37, well under the DNS 63-char
>    limit.
> 2. https://github.com/bitaxeorg/ESP-Miner/blob/8b995f9/main/http_server/http_server.c#L1067 — NULL deref on  mdns_hostname :  mdns_hostname  is declared as  char mdns_hostname[64]  in the  GlobalState  struct (https://github.com/bitaxeorg/ESP-Miner/blob/8b995f9/main/global_state.h#L116) — a fixed array, not a pointer. It can never be NULL. The  strlen()  call is always safe.
> 
> Let me know if I am mistaken, happy to fix those if needed

Thank you for replying. Yes you're right the NULL dereference risk does not happen. I was missing that.

WantClue (approved):
lgtm now 🚀

mutatrum:
Congratulations, @camalolo! :rocket: 

This was a long road and I think holds the record for longest conversation on a PR. :1st_place_medal:

camalolo:
Hey @mutatrum ! Thanks a lot 😊

Thank you and everyone who helped testing and provided feedback, that was
the most important part

frantisekbohacek:
Thank you for this feature!

Issue #1242, opened 2025-09-20
Memory leak


mutatrum:
It's leaking ~40kb in 24 hours.

<img width="1915" height="976" alt="Image" src="https://github.com/user-attachments/assets/c2752e20-3cd2-4a14-9164-0cd57cfeed7e" />

duckaxe:
Where should we look? AxeOS or ESP Miner?

mutatrum:
This is heap on the ESP32, so esp-miner. It might have been there already, but the new charts makes it easier to show.

mutatrum:
It might just be the statistics task collecting 720 data points.

mutatrum:
Confirmed, I let it run for 6 hours on the shortest statistics setting (every 30 seconds for 6 hours). After 6 hours and 10 minutes, this was the graph:

<img width="1915" height="976" alt="Image" src="https://github.com/user-attachments/assets/e5a6317c-6175-43ab-9959-b1ce5532ed2e" />

It might make sense to switch from dynamic allocation to static, if we then run into issues, we can catch the error and still continue running. And I guess this graph will get people asking questions.

WantClue:
agreed, a static allocation makes more sense here, this could get confusing if we actually run into a mem leak issue

duckaxe:
<img width="1602" height="446" alt="Image" src="https://github.com/user-attachments/assets/74052db9-3754-4ab3-9b91-7cb3a9b1c4ce" />

4h with data logging off.

PR #1246, opened 2025-09-22
Static allocation of statistics buffer

terratec:
For issue #1242

### DIFF
--- a/main/tasks/statistics_task.c
+++ b/main/tasks/statistics_task.c
@@ -18,8 +18,8 @@
 static const char * TAG = "statistics_task";
 
 static StatisticsDataPtr statisticsBuffer;
-static int16_t statisticsDataStart = -1;
-static int16_t statisticsDataEnd = -1;
+static uint16_t statisticsDataStart;
+static uint16_t statisticsDataSize;
 static pthread_mutex_t statisticsDataLock = PTHREAD_MUTEX_INITIALIZER;
 
 static const uint16_t maxDataCount = 720;
@@ -32,9 +32,6 @@ void createStatisticsBuffer()
 
         if (NULL == statisticsBuffer) {
             statisticsBuffer = (StatisticsDataPtr)heap_caps_malloc(sizeof(struct StatisticsData) * maxDataCount, MALLOC_CAP_SPIRAM);
-            statisticsDataStart = -1;
-            statisticsDataEnd = -1;
-
             if (NULL == statisticsBuffer) {
                 ESP_LOGW(TAG, "Not enough memory for the statistics data buffer!");
             }
@@ -53,8 +50,8 @@ void removeStatisticsBuffer()
             heap_caps_free(statisticsBuffer);
 
             statisticsBuffer = NULL;
-            statisticsDataStart = -1;
-            statisticsDataEnd = -1;
+            statisticsDataStart = 0;
+            statisticsDataSize = 0;
         }
 
         pthread_mutex_unlock(&statisticsDataLock);
@@ -74,17 +71,16 @@ bool addStatisticData(StatisticsDataPtr data)
     pthread_mutex_lock(&statisticsDataLock);
 
     if (NULL != statisticsBuffer) {
-        statisticsDataEnd++;
-        statisticsDataEnd = statisticsDataEnd % maxDataCount;
+        statisticsDataSize++;
 
-        if (-1 == statisticsDataStart) {
-            statisticsDataStart = statisticsDataEnd;
-        } else if (statisticsDataStart == statisticsDataEnd) {
+        if (maxDataCount < statisticsDataSize) {
+            statisticsDataSize = maxDataCount;
             statisticsDataStart++;
             statisticsDataStart = statisticsDataStart % maxDataCount;
         }
 
-        statisticsBuffer[statisticsDataEnd] = *data;
+        const uint16_t last = (statisticsDataStart + statisticsDataSize - 1) % maxDataCount;
+        statisticsBuffer[last] = *data;
         result = true;
     }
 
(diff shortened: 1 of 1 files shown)


main/tasks/statistics_task.c
```diff
@@ -29,75 +29,87 @@ StatisticsNodePtr addStatisticData(StatisticsNodePtr data)
...
-        if ((NULL != statisticsDataEnd) && (newData != statisticsDataEnd)) {
-            statisticsDataEnd->next = newData; // link data block
+        if (statisticsDataStart == statisticsDataEnd) {
+            statisticsDataStart = statisticsDataStart->next;
+        } else if(NULL == statisticsDataStart) {
+            statisticsDataStart = statisticsDataEnd;
         }
-        statisticsDataEnd = newData; // set DataEnd to new data
+
+        StatisticsNextNodePtr next = statisticsDataEnd->next;
```
mutatrum: I don't think this needs to be a linked list now, it can just be an array.

main/tasks/statistics_task.c
```diff
@@ -29,75 +29,87 @@ StatisticsNodePtr addStatisticData(StatisticsNodePtr data)
...
     return nextNode;
 }
 
-void clearStatisticData()
+void createStatisticsBuffer()
 {
-    if (NULL != statisticsDataStart) {
+    removeStatisticsBuffer();
+
+    StatisticsNodePtr buffer = (StatisticsNodePtr)malloc(sizeof(struct StatisticsData) * maxDataCount);
```
mutatrum: Use `heap_caps_malloc` with `MALLOC_CAP_DEFAULT | MALLOC_CAP_SPIRAM` to force it into PSRAM?

main/tasks/statistics_task.c
terratec: Should we instead use a virtual index (0-720) for data retrieval and convert it to the current index (e.g. 359-719-0-358)?
Or should we return all the copied and reordered data at once with newly allocated memory?

mutatrum: Not sure if I understand the question. I would keep stack of the size and the start of the data in the array (head), then you can convert any position to the array position by modulo 720.

WantClue:
There is a current potential for a memory leak with duplicate staticbuffer creating.

my suggested change would be

```
if (NULL == statisticsBuffer) {
-        createStatisticsBuffer();
+        pthread_mutex_lock(&statisticsDataLock);
+        if (NULL == statisticsBuffer) {
+            pthread_mutex_unlock(&statisticsDataLock);
+            createStatisticsBuffer();
+        } else {
+            pthread_mutex_unlock(&statisticsDataLock);
+        }
     }
```

Also please make sure to simplify it and I agree with @mutatrum that there is no need for a linked list

main/tasks/statistics_task.c
```diff
@@ -16,110 +17,105 @@
...
-        createStatisticsBuffer();
-    }
-
-    if (NULL != statisticsBuffer) {
         pthread_mutex_lock(&statisticsDataLock);
 
-        statisticsDataEnd = statisticsDataEnd->next;
+        if (NULL == statisticsBuffer) {
+            statisticsBuffer = (StatisticsDataPtr)heap_caps_malloc(sizeof(struct StatisticsData) * maxDataCount, MALLOC_CAP_DEFAULT | MALLOC_CAP_SPIRAM);
+            statisticsDataStart = -1;
```
mutatrum: If you init start and end at 0, I think they're valid and you don't need to test on -1 further down, assuming you already test for statisticsBuffer being allocated or not.

main/tasks/statistics_task.c
```diff
@@ -16,110 +17,105 @@
...
-    if (NULL != statisticsBuffer) {
-        pthread_mutex_lock(&statisticsDataLock);
+    bool result = false;
+
+    if ((NULL == statisticsBuffer) || (NULL == dataOut) || (maxDataCount <= index)) {
+        return result;
+    }
 
-        free(statisticsBuffer);
+    pthread_mutex_lock(&statisticsDataLock);
```
mutatrum, line 100: To further simplify and minimize the mutex locks, it's probably more efficient to just copy the whole array into a temporary copy within a single mutex lock, instead of a lock per entry.

Not sure what's easier, return the whole array as-is, with current start/end, or do 2 memcpys into a fit-to-size array.

main/tasks/statistics_task.c
terratec: Start=0 and End=0 means 1 valid data element. After buffer allocation, we would return garbage data until the first element is actually written.
`addStatisticData` would also need a special case to not increment `statisticsDataEnd`.

main/tasks/statistics_task.c
terratec: Yes, I'm torn between entry or block based. Is this something for a new PR?

Things to clarify:
- measure 1 single lock (maybe 2 or 4 or x locks are needed (intern) to not delay the statistics task) ?
- return whole array as-is or a rearranged fit-to-size array?
- buffer handling (prebuffer from caller or ownership transfer or ...) ?

main/tasks/statistics_task.c
mutatrum: Ah I see, you're right. Maybe try with start/size, instead of start/end, then it should be valid. Same holds as above, you don't need the -1 checks then later on.

main/tasks/statistics_task.c
mutatrum: One memcpy is for sure faster than 720 mutex lock/unlocks.

Array as is, is a straight copy. Drawback is that http_server needs to know the administration. A fit-to-size array would return an array with the start at 0 and the last item at size-1, so http-server can just loop over it.

Not sure about buffer handling. With fit-to-size, I think statistics_task should allocate it, to shield http_server from most of the internal details.

It's fine to push this to the next PR, having this out is already a benefit.

main/tasks/statistics_task.c
terratec: nice!

mutatrum: Apologies for the conflict, I didn't put that change in to keep out of this PR, but not having that line in #1280 kinda messed up testing on that one, as the statistics data (previous version) kept on gobbling up all internal memory.

main/tasks/statistics_task.c
terratec: #1283

main/tasks/statistics_task.c
terratec: No problem :)

PR #1247, opened 2025-09-23
Swarm Grid View

duckaxe:
This PR introduces the grid view for swarm. The grid view can be accessed via a toggle switch. This allows users to choose between list and grid view. Their selection is stored.

**Sorting**
* List view: Via table columns
* Grid view: Via dropdown

Both sorting methods are synchronized.

**Filter**
Filter is on:
* `IP`
* `Hostname`
* `ASICModel`
* `deviceModel`

**Desktop**
Desktop

**Mobile**
Mobile

**Video: Grid View**

https://github.com/user-attachments/assets/1ad937ab-d03e-46ce-9fca-08cd59b44c37

**Video: Grid View Filter**

https://github.com/user-attachments/assets/4de6cac5-e33a-420b-b722-783244df8c43

### DIFF
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.html
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.html
@@ -88,15 +88,19 @@
                     <div class="card flex justify-content-between mb-0 h-full">
                         <div class="flex flex-column justify-content-between">
                             <div class="flex flex-column">
-                                <a [href]="'http://' + axe.IP" target="_blank" class="flex flex-column">
-                                    <span class="text-800" pTooltip="Hostname" tooltipPosition="top">
-                                        {{axe.hostname}}
-                                        <span class="ml-1 text-lg line-height-1" [ngStyle]="{color: axe.swarmColor}">●</span>
-                                    </span>
-                                    <span class="text-900" pTooltip="IP" tooltipPosition="top">
+                                <div class="flex flex-column">
+                                    <div>
+                                        <a [href]="'http://' + axe.IP" target="_blank" class="text-900" pTooltip="Hostname" tooltipPosition="top">
+                                            {{axe.hostname}}
+                                        </a>
+                                        <span class="ml-1 text-lg line-height-1 cursor-pointer" [ngStyle]="{color: axe.swarmColor}" pTooltip="{{stringifyDeviceLabel(axe)}}" tooltipPosition="top">
+                                            ●
+                                        </span>
+                                    </div>
+                                    <a [href]="'http://' + axe.IP" target="_blank" class="text-800" pTooltip="IP" tooltipPosition="top">
                                         {{axe.IP}}
-                                    </span>
-                                </a>
+                                    </a>
+                                </div>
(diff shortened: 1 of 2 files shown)


mutatrum:
I like the grid view, but have a few suggestions:
 * I would go for an icon bar (f.e. `pi-list` and `pi-th-large`) instead of the toggle;
 * I'm not keen on extracting the sort out of the table. I understand why you would want this, but for the table, clicking on the table headers is better UX. Not sure how to solve this;
 * The obvious thing that's missing is a filter box. Thinking about it, maybe that would be a good thing to put where the sort dropdown is. Filtering in grid makes more sense than sorting a grid anyways, IMO. Plus one would get filtering on the table view as well.

duckaxe:
> * I would go for an icon bar (f.e. `pi-list` and `pi-th-large`) instead of the toggle;

The toggle was ready to use, so I decided not to use icons. 

Icons require explanation. So we have to work with either text or tooltips. And they have to be displayed as clickable buttons so that users know they can click on them. That's exactly what a toggle is for.

But I can try to implement an icon bar like you suggested.

> * I'm not keen on extracting the sort out of the table. I understand why you would want this, but for the table, clicking on the table headers is better UX. Not sure how to solve this;

I don't have a good idea how to solve this elegantly either. Yes, sorting in the table is easy, but we also need sorting for the grid. Does it make sense to implement two different sorting methods? That's why I decided on one general sorting method. 

Or we could do two sorting methods: one for the list within the table (no dropdown), and one for the grid, then show the dropdown.

> * The obvious thing that's missing is a filter box. Thinking about it, maybe that would be a good thing to put where the sort dropdown is. Filtering in grid makes more sense than sorting a grid anyways, IMO. Plus one would get filtering on the table view as well.

Yes, it will come. To avoid increasing the complexity of this PR, I only wanted to implement functions that are already there.

duckaxe:
- [x] Add icon bar as toggle for the grid view.
- [x] Restore table sorting. Dropdown is visible for grid view only.
- [x] Add filter for both views.

PR description updated.

mutatrum:
> * [x]  Add icon bar as toggle for the grid view.
> * [x]  Restore table sorting. Dropdown is visible for grid view only.
> * [x]  Add filter for both views.
> 
> PR description updated.

That looks absolutely wicked! Only little nit is I would put the icons right of the filter box, as they are on the logs window. IMO the function of icons is very clear.

Filter is on IP, hostname and device name, I assume? Maybe asic model (bm1370) as well, if that's available?

Looking at it again, asic model is not even in the list view. Not sure if there's room for that. Maybe next PR sometime :-)

duckaxe:
Made few adjustments. 

Filter is on:
* `IP`
* `Hostname`
* `ASICModel`
* `deviceModel`

Updated screenshots, video, description.

duckaxe:
Vid for filter:

https://github.com/user-attachments/assets/8571cd21-d320-4be1-8107-5b651fff6c17

ghost:
This is much better - I like the grid view as well (thumbs up).

The dark blue text on a dark background just was not working for me with #1232.

PR #1249, opened 2025-09-25
Hashrate registers

mutatrum:
First pass at reading hashrate registers from the ASIC chip.

Current code is only active for BM1370. I think the same mechanism can be enabled for the BM1366 and BM1368, but I haven't done this yet. I tried to get as much asic specific code out. This should also work for multi-chip devices.

The aggregated hashrate is only passed through an EMA filter to get an ~30 second average. When the frequency is changed, the counters are cleared, to quickly reflect the new hashrate.

Open points are to add the individual domain hashrates to the api, and show them somewhere. The error rate is currently exposed as a numeric value, but it is probably better to show percentage of total hashrate. There's also a temperature register, which needs some init, haven't looked in detail at that. There's some placeholder code and some commented out code lifted out of the NerdAxe repo.

The registers are not reset, they will wrap over at some point. My assumption is it will use the full `uint32_t` range, if that's the case it is handled in the code.

IMO the register hashrate should be the main hashrate for devices that have these registers. It's way more stable and settles much more quickly than the current stochastic hashrate.

Made possible by lots and lots of reverse engineering and coding by @CTCaer, @korbin, @shufps, @WantClue and many others.

Dashboard, 12 hours, comparing stochastic hashrate (green) vs register hashrate:
image
Same spikiness as it's still a stochastic hashrate, but with a difficulty of 1, so the variance is a lot smaller.

Expected hashrate: 1.07 Gh/s
Current hashrate: 1.106 Gh/s  (stdev 104)
Register hashrate: 1.070 Gh/s (stdev 6)

API:
image

### DIFF
--- a/main/tasks/hashrate_monitor_task.c
+++ b/main/tasks/hashrate_monitor_task.c
@@ -45,14 +45,21 @@ static void clear_measurements(HashrateMonitorModule * HASHRATE_MONITOR_MODULE, 
     memset(HASHRATE_MONITOR_MODULE->error_measurement, 0, asic_count * sizeof(measurement_t));
 }
 
+static float hash_counter_to_ghs(uint32_t duration_ms, uint32_t counter)
+{
+    if (duration_ms == 0) return 0.0f;
+    float seconds = duration_ms / 1000.0;
+    float hashrate = counter / seconds * (float)HASH_CNT_LSB; // Make sure it stays in float
+    return hashrate / 1e9f; // Convert to Gh/s
+}
+
 static void update_measurement(uint32_t time_ms, uint32_t value, measurement_t * measurement, int asic_nr)
 {
     uint32_t previous_time_ms = measurement[asic_nr].time_ms;
     if (previous_time_ms != 0) {
-        uint32_t value_diff = value - measurement[asic_nr].value; // Compute counter difference, handling uint32_t wraparound
-        float seconds = (time_ms - previous_time_ms) / 1000.0;
-        float hashrate = (float)value_diff / seconds * HASH_CNT_LSB;
-        measurement[asic_nr].hashrate = hashrate / 1e9; // Convert to Gh/s
+        uint32_t duration_ms = time_ms - previous_time_ms;
+        uint32_t counter = value - measurement[asic_nr].value; // Compute counter difference, handling uint32_t wraparound
+        measurement[asic_nr].hashrate = hash_counter_to_ghs(duration_ms, counter);
     }
 
     measurement[asic_nr].value = value;


CTCaer:
From my notes:
```c
// All counters can be reset with a any write.
#define BM_HASH_CNT_ERROR   0x4C    // Hash error count for all.
#define BM_HASH_CNT_DOM0    0x88    // Hash count domain 0: 00-1F cores.
#define BM_HASH_CNT_DOM1    0x89    // Hash count domain 1: 20-3F cores.
#define BM_HASH_CNT_DOM2    0x8A    // Hash count domain 2: 40-5F cores.
#define BM_HASH_CNT_DOM3    0x8B    // Hash count domain 3: 60-7F cores.
#define BM_HASH_CNT_TOTAL   0x8C    // Hash count for all:  00-7F cores.
#define BM_HASH_CNT_VALID   0x90    // Hash count for all without errors. Needs counter div correction.
#define  HASH_CNT_LSB       0x100000000uLL // All counters above increment every 4 billion hashes.
#define  HASH_CNT_LSB_MH    0x1000  // Value used by OFW (4.6% lower).
#define  HASH_CNT_DIV_WAR(x) ((double)(x) * 256 / 255) // Only for BM_HASH_CNT_VALID.
#define BM_HASH_CNT_TARGET  0x94    // Result count at/above target (ticket).
```
All these work with bm1366/68 too. And you can do DOM1-3 requests to 1366 too but they should be zero (maybe DOM0 is also zero).
That's not actually tested by me, but it's evidenced by RE that validates that all 3 are from the same series/gen with 1366 being a rev1 design of that series and the other 2 being rev2. Or bm1366 being a cut down revision of 1368/70. Plus that the total/valid/err were already tested by others.

0x80 is not register response but job/work result, so it should be appropriately named.
So either
```c
#define RESPONSE_CMD 0x00
#define RESPONSE_JOB 0x80
```
or 
```c
#define RESPONSE_REG 0x00
#define RESPONSE_JOB 0x80
```
But since some commands can cause a reply with a write, which uses the same exact format as a register read, `CMD` is more proper for 0x00.

The `0xB4` reg is either not real, aka a virtual reg which is internally replaced to get the i2c temp sensor ic temp or the asic supports ext temp sensor.
I doubt it's the latter because I don't see any calibration happening for adc in (and TEMP N/P are still used as analog out).

mutatrum:
> ```c
> #define BM_HASH_CNT_TARGET  0x94    // Result count at/above target (ticket).
> ```
Refactored them into a lookup map. This probably makes different domain count or availability of certain registers per device easier for other models, see below.

> All these work with bm1366/68 too. And you can do DOM1-3 requests to 1366 too but they should be zero (maybe DOM0 is also zero). That's not actually tested by me, but it's evidenced by RE that validates that all 3 are from the same series/gen with 1366 being a rev1 design of that series and the other 2 being rev2. Or bm1366 being a cut down revision of 1368/70. Plus that the total/valid/err were already tested by others.

I'll test with 1366/1368, and probably move some of the definitions into DeviceConfig. Probably register map and maybe `HASH_CNT_LSB`. I also have a 1397, and there are at least some registers there, not sure if I can get anything out. yet. Haven't researched that too much.

> 0x80 is not register response but job/work result, so it should be appropriately named. So either

Forgot to clean that up, I inverted the if statement and missed the rename, thanks! Decided to get rid of them completely by properly defining the bm1370_asic_result, at least the cmd part. The job part could also improve a bit, so we don't need as many bit fiddling in code.

> The `0xB4` reg is either not real, aka a virtual reg which is internally replaced to get the i2c temp sensor ic temp or the asic supports ext temp sensor. I doubt it's the latter because I don't see any calibration happening for adc in (and TEMP N/P are still used as analog out).

Removed for now, to make it clearer. Finding out if temp and/or voltage registers is possible will be for another PR.

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -293,6 +293,9 @@ export class HomeComponent implements OnInit, OnDestroy {
         stats.statistics.forEach(element => {
           element[idxHashrate] = element[idxHashrate] * 1000000000;
           switch (chartLabelValue(chartY1DataLabel)) {
+            case eChartLabel.hashrateRegister:
+              element[idxChartY1Data] = element[idxChartY1Data] * 1000000000;
```
mutatrum, line 297: Ideally, all hashrates from the API should be in h/s, not Gh/s.

components/asic/bm1370.c
```diff
@@ -34,14 +34,42 @@
 
 #define MISC_CONTROL 0x18
 
+static const register_type_t REGISTER_MAP[] = {
```
mutatrum, line 37: This will probably move to device_config for a general solution.

main/self_test/self_test.c
```diff
@@ -330,7 +330,7 @@ bool self_test(void * pvParameters)
         tests_done(GLOBAL_STATE, false);
     }
 
-    POWER_MANAGEMENT_init_frequency(&GLOBAL_STATE->POWER_MANAGEMENT_MODULE);
+    POWER_MANAGEMENT_init_frequency(GLOBAL_STATE);
```
mutatrum, line 333: This now also calculates expected_hashrate, which is be used on L440.

main/tasks/hashrate_monitor_task.c
```diff
@@ -0,0 +1,154 @@
...
+        return;
+    }
+
+    // Reset statistics on start and when frequency changes
+    if (POWER_MANAGEMENT_MODULE->frequency_value != frequency_value) {
+        clear_measurements(HASHRATE_MONITOR_MODULE, asic_count);
+        frequency_value = POWER_MANAGEMENT_MODULE->frequency_value;
+    }
+
+    switch(register_type) {
```
mutatrum, line 128: Probably move to a lookup map for this as well.

mutatrum:
Follow-up tasks:
* Test other chips
* Expose domain hashrates on dashboard
* Replace current hashrate graph
* Bump share diff to closer to the pool diff, as it's no longer necessary to have 256 diff.

skot:
I suggest keeping the chip difficulty at a fixed number. I'd like to add a chip "zombie" watchdog timeout in the future.

PR #1253, opened 2025-09-27
alloc static tasks into psram

WantClue:
Moving tasks into the PSRAM will leave us more room on the internal ram for other things that might come, we're currently heavily loaded on the internal ram and should seek more into PSRAM.

This is currently proposed as a draft and needs some more changes as well as additional testing.


mutatrum:
Superseded by #1286.

PR #1254, opened 2025-09-28
Allow ipv6 stratum and local address

### LINKED ISSUE #263: Add IPv6 support to AxeOS
Noticed that AxeOS is not compiled with IPv6 support and as such doesn't get an IPv6 address even when the LAN has full IPv6 support.

IPv6 is already the default Internet Protocol corresponding in some countries to over 40 - 50% of peak traffic according to several statistics specialist websites. IPv4 is legacy and several Internet Broadband Services Providers don't issue a Public IPv4 address to users anymore forcing the usage of CGNAT which makes the connecting have to pass through extra equipment, adding extra delay which may make a difference when finding a block.

Having a native IPv6 connection not only aligns it to use the newer and more modern Internet Protocol, but gives benefits in terms of network performance, resiliency to IPv4 and in some cases can be an alternative for some network filters.

ESP32 has enough space for IPv6 LWIP network stack along with IPv4 and is support for quiet a while since ESP-IDF V2.1. Tasmota firmware which is very well know, widely used and produces a bigger binary has IPv6 enabled by default for ESP32.

Verified is by trying to connect to solo6.ckpool.org which has IPv6 support and got the following on the logs:
```
₿ (21536) stratum_task: Socket unable to connect to solo6.ckpool.org:3333 (errno 113)
₿ (26536) stratum_task: Get IP for URL: solo6.ckpool.org
```

References:
[1] - https://tasmota.github.io/docs/IPv6/

WantClue:
This introduces ipv6 to the bitaxe, it enables ipv6 resolution of stratum v1 connections and local resolution.

closes #263 

Additional note:
- maybe remove http_server get endpoint for the ipv6, was just introduced for testing but should not be necessary and optionally removed before merge


WantClue:
Will do some refacotring of the stratum code later today @eandersson

Travetown:
It would be great if the system view showed whether IPv4 or IPv6 is being used for the pool. Perhaps with a small dot or something similar.

eandersson:
> It would be great if the system view showed whether IPv4 or IPv6 is being used for the pool. Perhaps with a small dot or something similar.

I like that too, but could be a follow up PR as well.

WantClue:
> It would be great if the system view showed whether IPv4 or IPv6 is being used for the pool. Perhaps with a small dot or something similar.

Ye I agree with @eandersson that should be a follow up PR

PR #1257, opened 2025-10-01
Bitaxe Hex 303

mutatrum:
Migrated Hex 303 support from https://github.com/bitaxeorg/esp-miner-multichip. Changes include:

 * 303 hardware configuration in `DeviceConfig.h`;
 * Enable `TMP1075` for board temperature reading, increased resolution from 8 bit to 12 bit;
 * Added Hex configuration for `TPS546`;
 * Added `EMC2302` fan controller support. Both fans are driven the same, rpm for both fans is shown;
 * Added voltage domain number;
 * Changed how ASIC nr is taken from nonce;
 * Read **Measured ASIC Voltage** from `TPS546`, instead of the `ADC`. This is changed for all devices with `TPS546`:
   * Ultra 207
   * Supra 402+
   * Gamma
   * Gamma Turbo

:rotating_light: Make sure to unplug the power and cold restart it after the firmware update to fully reset all the peripherals :rotating_light:

Things to do:

 * ~Verify fan controller (I don't have a PWM fan currently);~
 * ~Verify self test;~
 * ~Properly support second fan header;~
 * ~Generalize function to get ASIC address out of nonce for other chips;~
 * Temperature measurements in AxeOS are shown as ASIC Temperature. This should be Board Temperature for the Hex. Not sure how to improve this.

Running 2.10 on Hex:

image
image
image

### DIFF
--- a/main/thermal/EMC2302.c
+++ b/main/thermal/EMC2302.c
@@ -53,7 +53,7 @@ esp_err_t EMC2302_init()
 
     // Set the minimum fan speed measured and reported to 500 RPM
     ESP_RETURN_ON_ERROR(set_fan_range(EMC2302_FAN1_CONFIG1, RNG_500_RPM, &fan1_multiplier), TAG, "Failed to set fan 1 config");
-    ESP_RETURN_ON_ERROR(set_fan_range(EMC2302_FAN2_CONFIG1, RNG_500_RPM, &fan2_multiplier), TAG, "Failed to add fan 2 config");
+    ESP_RETURN_ON_ERROR(set_fan_range(EMC2302_FAN2_CONFIG1, RNG_500_RPM, &fan2_multiplier), TAG, "Failed to set fan 2 config");
 
     return ESP_OK;
 }
--- a/main/thermal/thermal.c
+++ b/main/thermal/thermal.c
@@ -22,13 +22,13 @@ esp_err_t Thermal_init(DeviceConfig * DEVICE_CONFIG)
         }
     }
     if (DEVICE_CONFIG->EMC2103) {
-        ESP_RETURN_ON_ERROR(EMC2103_init(DEVICE_CONFIG->temp_offset), TAG, "Failed to initialise EMC2103");;
+        ESP_RETURN_ON_ERROR(EMC2103_init(DEVICE_CONFIG->temp_offset), TAG, "Failed to initialise EMC2103");
     }
     if (DEVICE_CONFIG->EMC2302) {
-        ESP_RETURN_ON_ERROR(EMC2302_init(), TAG, "Failed to initialise EMC2302");;
+        ESP_RETURN_ON_ERROR(EMC2302_init(), TAG, "Failed to initialise EMC2302");
     }
     if (DEVICE_CONFIG->TMP1075) {
-        ESP_RETURN_ON_ERROR(TMP1075_init(DEVICE_CONFIG->temp_offset), TAG, "Failed to initialise TMP1075");;
+        ESP_RETURN_ON_ERROR(TMP1075_init(DEVICE_CONFIG->temp_offset), TAG, "Failed to initialise TMP1075");
     }
 
     return ESP_OK;


mutatrum:
Self test has an issue:
```
I (1127) self_test: Running manual self-test
I (1133) display: SSD1306 (128x32)
I (1136) display: Install panel IO
I (1140) display: Install panel driver
I (1146) display: Initialize LVGL
I (1149) LVGL: Starting LVGL task
I (1169) display: Rotation: 180
I (1271) display: Display init success!
I (1271) self_test: DISPLAY init success!
I (1271) input: Install button driver
I (1274) self_test: INPUT init success!
E (6270) task_wdt: Task watchdog got triggered. The following tasks/users did not reset the watchdog in time:
E (6270) task_wdt:  - IDLE0 (CPU 0)
E (6270) task_wdt: Tasks currently running:
E (6270) task_wdt: CPU 0: main
E (6270) task_wdt: CPU 1: IDLE1
E (6270) task_wdt: Print CPU 0 (current core) backtrace
```

mutatrum:
Replaced the fan with a Noctua NF-R8 PWM fan. As there's a fairly large difference between the two board temperature sensors, it might make more sense to take the maximum of the two.

image

It looks like the measured RPM is 50% of what it actually is. This fan is rated to a max rpm of 1800.

mutatrum:
> Replaced the fan with a Noctua NF-R8 PWM fan. As there's a fairly large difference between the two board temperature sensors, it might make more sense to take the maximum of the two.
> 
> It looks like the measured RPM is 50% of what it actually is. This fan is rated to a max rpm of 1800.

Fan controller has been configured properly now, and it takes the highest of the two temperatures. With the fan controller set to 60 °C:

image

The fan controller also starts reporting from 500 RPM, by default it only reported above 1000 RPM.

chillingbee:
Really great work so far! It runs really well but one thing i would mention: in the firmware of tch you could see the submission per chip and that was very helpful in the time of building and after... would be great to see that again!

mutatrum:
> Really great work so far! It runs really well but one thing i would mention: in the firmware of tch you could see the submission per chip and that was very helpful in the time of building and after... would be great to see that again!

That will be added later, with #1249. I'm currently working on getting this enabled for the BM1366 and multichip devices.

duckaxe:
What do you think about placing the two temp and fan values side by side (desktop only)? It's much clearer. On mobile and for single values, it remains as before, one below the other.

**Before**
Before

**After**
Screenshot

WantClue:
Als maybe we need to adjust it for the hex board in the ui, afaik we're not reading just two asic chips. If I remember correctly it's board temp 1 and 2

mutatrum:
> Als maybe we need to adjust it for the hex board in the ui, afaik we're not reading just two asic chips. If I remember correctly it's board temp 1 and 2

In the esp-multichip it's indeed *Board Temperature*. Will do a simple version for now, but it would be nice to have a bit better system for temp sensors and fan controllers in the future.

mutatrum:
> What do you think about placing the two temp and fan values side by side (desktop only)? It's much clearer. On mobile and for single values, it remains as before, one below the other.

Or split out the temperature and fan bars, so we have 3 panes at the bottom.

WantClue:
@mutatrum I think we can merge this into and bring back the hex 303 with 2.11.0
I know you wanted still to do the temp measurements

main/thermal/EMC2302.c
```diff
@@ -0,0 +1,118 @@
...
+ */
+esp_err_t EMC2302_init()
+{
+    ESP_RETURN_ON_ERROR(i2c_bitaxe_add_device(EMC2302_I2CADDR_DEFAULT, &EMC2302_dev_handle, TAG), TAG, "Failed to add device");
+
+    ESP_LOGI(TAG, "EMC2302 init");
+
+    // Set the minimum fan speed measured and reported to 500 RPM
+    ESP_RETURN_ON_ERROR(set_fan_range(EMC2302_FAN1_CONFIG1, RNG_500_RPM, &fan1_multiplier), TAG, "Failed to set fan 1 config");
+    ESP_RETURN_ON_ERROR(set_fan_range(EMC2302_FAN2_CONFIG1, RNG_500_RPM, &fan2_multiplier), TAG, "Failed to add fan 2 config");
```
WantClue: "Failed to set fan 2 config"

main/thermal/thermal.c
```diff
@@ -1,28 +1,37 @@
...
         if (DEVICE_CONFIG->emc_ideality_factor != 0x00) {
             ESP_LOGI(TAG, "EMC2101 configuration: Ideality Factor: %02x, Beta Compensation: %02x", DEVICE_CONFIG->emc_ideality_factor, DEVICE_CONFIG->emc_beta_compensation);
             EMC2101_set_ideality_factor(DEVICE_CONFIG->emc_ideality_factor);
             EMC2101_set_beta_compensation(DEVICE_CONFIG->emc_beta_compensation);
         }
-        return res;
     }
     if (DEVICE_CONFIG->EMC2103) {
-        return EMC2103_init(DEVICE_CONFIG->emc_temp_offset);
+        ESP_RETURN_ON_ERROR(EMC2103_init(DEVICE_CONFIG->temp_offset), TAG, "Failed to initialise EMC2103");;
```
WantClue: remove double semi colon

main/thermal/thermal.c
```diff
@@ -1,28 +1,37 @@
...
+    }
+    if (DEVICE_CONFIG->EMC2302) {
+        ESP_RETURN_ON_ERROR(EMC2302_init(), TAG, "Failed to initialise EMC2302");;
```
WantClue: here as well

main/thermal/thermal.c
```diff
@@ -1,28 +1,37 @@
...
+    }
+    if (DEVICE_CONFIG->TMP1075) {
+        ESP_RETURN_ON_ERROR(TMP1075_init(DEVICE_CONFIG->temp_offset), TAG, "Failed to initialise TMP1075");;
```
WantClue: and here

main/tasks/hashrate_monitor_task.c
WantClue: this malloc should have already been changed to heap_caps right ?

mutatrum: This was fixed in another PR, picked up with merge.

main/thermal/thermal.c
mutatrum: Done (3x)

main/thermal/EMC2302.c
mutatrum: Good catch, fixed!

mutatrum:
Review comments fixed and conflict resolved.

> @mutatrum I think we can merge this into and bring back the hex 303 with 2.11.0 I know you wanted still to do the temp measurements

As discussed, that's for another PR, needs more though to fix nicely.

WantClue (approved):

PR #1271, opened 2025-10-14
Hashrate registers for BM1366, BM1368 and BM1397

mutatrum:
This adds support for the other ASICs. There's a lot of code duplication, especially between the 66/68/70, this takes a bit more effort to clean up.

For the 66/68/70, the hashrate matches the expected hashrate almost perfectly, it seems on the 1397, it's slightly lower, so not 100% sure what's going on there, yet.

Now all ASICs are supported, I would opt for making this the only hashrate, and ditch the current 256 difficulty mask hashrate. They use the same principle, but the register hashrate uses a diff 1 difficulty mask to count (for 66/68/70) and something like diff 8 for the 1397. So the variance is still there, but on a orders smaller range.

Follow-up PR would be to raise the ticket mask to be highest power of 2 below pool diff, as there's no use for the 256 difficulty mask anymore. Especially with higher hashrate or multi-chip devices, this can make quite a big difference in the amount of work needed.


mutatrum:
Running on the Max:

image

WantClue (approved):
LFG loading it up to all models rn

PR #1274, opened 2025-10-15
Hashrate Heatmap

duckaxe:
Based on https://github.com/bitaxeorg/ESP-Miner/pull/1249 this PR displays a hashrate heatmap for available hashrate domains.

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -27,104 +27,45 @@
     </p-message>
 
     <div class="grid">
-        <div class="col-12 md:col-6 xl:col-3 card-flip">
-            <div class="card mb-0 p-0" [ngClass]="{'is-flipped': heatmapVisible}">
-                <div class="card__face card__face--front p-4 md:p-5">
-                    <div class="flex flex-column justify-content-between">
-                        <div class="flex flex-row justify-content-between align-items-center mb-3">
-                            <span class="block text-500 font-medium">Hashrate</span>
-                            <button pButton type="button"
-                                class="cursor-pointer p-button-icon-only-borderless"
-                                (click)="heatmapVisible = !heatmapVisible"
-                                pTooltip="Heatmap"
-                                tooltipPosition="bottom">
-                                    <i class="pi pi-table"></i>
-                            </button>
-                        </div>
+        <div class="col-12 md:col-6 xl:col-3">
+            <div class="card mb-0">
+                <div class="flex flex-column justify-content-between">
+                    <span class="block text-500 font-medium mb-3">Hashrate</span>
 
-                        <ng-container *ngIf="!info.power_fault; else powerFault">
-                            <div class="flex align-items-baseline gap-2 mb-3">
-                                <ng-container *ngIf="info.hashrateMonitor; else noHashrateMonitor">
-                                    <span class="text-900 font-medium text-2xl">
-                                        {{info.hashrateMonitor.hashrate | hashSuffix}}
-                                    </span>
-                                    <span class="text-500 text-xs">
-                                        Error:
-                                        <span class="text-{{info.hashrateMonitor.errorCount === 0 ? 'green' : 'red'}}-500">
-                                            {{info.hashrateMonitor.errorCount}}
-                                        </span>
-                                    </span>
-                                </ng-container>
-                                <ng-template #noHashrateMonitor>
-                                    {{info.hashRate | hashSuffix}}
-                                </ng-template>
-                            </div>
-
-                            <div>
-                                Average:
-                                <span class="text-green-500 font-medium">
-                                    {{getHashrateAverage() | hashSuffix}}
+                    <ng-container *ngIf="!info.power_fault; else powerFault">
+                        <div class="flex align-items-baseline gap-2 mb-3">
+                            <ng-container *ngIf="info.hashrateMonitor; else noHashrateMonitor">
+                                <span class="text-900 font-medium text-2xl">
+                                    {{info.hashrateMonitor.hashrate | hashSuffix}}
                                 </span>
-                            </div>
-                            <div class="text-500 text-xs" *ngIf="info.expectedHashrate">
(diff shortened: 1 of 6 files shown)


mutatrum:
Looks good! I'll give it a spin on the Hex :crossed_fingers:

0xf0xx0:
a little borked on firefox when flipped x3

image

mutatrum:
Maybe https://primeng.org/tabs is an option? Might align neatly with where the title is currently.

duckaxe:
@0xf0xx0 Thanks for testing. Fixed.

mutatrum:
On the Hex:

image

Maybe we can trim it a bit with the padding?

mutatrum:
Tested with `chroma-js` to get a nice gradient:

image

(Hacked to formatting by deleting header and close button, to make it fit. Picture is just for showing the colors)

duckaxe:
> On the Hex:
>
> Maybe we can trim it a bit with the padding?

Yes, I know it, I tested all scenarios/devices you gave me on discord by changing the `hashrateMonitor.asics` array in `system.service.ts`. As we have limited space, I have added a scroll bar so that the content does not overflow. 

I've already reduced the height compared to just 1 or 2 ASICs. We need the single tile space for the tooltip, and on mobile devices the domain hashrate is displayed inside the tile because a tooltip is not possible. Hex is an edge case.

> Tested with chroma-js to get a nice gradient:

I have no idea how we could play with multiple colors by using only CSS. I wouldn't use JavaScript for this. As with the other issue, I would just wait and see how many people use and understand the heat map before we invest more effort.

mutatrum:
I think this could work (mock-up):
image
Having all domains visible is important, especially if people want to push the limits of their device.

And as for the color in javascript, that currently already happens, just with intensity. Take a look, I think this solution is quite neat, even though code needs a bit more cleanup: https://github.com/mutatrum/ESP-Miner/commit/6f64990d5999a1743879b7796b7532173947838a

duckaxe:
> I think this could work (mock-up):  Having all domains visible is important, especially if people want to push the limits of their device.

As you sad 
> I was thinking about a heatmap. Most devices only have a single chip, or 2 for the GT, so that's not too bad.

I understand that we are optimizing for a maximum of 2 AISCs; anything else is an edge case. Even if you optimize the design for 6 ASICs now, a device with more than 6 AISCs will come.

Your mock-up won't work on a 1200px window width. Even my implementation is tight, but it works.

Screenshot


> 
> And as for the color in javascript, that currently already happens, just with intensity. Take a look, I think this solution is quite neat, even though code needs a bit more cleanup: [mutatrum@6f64990](https://github.com/mutatrum/ESP-Miner/commit/6f64990d5999a1743879b7796b7532173947838a)

396 kB (package) for a color gradient? Not sure its the way ;)

duckaxe:
Another option would be to place the heatmap in its own card. At the bottom right next to the block header. It fits perfectly there and is always visible without toggle the visibility.

**2 AISCs**


**6 ASICs**

mutatrum:
> 396 kB (package) for a color gradient? Not sure its the way ;)

Hmm, according to the website it's about 16kB, maybe our packing is different or I did it wrong, that's possible. But you're right, one (or 2) gradients can be done with some simple JS mathy bits.

mutatrum:
> Another option would be to place the heatmap in its own card. At the bottom right next to the block header. It fits perfectly there and is always visible without toggle the visibility.

I like that idea, especially if we split the temperature and fan cards into two as well, then the bottom card rows both have 3 cards.

duckaxe:
> > 396 kB (package) for a color gradient? Not sure its the way ;)
> 
> Hmm, according to the website it's about 16kB, maybe our packing is different or I did it wrong, that's possible. But you're right, one (or 2) gradients can be done with some simple JS mathy bits.

https://www.npmjs.com/package/chroma-js

duckaxe:
I moved the heatmap from top to bottom into a single card.

From my perspective the heatmap is done. If you like to add some JS magic for gradients go for it.

skot:
This heat map is cool AF, but I don't think it belongs on the default view of the main dashboard tab.

Maybe a detail view that can be expanded?

duckaxe:
@skot Initially, I placed the heat map on the back of the hashrate card, I had to move it for space reasons (see comments). However, I think the current position is just temporary, pending the discovery of a better location. Perhaps we could create an 'Advanced' page or something similar and collect information there that is not relevant for beginners. We could also move 'Block Header' there, as it is not relevant information for newcomers.

skot:
> @skot Initially, I placed the heat map on the back of the hashrate card, I had to move it for space reasons (see comments). However, I think the current position is just temporary, pending the discovery of a better location. Perhaps we could create an 'Advanced' page or something similar and collect information there that is not relevant for beginners. We could also move 'Block Header' there, as it is not relevant information for newcomers.

Advanced tab sounds good to me. Maybe it could even be a sub tab under the dashboard tab.

mutatrum:
I would love to see a Grafana style dashboard, where people can pick and place their panels to show. I think we're indeed getting to the point of saturation of the main dashboard. However, I disagree with moving information to an advanced page. It's prime educational information, curious people will learn about them, and out of sight is out of mind.

F.e. the block header data can be expanded by the actual payout addresses and amounts, I think this is super useful to have on the main page.

0xf0xx0:
> I would love to see a Grafana style dashboard, where people can pick and place their panels to show.

we might wanna look into providing a prometheus endpoint then, and an example grafana dashboard.

skot:
> However, I disagree with moving information to an advanced page. It's prime educational information, curious people will learn about them, and out of sight is out of mind.

We have a very wide range of user skill levels using AxeOS. I think the risk with showing too much on the main dashboard is that it gets overwhelming for new users. Advanced users (or users who want to become advanced) should have no problem finding an advanced tab and digging in.. And a spacious advanced tab could prolly make the more advanced UI elements easier to learn.

shalom7bin:
Thank you so much for giving me strength to become a better program
Can I leave a message here?
I use the gamma gt 800, but the latest version is missing the factory-800
When v2.11 is completed, can I upgrade the firmware?

Issue #1245, opened 2025-09-22
Isolate NVS and Flash write tasks


mutatrum:
From https://github.com/bitaxeorg/ESP-Miner/pull/1217#discussion_r2367274016

Each tasks allocates heap, and by design, this always uses internal memory. It is possible to move this heap off to PSRAM, but only if that task does not write to NVS or Flash. 

Currently NVS is misused as an event bus between tasks, this should be replaced by some other means of inter-task communications, and ideally only a single task should update the NVS store.

One alternative options would be to use the ESP Event handling system. This would also decouple components and reduce the use of GlobalState.

KillerInk:
nvs_config.c 

```
#include "freertos/FreeRTOS.h"
#include "freertos/queue.h"
#include "freertos/task.h"

// ---------------------------------------------------------------------------
//  Queue Item Definition
// ---------------------------------------------------------------------------
typedef enum {
    NVS_ITEM_TYPE_U8,
    NVS_ITEM_TYPE_U16,
    NVS_ITEM_TYPE_I32,
    NVS_ITEM_TYPE_U64,
    NVS_ITEM_TYPE_BOOL,
    NVS_ITEM_TYPE_FLOAT,
    NVS_ITEM_TYPE_STRING
} nvs_item_type_t;

typedef union {
    uint8_t u8;
    uint16_t u16;
    int32_t  i32;
    uint64_t u64;
    bool     b;
    float   f;
    char *  s;          // pointer to dynamically allocated string
} nvs_value_u;

typedef struct {
    const char *key;
    nvs_item_type_t type;
    nvs_value_u val;
} nvs_item_t;

// Global queue handle
static QueueHandle_t nvs_queue = NULL;

// ---------------------------------------------------------------------------
//  Task that processes the queue
// ---------------------------------------------------------------------------
void nvs_write_task(void *pvParameters)
{
    (void) pvParameters;

    nvs_item_t item;
    for (;;) {
        if (xQueueReceive(nvs_queue, &item, portMAX_DELAY) == pdPASS) {
            switch (item.type) {
                case NVS_ITEM_TYPE_U8:
                    // Convert to u16 because no u8 setter
                    nvs_config_set_u16(item.key, (uint16_t)item.val.u8);
                    break;

                case NVS_ITEM_TYPE_U16:
                    nvs_config_set_u16(item.key, item.val.u16);
                    break;

                case NVS_ITEM_TYPE_I32:
[...]

PR #1280, opened 2025-10-18
Misc optimizations to free more internal memory

mutatrum:
* Add allocator for LVGL
 * Enable `SPIRAM_ALLOW_BSS_SEG_EXTERNAL_MEMORY` and `SPIRAM_TRY_ALLOCATE_WIFI_LWIP`.
 * ~~Move bitmaps and font to PSRAM (by @WantClue)~~ WIP, couldn't get it to work yet.
 * Add prebuffer length cache to all http requests;
 * Move several malloc calls in mining loop to stack allocated;
 * Allocate to SPIRAM:
    * AsicTaskModule.active_jobs;
    * hashrate_monitor storage.

Not directly memory related:
 * Eliminated `_send_simple` from `bm1368.c` and `bm1370.c`;
 * `hex2bin` and `bin2hex` use table lookups instead of branching code.

Also added internal and Spiram free memory to system page:

image

### DIFF
--- a/main/screen.c
+++ b/main/screen.c
@@ -331,7 +331,8 @@ static void screen_show(screen_t screen)
         lv_obj_t * scr = screens[screen];
 
         if (scr && lvgl_port_lock(0)) {
-            lv_screen_load_anim(scr, LV_SCR_LOAD_ANIM_MOVE_LEFT, LV_DEF_REFR_PERIOD * 128 / 8, 0, false);
+            bool auto_del = current_screen == SCR_BITAXE_LOGO || current_screen == SCR_OSMU_LOGO;
+            lv_screen_load_anim(scr, LV_SCR_LOAD_ANIM_MOVE_LEFT, LV_DEF_REFR_PERIOD * 128 / 8, 0, auto_del);
             lvgl_port_unlock();
         }
 


mutatrum:
The main fix is probably the http_server prebuffer. I tested a bit with `CONFIG_SPIRAM_USE_CAPS_ALLOC=y` in `sdkconfig.defaults`, which pushes each malloc to go into Spiram, but this locks up the http server immediately. I think this is because the http server accesses NVS, and this doesn't play nice with Spiram allocated memory (See #1245).

Before, the cJSON serializer started with a 256 byte buffer, and re-allocated it roughly twice the size, until it fits. For the info api call, which is ~2.5 kb, this results in 0.25kb + 0.5kb + 1kb + 2kb + 4kb = 7.75 kb, which is really close to the internal free memory limit. It looks like this exhausts the internal memory, so it's either running out of memory or if it decides to allocate in spiram, it locks up due to conflicts with NVS access. Not sure which of the two is happening.

See https://github.com/DaveGamble/cJSON/blob/master/cJSON.c#L532

components/stratum/include/mining.h
```diff
@@ -30,13 +30,13 @@ void free_bm_job(bm_job *job);
...
 
-char *calculate_merkle_root_hash(const char *coinbase_tx, const uint8_t merkle_branches[][32], const int num_merkle_branches);
+void calculate_merkle_root_hash(const char *coinbase_tx, const uint8_t merkle_branches[][32], const int num_merkle_branches, char dest[65]);
 
 bm_job construct_bm_job(mining_notify *params, const char *merkle_root, const uint32_t version_mask, uint32_t difficulty);
 
 double test_nonce_value(const bm_job *job, const uint32_t nonce, const uint32_t rolled_version);
 
-char *extranonce_2_generate(uint64_t extranonce_2, uint32_t length);
+void extranonce_2_generate(uint64_t extranonce_2, uint32_t length, char dest[static length * 2 + 1]);
```
mutatrum, line 39: TIL you can do this.

mutatrum:
Issue for font/bitmap: https://github.com/lvgl/lvgl/issues/9085

PR #1285, opened 2025-10-21
Handle invalid temperature readings

### LINKED ISSUE #1186: random device overheats when rapidly restarting axe
**Describe the bug**
When restarting your axe multiple times (say, its flatlining and refusing to come back up), sometimes itll complain about a device overheat that never happened.

**To Reproduce**
1. Reboot the axe multiple times
2. eventually itll go into overheat mode

**Expected behavior**
The axe shouldnt trip overheat mode.

 - Bitaxe HW version: Gamma 602
 - Bitaxe HW vendor: GekkoScience
 - ESP-Miner FW version: 72be37b0 (v2.10.0b2)
 - Hash Frequency: 550
 - Voltage: 1065

**Additional context**
Unsure how to consistently reproduce this, i tend to get it after restarting from a power fault.

mutatrum:
Fixes #1186

When a temperature sensor cannot be read, the fan is forced to 100%. If the fan speed cannot be adjusted, it reboots the device.

### DIFF
--- a/main/tasks/power_management_task.c
+++ b/main/tasks/power_management_task.c
@@ -184,27 +184,35 @@ void POWER_MANAGEMENT_task(void * pvParameters)
                 // ESP_LOGD(TAG, "DEBUG: PID raw output: %.2f%%, Input: %.1f, SetPoint: %.1f", pid_output, pid_input, pid_setPoint);
 
                 power_management->fan_perc = (uint16_t) pid_output;
-                Thermal_set_fan_percent(&GLOBAL_STATE->DEVICE_CONFIG, pid_output / 100.0);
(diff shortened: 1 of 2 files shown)


WantClue:
This might cause constant reboots on 2pin fans obviously. And it also may create an issue with 3pin fans as well

mutatrum:
> This might cause constant reboots on 2pin fans obviously. And it also may create an issue with 3pin fans as well

Shouldn't, I think the fan controller will just try to set a PWM signal. I doubt the fan controller actually knows the difference between a 2, 3 of 4 pin fan, other than that the RPMs stay 0. But we should indeed verify this.

Issue #272, opened 2024-08-07
Create a general status task


skot:
I'd like to create a general mining status indicator for esp-miner. Currently it's too hard to tell if your Bitaxe is actively mining. This would probably be a separate rtos task that would always show the current status on the Bitaxe display and on the AxeOS dashboard. Maybe this takes the form of the watchdog task?

things to monitor:
- have we received a nonce from the ASIC in the expected interval?
- have we received work from the pool?
- is the network socket still open?
- is wifi connected?
- has the bitaxe overheated?
- what else?

WantClue:
something like this status: O.K. or status: mining that is been used by the standard bitmain firmware I assume. 

Sure doable. 

I think for a general status O.K. to pass we need all the above to be true execpt for the overheat_mode.


change:

- have we received work from the pool ? --> when was the last time we have received work < threshold

skot:
maybe even a little ✅ or ❌ icon for the display in the corner

WantClue:
Or we bring back the status LED like on a control board

PR #1286, opened 2025-10-22
Isolate NVS usage into task

### LINKED ISSUE #1245: Isolate NVS and Flash write tasks

mutatrum:
Delegate all NVS writes to a separate task. This solves the conflict of task heap allocations in Spiram with NVS usage.

To achieve this, the `Settings` struct from `http_server.c` is extracted, and extended to hold all NVS fields. All `#defines NVS_CONFIG_` strings have been refactored into an enum, to enable quick lookups.

All default values are also defined in the settings, and used if NVS is not available or the field is not set yet. This has the benefit that the getters do not need a `default_value` anymore, which in many cases, was duplicated several times throughout the code.

Finally added a nvs_get/set_bool, for ease of use in the code.

Also moved all non-critical task heaps and queues to spiram, which is now possible:

image

This could also help in thinning `GlobalState`, as there are several static fields in there that are only read from NVS  once and are never updated. These can just as well be read directly from nvs_config where they are used.

Fixes #1245

### DIFF
--- a/main/nvs_config.c
+++ b/main/nvs_config.c
@@ -170,7 +170,7 @@ esp_err_t nvs_config_init(void)
 
     err = nvs_open(NVS_CONFIG_NAMESPACE, NVS_READWRITE, &handle);
     if (err != ESP_OK) {
-        vTaskDelete(NULL);
+        ESP_LOGW(TAG, "Could not open nvs");
         return err;
     }
 
@@ -230,7 +230,8 @@ esp_err_t nvs_config_init(void)
 
     TaskHandle_t task_handle;
 
-    BaseType_t task_result = xTaskCreate(nvs_task, "nvs_task", 8192, NULL, 5, &task_handle);
+    // nvs_task heap _must_ be internal memory
+    BaseType_t task_result = xTaskCreate(nvs_task, "nvs_task", 8192, NULL, 5, &task_handle); 
     if (task_result != pdPASS) {
         ESP_LOGE(TAG, "Failed to create nvs_task");
 


components/dns_server/dns_server.c
```diff
@@ -288,7 +288,7 @@ dns_server_handle_t start_dns_server(dns_server_config_t * config)
     handle->num_of_entries = config->num_of_entries;
     memcpy(handle->entry, config->item, config->num_of_entries * sizeof(dns_entry_pair_t));
 
-    xTaskCreate(dns_server_task, "dns_server", 4096, handle, 5, &handle->task);
+    xTaskCreateWithCaps(dns_server_task, "dns_server", 8192, handle, 5, &handle->task, MALLOC_CAP_SPIRAM);
```
shufps, line 291: oh! :star_struck:  didn't know that exists!

shufps: `This solves the conflict of task heap allocations in Spiram with NVS usage.` interesting

mutatrum: > `This solves the conflict of task heap allocations in Spiram with NVS usage.` interesting

See this issue for more info: https://github.com/espressif/esp-idf/issues/11216#issuecomment-1522952737

main/global_state.h
```diff
@@ -67,7 +67,7 @@ typedef struct
     double response_time;
     bool use_fallback_stratum;
     bool is_using_fallback;
-    uint16_t overheat_mode;
+    bool overheat_mode;
     uint16_t power_fault;
```
WantClue, line 71: we should change that also to a bool

WantClue (changes requested):
LGTM testing tomorrow

main/global_state.h
mutatrum: True, but unrelated to this PR, it's not an NVS or REST API field. With this PR and #1290, we can think about #272, which will cover `power_fault` and other status related fields.

PR #1290, opened 2025-10-24
Swarm device notifications

duckaxe:
As an additional feature for all the Swarm modifications of recent weeks, this PR brings service notifications for single Swarm devices. 

This feature is particularly useful for users with multiple devices who monitor their devices via the Swarm page. Currently, they are not notified if something is wrong with a device, as service notifications like power faults, etc., are displayed on the dashboard of the individual device.

This PR brings these service notifications to Swarm. If there is a problem with a device, a small notification appears directly on the device row/card. Since there is very little space, these are small color-based notification messages. 

**List View**
List

**Cards View**
Cards

### DIFF
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
@@ -455,8 +455,10 @@ export class SwarmComponent implements OnInit, OnDestroy {
         return { color: 'red', msg: 'Power Fault' };
       case !axe.frequency || axe.frequency < 400:
         return { color: 'orange', msg: 'Frequency Low' };
-      case axe.isUsingFallbackStratum == 1:
+      case axe.isUsingFallbackStratum === 1:
         return { color: 'orange', msg: 'Fallback Pool' };
+      case axe.blockFound === 1:
+        return { color: 'green', msg: 'Block found' };
       default:
         return undefined;
     }


mutatrum:
I like the feature, bit i think it would make sense to invert the logic, where the polling device sets a string. Swarm shouldn't do any logic on another device, as this will cause future headaches and backwards compatibility issues.

This would also tie in with #272, that would only need to expose one status line, or maybe with a severity flag or so.

duckaxe:
> Swarm shouldn't do any logic on another device, as this will cause future headaches and backwards compatibility issues.

What do you mean? The feature only shows warnings regarding to `info` API data.

mutatrum:
> > Swarm shouldn't do any logic on another device, as this will cause future headaches and backwards compatibility issues.
> 
> What do you mean? The feature only shows warnings regarding to `info` API data.

Say, in the future we have a new flag. Now we have to add conditional logic in swarm, as no teach device polled by swarm will have this from the get go. It's easier if, in case of a new flag, the mining device sets it's status message accordingly, so the swarm client just requests the status string, instead of extending the `getDeviceNotification` in swarm with code that has to handle old and new versions.

So, in short, the `getDeviceNotification` shouldn't be in the swarm code, but on the api side in a `statusMessage` field.

It could be augmented with an enum `status` field (ok/info/warning/error) or so, to set a color. (white/blue/orange/red)

Extra benefit of this is that other clients that gather information can also directly use this, instead of replicating the logic in their code.

duckaxe:
Okay, got it. But until the API gets the new flag, we can still use my PR, right?

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -446,4 +446,19 @@ export class SwarmComponent implements OnInit, OnDestroy {
...
+  getDeviceNotification(axe: any): { color: string; msg: string } | undefined {
+    switch (true) {
+      case axe.overheat_mode === 1:
+        return { color: 'red', msg: 'Overheated' };
+      case !!axe.power_fault:
+        return { color: 'red', msg: 'Power Fault' };
+      case !axe.frequency || axe.frequency < 400:
+        return { color: 'orange', msg: 'Frequency Low' };
+      case axe.isUsingFallbackStratum:
+        return { color: 'orange', msg: 'Fallback Pool' };
```
0xf0xx0: nit: red and orange arent colorblind-friendly, maybe we should look into different icons? flame for overheat, plug for power fault?

duckaxe: The color is secondary. The notification text is more important. Even if users cannot see the color, they can read the text. If we only used the icon and the color, that would be problematic. But if the text is there, the color becomes less important.

I have thought about using icons for each message. However, there are not enough suitable icons. And I don't think it makes sense to replace just two icons. So if we do this, then each notification should have its own icon.

WantClue (approved):
LGTM

PR #1292, opened 2025-10-25
feat: add stratum disconnect on 5min timeout

WantClue:
with this pr we're gonna be able to close the stratum  conn if we don't receive any mining.notify for 5minutes.
This should handle edge cases, where the sock is still open but no actual work will be passed to the miner

### DIFF
--- a/main/tasks/stratum_task.c
+++ b/main/tasks/stratum_task.c
@@ -346,8 +346,8 @@ static bool check_notify_timeout(GlobalState * GLOBAL_STATE, int64_t timeout_sec
 {
     if (last_notify_timestamp > 0) {
         int64_t current_time_ms = esp_timer_get_time() / 1000;
-        int64_t time_diff_ms = current_time_ms - last_notify_timestamp;
-        int64_t time_diff_seconds = time_diff_ms / 1000;
+        int32_t time_diff_ms = current_time_ms - last_notify_timestamp;
+        int32_t time_diff_seconds = time_diff_ms / 1000;
         if (time_diff_seconds > timeout_seconds) {
             ESP_LOGW(TAG, "No mining.notify received for %lld seconds, reconnecting...", time_diff_seconds);
             return true;


main/global_state.h
```diff
@@ -37,6 +37,7 @@ typedef struct
     uint64_t shares_accepted;
     uint64_t shares_rejected;
     uint64_t work_received;
+    int64_t last_notify_timestamp;
```
mutatrum: Minor thing: field is only used in stratum_task, so it can just be a static field in there.

main/tasks/stratum_task.c
```diff
@@ -463,6 +476,12 @@ void stratum_task(void * pvParameters)
         GLOBAL_STATE->abandon_work = 0;
 
         while (1) {
+            if (check_notify_timeout(GLOBAL_STATE, 300)) {
+                retry_attempts++;
+                stratum_close_connection(GLOBAL_STATE);
+                break;
+            }
+
             char * line = STRATUM_V1_receive_jsonrpc_line(GLOBAL_STATE->sock);
```
mutatrum, line 490: if the socket doesn't return anything anymore, doesn't it just hang inside `STRATUM_V1_receive_jsonrpc_line`?

Two avenues, it might need to be done both:
 * Spin up a timer to check the timeout periodically;
 * Set socket timeout, something like:
 `setsockopt (sockfd, SOL_SOCKET, SO_RCVTIMEO, (char *)&timeout, sizeof(timeout);`
 (taken from https://github.com/espressif/esp-idf/issues/8296)

main/tasks/stratum_task.c
```diff
@@ -341,6 +341,19 @@ static void decode_mining_notification(GlobalState * GLOBAL_STATE, const mining_
     }
 }
 
+static bool check_notify_timeout(GlobalState * GLOBAL_STATE, int64_t timeout_seconds)
+{
+    if (GLOBAL_STATE->SYSTEM_MODULE.last_notify_timestamp > 0) {
+        int64_t current_time = esp_timer_get_time();
```
mutatrum: `int32_t current_time_ms = esp_timer_get_time() / 1000;`

Milliseconds are easier to work with, IMO. And the timeout_seconds parameter doesn't need `uint64_t`.

main/tasks/stratum_task.c
```diff
@@ -28,14 +28,15 @@ static StratumApiV1Message stratum_api_v1_message = {};
...
+static int64_t last_notify_timestamp = 0;
 
 struct timeval tcp_snd_timeout = {
     .tv_sec = 5,
     .tv_usec = 0
 };
 
 struct timeval tcp_rcv_timeout = {
-    .tv_sec = 60 * 10,
+    .tv_sec = 180,
```
mutatrum, line 39: Ah, there was a timeout defined, totally missed that. 👍

mutatrum (approved):

main/tasks/stratum_task.c
```diff
@@ -463,14 +481,20 @@ void stratum_task(void * pvParameters)
...
-                ESP_LOGE(TAG, "Failed to receive JSON-RPC line, reconnecting...");
+            if (check_notify_timeout(GLOBAL_STATE, 300)) {
                 retry_attempts++;
                 stratum_close_connection(GLOBAL_STATE);
                 break;
             }
 
+            char * line = STRATUM_V1_receive_jsonrpc_line(GLOBAL_STATE->sock);
+            if (!line) {
+                ESP_LOGD(TAG, "No data received from socket (may be timeout), checking notify timeout...");
```
eandersson, line 492: Why are we no longer stopping the loop here and closing the connection? This means we now wait for up to five minutes before closing, instead of closing immediately when the socket is gone.

PR #1299, opened 2025-10-27
Fix theme color preselect

### LINKED ISSUE #1298: Theme page selected color scheme not selected
The navigating to the theme page, it doesn't select the current color scheme. On full page reload of the theme page it does work.

So, opening the dashboard, navigating to theme. The chosen theme (green) is applied, but there's no checkmark there:

Image

duckaxe:
This PR fixes the problem that the user-defined theme color was not preselected.

https://github.com/user-attachments/assets/016e39be-fed7-408a-a7ae-3dc15c384487

Fixes #1298


mutatrum:
I'm surprised by the amount of code for this fix? Has this never worked before?

duckaxe:
> I'm surprised by the amount of code for this fix? Has this never worked before?

I fixed all deprecated `subscribe` calls.

PR #1304, opened 2025-10-28
Overheat protection

WantClue:
this pr intorduces another layer of protection to the overheat mode. 
It now tracks the last set frequency and if overheat mode get's applied it completely shuts off the power to the asic, power draw in the cool down phase is between 3-4W now. The asic is safely pulled into RST which prevents it from any power draw.

After the asic has reached a safe level of temperature the asic init will be done again with a reduced frequency by a factory of 100. 

It might be useful to adjust the fan curve for the asic cooldown phase in another PR.

This PR also lays the base for the next PR to shutdown the asic via webui.

### DIFF
--- a/main/power/asic_init.c
+++ b/main/power/asic_init.c
@@ -19,15 +19,25 @@ uint8_t asic_initialize(GlobalState *GLOBAL_STATE, asic_init_mode_t mode, uint32
         return 0;
     }
 
-    // CRITICAL: Different handling for cold boot vs recovery
-    if (mode == ASIC_INIT_COLD_BOOT) {
+    // Check actual UART state for safety
+    bool uart_initialized = SERIAL_is_initialized();
+    
+    // Verify mode matches actual state
+    if (mode == ASIC_INIT_COLD_BOOT && uart_initialized) {
+        ESP_LOGW(TAG, "Cold boot mode but UART already initialized - will reset baud only");
+    } else if (mode == ASIC_INIT_RECOVERY && !uart_initialized) {
+        ESP_LOGW(TAG, "Recovery mode but UART not initialized - will do full init");
+    }
+    
+    // Use actual state for decision, not just mode
+    if (!uart_initialized) {
         // Fresh boot - full UART initialization
         ESP_LOGI(TAG, "Performing full UART initialization");
         SERIAL_init();
     } else {
         // Live recovery - ASIC was reset, UART needs baud reset to 115200
         // This preserves the running system and avoids reboot
-        ESP_LOGI(TAG, "Resetting UART to %d baud for ASIC reinitialization", UART_FREQ);
+        ESP_LOGI(TAG, "UART already initialized, resetting baud to %d", UART_FREQ);
         SERIAL_set_baud(UART_FREQ);
         vTaskDelay(100 / portTICK_PERIOD_MS);
     }
--- a/components/asic/include/serial.h
+++ b/components/asic/include/serial.h
@@ -17,5 +17,6 @@ void SERIAL_debug_rx(void);
 int16_t SERIAL_rx(uint8_t *, uint16_t, uint16_t);
 void SERIAL_clear_buffer(void);
 esp_err_t SERIAL_set_baud(int baud);
+bool SERIAL_is_initialized(void);
 
 #endif /* SERIAL_H_ */
--- a/components/asic/serial.c
+++ b/components/asic/serial.c
@@ -41,6 +41,11 @@ esp_err_t SERIAL_init(void)
     return uart_driver_install(UART_NUM_1, BUF_SIZE * 2, BUF_SIZE * 2, 0, NULL, 0);
 }
 
+bool SERIAL_is_initialized(void)
+{
+    return uart_is_driver_installed(UART_NUM_1);
+}
+
 esp_err_t SERIAL_set_baud(int baud)
 {
     ESP_LOGI(TAG, "Changing UART baud to %i", baud);


main/tasks/power_management_task.c
```diff
@@ -131,16 +130,107 @@ void POWER_MANAGEMENT_task(void * pvParameters)
...
+                
+                // Only check ASIC temps if they're valid (not using ASIC thermal diode)
+                if (asic_temp_valid) {
+                    power_management->chip_temp_avg = Thermal_get_chip_temp(GLOBAL_STATE);
+                    power_management->chip_temp2_avg = Thermal_get_chip_temp2(GLOBAL_STATE);
+                    ESP_LOGW(TAG, "Safe mode active (cycle %d) - VR: %.1fC ASIC1: %.1fC ASIC2: %.1fC",
+                             cooling_cycles, power_management->vr_temp, power_management->chip_temp_avg, power_management->chip_temp2_avg);
+                    
+                    // Continue if ASIC temps still too high
+                    if (power_management->chip_temp_avg > THROTTLE_TEMP - 30 || power_management->chip_temp2_avg > THROTTLE_TEMP - 30) {
```
mutatrum: Put 30 in define up top

main/tasks/power_management_task.c
```diff
@@ -131,16 +130,107 @@ void POWER_MANAGEMENT_task(void * pvParameters)
...
+                } else {
+                    // For boards using ASIC thermal diode (600 series), rely on VR temp and time
+                    ESP_LOGW(TAG, "Safe mode active (cycle %d/%d) - VR: %.1fC (ASIC temps unavailable while powered down)",
+                             cooling_cycles, MIN_COOLING_CYCLES, power_management->vr_temp);
+                }
+            }
+            ESP_LOGI(TAG, "Temperature normalized after %d cooling cycles. Reinitializing ASIC...", cooling_cycles);
+            
+            uint16_t reduced_voltage = last_known_asic_volt > 100 ? last_known_asic_volt - 100 : 1000;
+            uint16_t reduced_freq = last_known_asic_freq > 100 ? last_known_asic_freq - 100 : 400;
```
mutatrum: Same with these constants.

main/tasks/power_management_task.c
```diff
@@ -97,6 +99,10 @@ void POWER_MANAGEMENT_task(void * pvParameters)
 
     vTaskDelay(500 / portTICK_PERIOD_MS);
     uint16_t last_core_voltage = 0.0;
+    
+    uint16_t last_known_asic_volt = 0;
+    uint16_t last_known_asic_freq = 0;
```
mutatrum, line 108: The `uint16_t last_known_asic_freq` is not needed, it's only written to NVS for backwards compatibility, in case someone downgrades to pre 2.10.0 (#1051). In code you can let the float be leading, just make sure you also write the int version if needed.

It might be neater to pull this backwards compatibility code into `nvs_config`, but that's for a different PR.

main/tasks/power_management_task.c
```diff
@@ -131,16 +130,107 @@ void POWER_MANAGEMENT_task(void * pvParameters)
...
+            uart_flush(UART_NUM_1);
+            vTaskDelay(100 / portTICK_PERIOD_MS);
+            
+            if (asic_reset() != ESP_OK) {
+                ESP_LOGE(TAG, "ASIC reset failed during reinitialization!");
+                GLOBAL_STATE->SYSTEM_MODULE.asic_status = "ASIC reset failed";
+            } else {
+                // Reset UART to 115200 baud - the ASIC expects this after reset
+                ESP_LOGI(TAG, "Resetting UART to 115200 baud for ASIC reinitialization...");
+                SERIAL_set_baud(115200);
```
mutatrum: Three things:
 * Resetting the baud back to 115200 should be part of the asic reset I think, as I assume that's the point the ASIC starts talking on this baud again?
 * This whole start sequence from `asic_reset` to `ASIC_initalized = true` is mostly the same as the one in `main.c`. Maybe this code can be extracted and re-used.
 * This `115200` should be a #define in `serial.h`, as it's also used on `uart_config`.

mutatrum (commented):
Code level review, haven't run the code yet. Mostly small things.

main/power/asic_init.c
```diff
@@ -0,0 +1,57 @@
...
+    ESP_LOGI(TAG, "Starting ASIC initialization (%s mode)", mode_str);
+
+    if (asic_reset() != ESP_OK) {
+        GLOBAL_STATE->SYSTEM_MODULE.asic_status = "ASIC reset failed";
+        ESP_LOGE(TAG, "ASIC reset failed!");
+        return 0;
+    }
+
+    // CRITICAL: Different handling for cold boot vs recovery
+    if (mode == ASIC_INIT_COLD_BOOT) {
```
mutatrum: Add a `SERIAL_is_initialized` or something that calls `uart_is_driver_installed` to determine this.

`uart_driver_install` is the last thing that `SERIAL_init` does, so should be possible I guess?

And maybe you could internalize that into `SERIAL_init`, just skip everything if uart driver is already installed, but not sure if that's better or not.

main/power/asic_init.c
```diff
@@ -0,0 +1,57 @@
...
+#include "esp_log.h"
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+#include "asic.h"
+#include "serial.h"
+#include "asic_reset.h"
+
+static const char *TAG = "asic_init";
+
+uint8_t asic_initialize(GlobalState *GLOBAL_STATE, asic_init_mode_t mode, uint32_t stabilization_delay_ms)
```
mutatrum, line 11: Not fond of the name, it clashes with the `ASIC_init` naming. But I also don't have a better alternative. it's  a bit of a spider class, going to serial, asic, and asic_init.

Maybe it's better to house it in `system.c`, that's basically the big orchestration class.

And I'm aware that `asic_reset` has the same issue, that's named wrong as well, but we can fix that at a later moment.

WantClue: ye I would leave it like that for now. I'll open up a follow-up issue for the naming schema

mutatrum (approved):

PR #1318, opened 2025-11-03
Fix heatmap on Max and Hex

mutatrum:
Dashboard crashed because the BM1397 doesn't have separate domain hashrates. Furthermore, the gradient calculation was wrong which resulted in a black heatmap on the Hex.

Max:
image

Gamma:
image

Hex (with #1257):
image

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -387,9 +387,11 @@
                     <thead>
                         <tr>
                             <th class="text-left">#</th>
-                            <th *ngIf="getAsicDomainsAmount(info) > 0" [attr.colspan]="getAsicDomainsAmount(info)">
-                                {{ getAsicDomainsAmount(info) == 1 ? "Domain" : "Domains" }}
-                            </th>
+                            <ng-container *ngIf="getAsicDomainsAmount(info) as asicDomainsAmount">
+                                <th *ngIf="asicDomainsAmount > 0" [attr.colspan]="asicDomainsAmount">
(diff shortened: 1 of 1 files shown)


WantClue (approved):

main/http_server/axe-os/src/app/components/home/home.component.html
```diff
@@ -387,12 +387,9 @@ <h4 class="white-space-nowrap overflow-hidden text-overflow-ellipsis">Hashrate R
                     <thead>
                         <tr>
                             <th class="text-left">#</th>
-                            <th *ngIf="getAsicsAmount(info) === 1 && getAsicDomainsAmount(info, 0) === 1; else multiDomain">
-                                Domain
+                            <th *ngIf="getAsicDomainsAmount(info) > 0" [attr.colspan]="getAsicDomainsAmount(info)">
```
duckaxe: To avoid calling the function several times, we should assign it to a variable.

```
<ng-container *ngIf="getAsicDomainsAmount(info) as asicDomainsAmount">
  <th *ngIf="asicDomainsAmount > 0" [attr.colspan]="asicDomainsAmount">
    {{ asicDomainsAmount == 1 ? "Domain" : "Domains" }}
  </th>
</ng-container>

```

duckaxe: Perhaps it would make sense to hide the entire card if the value of getAsicDomainsAmount is 0?

mutatrum: It still has error percentage on the Max. It's not much, but it's something.

And the other cards need to be resized then, so it leaks all over.

duckaxe (approved):

PR #420, opened 2024-10-20
functions to control nonce space and timeouts for all chip topologies

adammwest:
**depends on**
* https://github.com/bitaxeorg/ESP-Miner/pull/1576

**closes**
* https://github.com/skot/ESP-Miner/issues/248

**relavent**
* https://github.com/bitaxeorg/ESP-Miner/issues/514
* https://github.com/bitaxeorg/ESP-Miner/pull/1553
* https://github.com/skot/ESP-Miner/pull/167
* https://github.com/bitaxeorg/ESP-Miner/issues/1507, 
* https://github.com/bitaxeorg/ESP-Miner/issues/939 
* https://github.com/bitaxeorg/ESP-Miner/issues/824

Goals
* to set the nonce space to 100% for configurations that allow
* support extended jobs for SV2

Search Space
there is the nonce space 32 bits
there is the version space 16 bits (BIP320)
there is ntime space ~12 bits (mpt, current time + 7200)
there is extranonce2 space  ~64+bits

General mining info for ASICS
Typically ASICs will mine nonces first, but now they are so fast they have to mine more things,
in terms of cheapness ntime is good and versions are good due to ASICBOOST.

The general hierarchy is
hash boards -> chips -> cores

The older chips (Bitaxe Max), were supplied midstates
BM1397
the bitcoin header is to big to fit in 1 SHA compression so its split into 2

block0     block1
[midstate0][0,2,3,....]
[midstate1][0,2,3,....]
[midstate2][0,2,3,....]
[midstate3][0,2,3,....]
the purpose is share the message scheduler of SHA1 (first sha second block)
but this budens the controller to send work very fast to the chips

anyway there were 4 midstates and 672 small cores
so that means 168 cores on the chip each core does a independant search on the nonce range
core0    [0,2,3,....]
core1    [0,2,3,....] * offset1
...
core167  [0,2,3,....] * offset167

but how do we divide 168 cores into 2^32, we cant so this means there is a hole in the search.
168/256 = ~62.5% of the nonce space is covered due to the HW cores

in more recent BITMAIN chips
BM1370 we have version rolling

with version rolling
we manipulate the nversion field so we have more space to search, we have 2^16 available values
the BM1370 has 128 cores and 2040 small cores, there are 8 cores missing, 4 cores are missing on core 15 and 4 on core 127
these are the version generators

now our pipeline is more advanced
we generate 16 versions and supply then to each core
core 0   [0,1,2,3,....]
...
core 127 [0,1,2,3,....] * offset128
[...]

PR #1321, opened 2025-11-04
Optimize construct_bm_job

mutatrum:
This PR eliminates several inefficient codepaths in constructing new jobs:
 * Limit back and forth reversing of byte or word orders;
 * Remove `merkle_root_be` and `prev_block_hash_be` from `bm_job`;
 * `calculate_coinbase_tx_hash` and `calculate_merkle_root_hash` return hashes instead of hex strings;
 * All parameters are converted to binary once;
 * Stack allocation instead of malloc;
 * Remove unused functions;
 * Unroll endiannness and word reversal functions.

Verified on BM1397 and BM1370.

### DIFF
--- a/components/stratum/test/test_mining.c
+++ b/components/stratum/test/test_mining.c
@@ -99,8 +99,10 @@ TEST_CASE("Validate bm job construction", "[mining]")
     uint8_t expected_midstate_bin[32];
     hex2bin("91DFEA528A9F73683D0D495DD6DD7415E1CA21CB411759E3E05D7D5FF285314D", expected_midstate_bin, 32);
     // bytes are reversed for the midstate on the bm job command packet
-    reverse_endianness_per_word(expected_midstate_bin);
-    TEST_ASSERT_EQUAL_UINT8_ARRAY(expected_midstate_bin, job.midstate, 32);
+    uint8_t expected_midstate_bin_reversed[32];
+    reverse_32bit_words(expected_midstate_bin, expected_midstate_bin_reversed);
+    reverse_endianness_per_word(expected_midstate_bin_reversed);
+    TEST_ASSERT_EQUAL_UINT8_ARRAY(expected_midstate_bin_reversed, job.midstate, 32);
 }
 
 TEST_CASE("Validate version mask incrementing", "[mining]")


components/stratum/mining.c
```diff
@@ -13,94 +13,86 @@ void free_bm_job(bm_job *job)
...
         double_sha256_bin(both_merkles, 64, both_merkles);
     }
 
-    bin2hex(both_merkles, 32, dest, 65);
+    memcpy(dest, both_merkles, 32);
 }
 
 // take a mining_notify struct with ascii hex strings and convert it to a bm_job struct
-bm_job construct_bm_job(mining_notify *params, const char *merkle_root, const uint32_t version_mask, const uint32_t difficulty)
+void construct_bm_job(mining_notify *params, const uint8_t merkle_root[32], const uint32_t version_mask, const uint32_t difficulty, bm_job *new_job)
```
mutatrum, line 49: It might actually be neater to do all the `hex2bin` calls in the stratum_api, instead of here. But that'll be later, this is confusing enough already.

Georges760 (approved):
A very GOOD PR, GG Mutatrum!

Didn't test it on HW but since you did it on both 97 and 70, should be fine, this kind of change easily brake mining, so as long as it mine well, it is fine.

mutatrum:
> A very GOOD PR, GG Mutatrum!

Thanks, it messes with your sanity this stuff. 
 
> Didn't test it on HW but since you did it on both 97 and 70, should be fine, this kind of change easily brake mining, so as long as it mine well, it is fine.

Indeed, if something is wrong, there's three options:
- ASIC won't return any shares;
- Diff verification in firmware fails (everything is diff 0.0);
- Pool rejects the share (on ckpool all rejected with 'Above target').

Georges760:
> > A very GOOD PR, GG Mutatrum!
> 
> Thanks, it messes with your sanity this stuff.
Very true, only those who messed with endiannes in bitcoin structures knows...

skot:
bold PR; thanks for risking your sanity to improve the mining job construction. I'll give this a try on my BM1368 Supra today.

@johnny9 you might be interested in seeing this.

components/stratum/test/test_mining.c
```diff
@@ -65,14 +91,18 @@ TEST_CASE("Validate bm job construction", "[mining]")
...
+    construct_bm_job(&notify_message, merkle_root, 0, 1000, &job);
 
     uint8_t expected_midstate_bin[32];
     hex2bin("91DFEA528A9F73683D0D495DD6DD7415E1CA21CB411759E3E05D7D5FF285314D", expected_midstate_bin, 32);
     // bytes are reversed for the midstate on the bm job command packet
-    reverse_bytes(expected_midstate_bin, 32);
-    TEST_ASSERT_EQUAL_UINT8_ARRAY(expected_midstate_bin, job.midstate, 32);
+    uint8_t expected_midstate_bin_reversed[32];
+    reverse_32bit_words(expected_midstate_bin, expected_midstate_bin_reversed);
+    reverse_endianness_per_word(expected_midstate_bin_reversed);
```
mutatrum, line 104: I don't like this, my suspicion is that the hex string above is not in the same order as it would actually come out of the midstate. In code, the endianness isn't changed, only word order.

mutatrum:
> bold PR; thanks for risking your sanity to improve the mining job construction. I'll give this a try on my BM1368 Supra today.

Paving the way for #420 and dropping the job queues, slowly getting there.

johnny9:
Looks like good cleanup at first glance. I need to take a closer look at the test case changes to be sure it's functionally working well.

mutatrum:
> Looks like good cleanup at first glance. I need to take a closer look at the test case changes to be sure it's functionally working well.

Only the setup of test cases has changed, or expected values are converted to binary, all the assertions in the test cases are the same. The only one I'm not happy with is this one: https://github.com/bitaxeorg/ESP-Miner/pull/1321#discussion_r2495901347

One other thing I'm considering is to move out all the `reverse_32bit_words` for `prev_block_hash`, `merkle_root` and `mid_states` out from `mining.c` and only do that in the `bm13##.c` code, as that's a Bitmain specific thing AFAIK.

The only manipulation happening then is the word endianness swap of prev_block_hash, but that is an actual bitcoin vs stratum difference.

PR #1323, opened 2025-11-04
Add Wi-Fi connect QR code to display

mutatrum:
*Experimental*

This add a QR code to the welcome and the connection screen, so whenever the AP is visible, a Wi-Fi Connect QR is visible on the screen.

image

Display is very small, so not all phones will be able to scan the QR code. I can scan it on a Pixel 6, but you have to zoom in quite a bit and the distance range between the QR code being big enough and the lens still being able to focus is not brilliant. But I did get it to recognize:

image

### DIFF
--- a/main/screen.c
+++ b/main/screen.c
@@ -37,7 +37,6 @@ extern const lv_img_dsc_t osmu_logo;
 static lv_obj_t * screens[MAX_SCREENS];
 static int delays_ms[MAX_SCREENS] = {0, 0, 0, 0, 0, 1000, 3000, 3000, 10000, 10000, 10000, 10000};
 
-static screen_t current_screen = -1;
 static int current_screen_time_ms;
 static int current_screen_delay_ms;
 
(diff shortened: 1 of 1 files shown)


main/lv_conf.h
```diff
@@ -1,6 +1,9 @@
 #ifndef LV_CONF_H
 #define LV_CONF_H
 
+#define LV_USE_LOG 1
+#define LV_LOG_LEVEL LV_LOG_LEVEL_WARN
```
mutatrum, line 5: I enabled logging for LVGL, as it turns out `LV_USE_QRCODE` has some requirements for other features to be enabled. I've been bitten by this before, and having logging helps quite a bit here.

duckaxe:
The QR code is displayed after a reboot if my device is already connected to Wi-Fi. Do we need it there?

mutatrum:
> The QR code is displayed after a reboot if my device is already connected to Wi-Fi. Do we need it there?

It's the same screen as when someone enabled the AP again by holding the boot button. And at initial boot, we don't know yet if the connection will succeed. Not sure how to distinguish these paths.

WantClue:
I think it's a nice addon that might improve the life of some. It's small yes but who cares, and maybe we also add some switches for different behaviour later.

PR #1324, opened 2025-11-04
show ipv4 ipv6 on home component

### LINKED ISSUE #1259: Feature request - IPv6 / IPv4 - Pool connection indicator
With the new option to use an IPv6 connection to the pool, this should be indicated in the dashboard. It could be indicated after the pool name as “(IPv6)” or “(IPv4)”. 

For example, CK-Pool offers both types of access under the general address (solo.ckpool.org). So if you don't use the dedicated accesses (solo4 or solo6), it is not possible to identify which access is being used. 

Addendum:
When logging in to eusolo.ckpool.org, an IPv6 connection is not established, as stated on CK-Pool. It remains an IPv4 connection.
However, I cannot verify which site is at fault here.
IPv6 works on eusolo6.ckpool.org.

WantClue:
adds an indicator to the pool card if ipv4 or ipv6 is being used
fixes #1259

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -324,6 +324,18 @@
                 <div class="flex flex-nowrap justify-content-between mb-3 relative">
                     <h4 class="mb-0">
                         Pool
+                        <a
+                            *ngIf="(quickLink$ | async) as quickLink"
+                            [href]="quickLink"
+                            target="_blank"
+                            rel="noopener noreferrer"
+                            class="ml-1"
+                            [attr.aria-label]="'Open pool dashboard in new tab'"
+                            pTooltip="Open pool dashboard"
+                            tooltipPosition="top"
+                        >
+                            <i class="pi pi-external-link"></i>
+                        </a>
                     </h4>
                     <p-dropdown
                         [options]="(pools$ | async) ?? []"
@@ -340,20 +352,11 @@
                         <span>URL <i class="pi pi-info-circle text-xs px-1" pTooltip="{{ getPoolProtocolType(info) }}" tooltipPosition="top"></i></span>
                     </div>
                     <div>
-                        <a [href]="quickLink$ | async" target="_blank" class="underline break-all">
+                        <span>
                             {{ activePoolURL }}:{{activePoolPort}}
-                        </a>
+                        </span>
                     </div>
                 </div>
-
-                <!--<div class="flex flex-nowrap">
-                    <div class="col-fixed-width text-500">
-                        <span>Port</span>
-                    </div>
-                    <div>
-                        {{ activePoolPort }}
-                    </div>
-                </div>-->
 
                 <div class="flex flex-nowrap ellipsis-magic overflow-hidden" pTooltip="{{activePoolUser}}" tooltipPosition="top">
                     <div class="col-fixed-width text-500">


mutatrum:
It works, but not sure why it's there?

image

Furthermore, most people don't care about it, and IPv6 is the outlier here, so maybe that's the only indicator that's needed, so see if it's actually connected to IPv6?

Mockup, as I don't have IPv6:

image

It's also slightly confusing to have both the Primary indicator and the dropdown, but maybe that's more of an issue with the dropdown: that's the only configuration thing on the whole dashboard. It would make more sense to have that in the pool config page. And IIRC if it's connected to the fallback, there's a mighty big banner at the top.

Alternatively, maybe put both in a tooltip?

WantClue:
mhh ye i think this can be reduced to ipv6 maybe or i just put it into an info box instead of that big visibility

duckaxe:
Or just put the pool IP address in a new line after URL.

duckaxe:
Screenshot

Something like that?

WantClue:
@mutatrum @duckaxe I added a pi-info instead of showing it that big, with that the visible change is reduced and it still can give you information about if it's ipv4 or ipv6

WantClue:
image

WantClue:
@duckaxe I think adding the ip address might not be an ideal solution, the dns name should be enough

duckaxe:
Tooltip is a good solution!

But the way, we need https://github.com/bitaxeorg/ESP-Miner/pull/1175 for easy tooltips.

WantClue:
Let me dig through that, and I'll potentially update my change

mutatrum:
> Screenshot
> Something like that?

I like the URL + port together, but then with the quicklink behind a `:pi-external-link:` instead of underlining only the URL part.

IP is not helpful indeed.

WantClue:
image
how about this @mutatrum

WantClue:
i think thats more beautiful than having everything seperate

duckaxe:
I have often wondered why we link the Stratum URL/Host? Because it is misleading for the user to display an URL (stratum host) and set the link to what others (stats page). For example, if it is a local pool, the link is useless. 

> image

This solution suggests to me I click to the URL and public-pool.io:21496 will open. But that is not the case.

Can't we solve this in another way? Don't link the startum host. Just add an icon or whatever to jump to pool statistics?

duckaxe:
Icon with a link to pool stats.

Screenshot

duckaxe:
Another option (without icon) is to add a pool name and link it to the statistics page.
Screenshot

mutatrum:
> I have often wondered why we link the Stratum URL/Host? Because it is misleading for the user to display an URL (stratum host) and set the link to what others (stats page). For example, if it is a local pool, the link is useless.

A local pool won't show a quick link. It's only a small set of known host names for different pools.

duckaxe:
> > I have often wondered why we link the Stratum URL/Host? Because it is misleading for the user to display an URL (stratum host) and set the link to what others (stats page). For example, if it is a local pool, the link is useless.
> 
> A local pool won't show a quick link. It's only a small set of known host names for different pools.

No, but it links to the IP address, which I can't do anything with.

mutatrum:
> > > I have often wondered why we link the Stratum URL/Host? Because it is misleading for the user to display an URL (stratum host) and set the link to what others (stats page). For example, if it is a local pool, the link is useless.
> > 
> > 
> > A local pool won't show a quick link. It's only a small set of known host names for different pools.
> 
> No, but it links to the IP address, which I can't do anything with.

That's the fall-through:

https://github.com/bitaxeorg/ESP-Miner/blob/2535ad4451eb51e966ab1c4036826939eb679ec3/main/http_server/axe-os/src/app/services/quicklink.service.ts#L47

Not sure if that's correct indeed.

WantClue:
reworking the quicklink service is probably out of scope here. But I'll create an issue for that, so that we only display quick links for known hosts and not everyone

PR #1326, opened 2025-11-05
Disconnect Wi-Fi if IP address is not assigned

### LINKED ISSUE #1315: dhcp none when ip range full
we need to track if the esp actually get's an ip address
becuase it's possible that we connect to an AP but don't get an ip address and we need to make sure granny is aware about that

mutatrum:
Fixes #1315

After connecting to Wi-Fi, waits max 30 seconds before an IPv4 address is assigned. If this doesn't happen, Wi-Fi is disconnected, triggering a reconnect.


mutatrum:
Note: only when an IPv4 address is assigned, the device will go to connected. So it's currently not possible to run this on an IPv6-only network.

WantClue (approved):

PR #1331, opened 2025-11-06
Fix manual fan speed setting

mutatrum:
Bug was caused by trying to read the wrong key format from NVS.

Added `manualFanSpeed` API field to separate current fan speed and manual fan speed configuration.

Also added error logging when using wrong nvs_config_get methods

### DIFF
--- a/main/idf_component.yml
+++ b/main/idf_component.yml
@@ -2,7 +2,7 @@
 dependencies:
   lvgl/lvgl: "9.3.0"
   espressif/esp_lvgl_port: "^2.6.2"
-  esp_lcd_sh1107: "^1"
+  esp_lcd_sh1107: "^1"  
   ## Required IDF version
   idf:
     version: ">=5.5.0"


main/nvs_config.c
```diff
@@ -69,7 +69,7 @@ static Settings settings[NVS_CONFIG_COUNT] = {
     [NVS_CONFIG_DISPLAY_TIMEOUT]                       = {.nvs_key_name = "displayTimeout",  .type = TYPE_I32,   .default_value = {.i32 = -1},                                          .rest_name = "displayTimeout",                     .min = -1, .max = UINT16_MAX},
 
     [NVS_CONFIG_AUTO_FAN_SPEED]                        = {.nvs_key_name = "autofanspeed",    .type = TYPE_BOOL,  .default_value = {.b   = true},                                        .rest_name = "autofanspeed",                       .min = 0,  .max = 1},
-    [NVS_CONFIG_FAN_SPEED]                             = {.nvs_key_name = "fanspeed",        .type = TYPE_U16,   .default_value = {.u16 = 100},                                         .rest_name = "fanspeed",                           .min = 0,  .max = 100},
+    [NVS_CONFIG_MANUAL_FAN_SPEED]                      = {.nvs_key_name = "fanspeed",        .type = TYPE_U16,   .default_value = {.u16 = 100},                                         .rest_name = "manualFanSpeed",                     .min = 0,  .max = 100},
```
mutatrum, line 72: Not sure if I should rename the actual `nvs_key_name` as well. Probabgly not, as that might cause issues when upgrading/downgrading.

ghost:
Tested, re-tested, all good.

main/nvs_config.c
WantClue: Then add a nvs key reader and if it finds the old key rename it in the nvs and reboot the device we can than remove that in the next release

skot:
tested, manual fan speed issue is fixed!

PR #1332, opened 2025-11-06
Fix: Ensure that isUsingFallbackStratum is a number

duckaxe:
As `isUsingFallbackStratum` is a [number](https://github.com/bitaxeorg/ESP-Miner/blob/master/main/http_server/http_server.c#L758), not a boolean, we have to convert it project wide to a number so that it always works as a number, not a boolean.


0xf0xx0:
might be superseded by the v2 api refactor, itll be a proper boolean there (along with the others, like `power_fault` and `overheatMode`)

WantClue:
please resolve conflicts

duckaxe:
@WantClue done

mutatrum:
> might be superseded by the v2 api refactor, itll be a proper boolean there (along with the others, like `power_fault` and `overheatMode`)

It would make sense to make it a number there as well, we're slowly moving towards more than 2 pools. F.e. #1236 adds code to have numbered properties in NVS.

Issue #1310, opened 2025-10-30
Add components and types to system page


mutatrum:
List all i2c devices on the system page, with vendor id and device variant information if available.

skot:
[i2cdetect](https://kernel.googlesource.com/pub/scm/utils/i2c-tools/i2c-tools/+/v3.1.2/tools/i2cdetect.c) on linux does this.. I believe it just sends an i2c read to the whole 7bit address space and looks for an ACK. as far as determining what device is responding, that's a bit harder. afaik there is not standard register for manufacturer ID.

mutatrum:
Relevant: https://github.com/bitaxeorg/ESP-Miner/pull/1333#issuecomment-3500031999

PR #1333, opened 2025-11-06
cleaned up TPS546D24 device_id checking

skot:
cleanup the code that checks the TPS546D24 voltage regulator device ID

### DIFF
--- a/main/power/TPS546.c
+++ b/main/power/TPS546.c
@@ -26,9 +26,9 @@
 
 static const char *TAG = "TPS546";
 
-static uint8_t DEVICE_ID1[] = {0x54, 0x49, 0x54, 0x6B, 0x24, 0x41}; // TPS546D24A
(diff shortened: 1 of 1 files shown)


mutatrum:
Duplicate of #1319 :see_no_evil: 

There's another startup failure condition on the Hex that the old code doesn't handle.

I'm also thinking on how to tackle #1310, so we can expose what devices are on a board, with id markers and manufacturer codes and such (if available), but not sure yet how to do that nicely.

skot:
Ah, nice. yours looks better. Maybe just add the debug print showing which regulator was found?
```
   if (memcmp(id, DEVICE_ID_TPS546D24A, 6) == 0) {
        ESP_LOGI(TAG, "TPS546D24A regulator found");
    } else if (memcmp(id, DEVICE_ID_TPS546D24S, 6) == 0) {
        ESP_LOGI(TAG, "TPS546D24S regulator found");
    } else {
        ESP_LOGE(TAG, "Unknown TPS546D24 regulator device ID");
```

mutatrum:
Duplicate of #1319, added last comment to #1310

PR #1334, opened 2025-11-06
Refactor fallback for asic frequency and fan speed NVS config

mutatrum:
Moved fallback code for `asicfrequency`/`asicfrequency_f` and `fanspeed`/`manualfanspeed` NVS keys into `nvs_config.c`. This cleaned up fallback code in random places and uncovered a small bug in bap code, where only the old `asicfrequency` key was set.

By having this fallback in a single place makes future changes to the NVS config also easier.

### DIFF
--- a/main/nvs_config.c
+++ b/main/nvs_config.c
@@ -28,8 +28,8 @@
     #define FALLBACK_STRATUM_EXTRANONCE_SUBSCRIBE 0
 #endif
 
-#define FALLBACK_KEY_ASICFREQUENCY "asicfrequency"
-#define FALLBACK_KEY_FANSPEED "fanspeed"
+#define FALLBACK_KEY_ASICFREQUENCY "asicfrequency" // Since v2.10.0 (https://github.com/bitaxeorg/ESP-Miner/pull/1051)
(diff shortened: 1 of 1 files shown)


mutatrum:
Code up for review, verified by removing new keys. Migration code only runs once.

main/nvs_config.c
```diff
@@ -110,6 +112,43 @@ Settings *nvs_config_get_settings(NvsConfigKey key)
...
+                nvs_set_str(handle, setting->nvs_key_name, buf);
+            }
+        }
+    }
+    if (key == NVS_CONFIG_MANUAL_FAN_SPEED) {
+        if (nvs_find_key(handle, setting->nvs_key_name, NULL) == ESP_ERR_NVS_NOT_FOUND) {
+            uint16_t val;
+            ret = nvs_get_u16(handle, FALLBACK_KEY_FANSPEED, &val);
+            if (ret == ESP_OK) {
+                ESP_LOGI(TAG, "Migrating NVS config %s to %s (%d)", NVS_CONFIG_MANUAL_FAN_SPEED, setting->nvs_key_name, val);
```
skot, line 135: NVS_CONFIG_MANUAL_FAN_SPEED is not a string

PR #1337, opened 2025-11-08
don't try to printf %s a number in nvs_config_init_fallback()

skot:
nvs_config_init_fallback() was attempting to read from nv before writing the fallback values


skot:
this seemed to fix the problem, but now I'm not so sure why

skot:
ok, that should be the real fix. nvs_config_init_fallback() was trying to ESP_LOG print a enum as a string.

mutatrum (approved):
That was a silly mistake, well spotted!

PR #1348, opened 2025-11-12
Add 1m, 10m and 1h hashrate graph

mutatrum:
Adding 1m, 10m and 1h hashrate averages, replacing the frontend calculated hashrate average. it's possible to add 1d as well, but don't think that's useful.

Most code changes are adding the fields. Added extra was making sure both graph axes are scaled the same if they both show a hashrate. The same could be done for other types, such as temperatures.

image
image
image

Open point is the average efficiency, for that the backend should also calculate average power, but having 3 timescales for that might be overkill. Not sure how to proceed there.


duckaxe:
Please add the new hashrate keys to `openapi.yml`.

duckaxe:
@mutatrum I've played around with the cards a bit so that we can improve clarity of new averages. What do you think?

Screenshot

mutatrum:
Like it, a lot. Maybe slightly more air between error and expected.

duckaxe:
> Like it, a lot. Maybe slightly more air between error and expected.

On smaller screens we have not much space for more air. I will try with % padding to have the space more responsive.

duckaxe:
https://github.com/mutatrum/ESP-Miner/pull/7

mutatrum:
> [mutatrum#7](https://github.com/mutatrum/ESP-Miner/pull/7)

Merged, much appreciated, as always!

duckaxe:
> Open point is the average efficiency, for that the backend should also calculate average power, but having 3 timescales for that might be overkill. Not sure how to proceed there.

This is what it would look like if you wanted to display the efficiency averages. Looks not bad.

Screenshot (1)

duckaxe:
@mutatrum I pushed the changes for https://github.com/bitaxeorg/ESP-Miner/pull/1348#issuecomment-3565815094
In case you would like it. https://github.com/mutatrum/ESP-Miner/commit/6074269c7466b011918ae6640c4349f3806ef073

mutatrum:
> @mutatrum I pushed the changes for [#1348 (comment)](https://github.com/bitaxeorg/ESP-Miner/pull/1348#issuecomment-3565815094) In case you would like it. [mutatrum@6074269](https://github.com/mutatrum/ESP-Miner/commit/6074269c7466b011918ae6640c4349f3806ef073)

Those need to be calculated on the backend as well. Will need to look into your change. Might be for another PR.

duckaxe:
Added hashrate averages to openapi.yaml https://github.com/mutatrum/ESP-Miner/commit/e2ab50cfe05a2770a7f62f4a031df102bdb53a8c
To speed up the PR approval ;)

mutatrum:
> Added hashrate averages to openapi.yaml [mutatrum@e2ab50c](https://github.com/mutatrum/ESP-Miner/commit/e2ab50cfe05a2770a7f62f4a031df102bdb53a8c) To speed up the PR approval ;)

Done

0xf0xx0:
not sure how i feel about this being done on the axe, why can't it be done in axeos? itd be more extensible in the future. average efficiency should be done by axeos as well, any other api consumers can easily calculate it themselves.

mutatrum:
> not sure how i feel about this being done on the axe, why can't it be done in axeos? itd be more extensible in the future. average efficiency should be done by axeos as well, any other api consumers can easily calculate it themselves.

The current front-end average hashrate value is rather hit-or-miss. Doing this in the background means it's always available, even if the dashboard is not opened, and also available f.e. on swarm, or on other front-ends.

PR #1351, opened 2025-11-15
Fix statistics logging period

### LINKED ISSUE #1342: Dashboard data logging error
In the system settings, I entered “Every 14 minutes for 1 week” for data logging.
However, only the last 12 hours are displayed in the dashboard (displayed every minute). Older entries disappear.

Bitaxe Gamma 601
Software from November 7 (db1b123) - Scoreboard. 

Image

Image

Image

terratec:
Fixes #1342 

The data logging setting is stored in seconds (in NVS u16) but the calculations are performed in milliseconds.

### DIFF
--- a/main/tasks/statistics_task.c
+++ b/main/tasks/statistics_task.c
@@ -23,7 +23,6 @@ static uint16_t statisticsDataSize;
 static pthread_mutex_t statisticsDataLock = PTHREAD_MUTEX_INITIALIZER;
 
 static const uint16_t maxDataCount = 720;
-static uint32_t statsFrequency;
 
 void createStatisticsBuffer()
 {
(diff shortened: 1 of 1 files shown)


mutatrum:
Testing now.

main/tasks/statistics_task.c
```diff
@@ -23,7 +23,7 @@ static uint16_t statisticsDataSize;
 static pthread_mutex_t statisticsDataLock = PTHREAD_MUTEX_INITIALIZER;
 
 static const uint16_t maxDataCount = 720;
-static uint16_t statsFrequency;
+static uint32_t statsFrequency;
```
mutatrum: This doesn't need to be a static field, as it's only used in the while loop. Might as well just define it there.

Maybe for clarity, get the value from NVS into a `uint16_t`, but then do the `* 1000` into a `uint32_t`.

mutatrum:
Interesting fix, as the uint16_t will overflow and wrap around. So what you'd get is something random between 0 and 65 seconds.

terratec:
> Interesting fix, as the uint16_t will overflow and wrap around. So what you'd get is something random between 0 and 65 seconds.

Yeah I was surprised when I stumbled across this variance in the API data.

mutatrum (approved):

PR #1357, opened 2025-11-16
Fan controller task

mutatrum:
This is a part of #1217 where only the fan controller is extracted. It increased the PID controller frequency from 1.8 seconds to 100ms, which results in a much smoother fan curve. Especially on older devices that uses the EMC2101 internal temp with a 1° C resolution, it's a much more restful soundscape.

I also decreased the PID tuning values quite a bit, but this needs further testing. Maybe the PID parameters need to be exposed under an advanced tab, so people can tune the controller themselves.

### DIFF
--- a/main/tasks/fan_controller_task.c
+++ b/main/tasks/fan_controller_task.c
@@ -81,6 +81,9 @@ void FAN_CONTROLLER_task(void * pvParameters)
                     }
                     
                     pid_compute(&pid);
+                    // Clamp PID output to valid range to prevent overshoot above 100%
+                    if (pid_output > 100) pid_output = 100;
+                    if (pid_output < 0) pid_output = 0;
                     // Uncomment for debugging PID output directly after compute
                     // ESP_LOGD(TAG, "DEBUG: PID raw output: %.2f%%, Input: %.1f, SetPoint: %.1f", pid_output, pid_input, pid_setPoint);
 
@@ -107,6 +110,7 @@ void FAN_CONTROLLER_task(void * pvParameters)
                 }
             } else { // Manual fan speed
                 uint16_t fan_perc = nvs_config_get_u16(NVS_CONFIG_MANUAL_FAN_SPEED);
+                if (fan_perc > 100) fan_perc = 100;
                 if (fabs(power_management->fan_perc - fan_perc) > EPSILON) {
                     power_management->fan_perc = fan_perc;
                     if (Thermal_set_fan_percent(&GLOBAL_STATE->DEVICE_CONFIG, fan_perc / 100.0f) != ESP_OK) {


0xf0xx0:
auto fan control isnt capped at 100, and will slowly over an hour or two go up to 255 before wrapping around and re-starting the fan

image

main/tasks/fan_controller_task.c
```diff
@@ -0,0 +1,124 @@
...
+                        ESP_LOGI(TAG, "Temperature sensor starting up, setting fan to 70%%");
+                        power_management->fan_perc = 70;
+                        if (Thermal_set_fan_percent(&GLOBAL_STATE->DEVICE_CONFIG, 0.7f) != ESP_OK) {
+                            exit(EXIT_FAILURE);
+                        }
+                    }
+                }
+            } else { // Manual fan speed
+                uint16_t fan_perc = nvs_config_get_u16(NVS_CONFIG_MANUAL_FAN_SPEED);
+                if (fabs(power_management->fan_perc - fan_perc) > EPSILON) {
```
WantClue, line 114: if (fan_perc > 100) {
  fan_perc=100;
}

I would add this for out of boundry safety

WantClue: before the if(fabs(...

main/main.c
```diff
@@ -77,6 +78,9 @@ void app_main(void)
     if (xTaskCreate(POWER_MANAGEMENT_task, "power management", 8192, (void *) &GLOBAL_STATE, 10, NULL) != pdPASS) {
         ESP_LOGE(TAG, "Error creating power management task");
     }
+    if (xTaskCreate(FAN_CONTROLLER_task, "fan_controller", 8192, (void *) &GLOBAL_STATE, 5, NULL) != pdPASS) {
+        ESP_LOGE(TAG, "Error creating fan controller task");
+    }
```
WantClue, line 85: we might wanna halt the execution for this and other tasks, if the creation of a task fails; 
+ return;

mutatrum:
> auto fan control isnt capped at 100, and will slowly over an hour or two go up to 255 before wrapping around and re-starting the fan

Apparently the PID max output is a soft limit and it can overshoot, reproduced it by setting the temp to 35C. Added clamping of the PID output.

main/tasks/fan_controller_task.c
mutatrum: Done

main/main.c
mutatrum: More for #272, IMO.

main/tasks/fan_controller_task.c
mutatrum: Also #1539

PR #1364, opened 2025-11-17
Prevent precision artifacts with floats in REST API

mutatrum:
cJSON has no support for floats. So when `cJSON_CreateNumber` or `cJSON_AddNumberToObject` is called with a float, it's upcast to double. However, this introduces precision artifacts. This PR adds two helper methods that make sure the upcast double is limited to 7 decimals, which limits superfluous digits. Upside is smaller JSON, especially for the statistics.

An example:

Before: `"power": 17.9117431640625,`

After: `"power": 18.4124756,`

This is done for all floats. In some cases it cases very annoying artifacts, f.e. the float value `0.1` ends up as `0.10000000149011612` in the JSON.

I'm not super happy with the code, as it's easy to forget about this with new fields, but not sure how to make that better.


WantClue (approved):
🥦

PR #1366, opened 2025-11-18
Wake screen briefly on button press or non-carousel screens

mutatrum:
Small QOL enhancement for people who turn off the screen. Press boot button to briefly wake up. It also shows the screen on the pre-carousel screens, f.e. the AP enable screen, overheat or ASIC failure.


terratec:
With this change there is no longer a difference between "always off" and "timeout" and the "always off" option could be removed?

I would argue that this is intended for users who want to remove the screen but cannot (because it is soldered on)
and the Bitaxe should respect the users choice. Everyone else can use the smallest timeout config.

mutatrum:
> With this change there is no longer a difference between "always off" and "timeout" and the "always off" option could be removed?
> 
> I would argue that this is intended for users who want to remove the screen but cannot (because it is soldered on) and the Bitaxe should respect the users choice. Everyone else can use the smallest timeout config.

For that you can select None on Display Type. That will not even initialize the display hardware.

terratec:
> > With this change there is no longer a difference between "always off" and "timeout" and the "always off" option could be removed?
> > I would argue that this is intended for users who want to remove the screen but cannot (because it is soldered on) and the Bitaxe should respect the users choice. Everyone else can use the smallest timeout config.
> 
> For that you can select None on Display Type. That will not even initialize the display hardware.

Ah, perfect!
Should the label "Always off" be renamed in the frontend?
Or can the specific code with the hidden control be simplified if we add a 5 second timeout?

PR #1373, opened 2025-11-19
Reverse proxy support

tbshfr:
This allows the web ui / API to be accessed  behind a reverse proxy with https. It also enables access by hostname (when the hostname resolves to a private IP address via LAN/DNS. This change does not add mDNS support). 

It uses the standard X-Forwarded-For header behind a reverse proxy. 
This header can be spoofed but this must be handled by the proxy (validating or stripping), not here. 
The reverse proxy (immediate peer) must be on a private network and the client IP (from X-Forwarded-For) must also be from a private network.
This enables typical setups where https terminates at the proxy and the proxy forwards requests to the device over the local network. I tested it behind traefik and nginx.

Unfortunately i cant test swarm mode, but from what i saw in other PRs, I dont think swarm mode will work behind a reverse proxy if other devices are on an older firmware version.
ipv6 only proxies are not supported but ipv6 only networks are also not supported currently, so this should not be a regression.

### DIFF
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.html
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.html
@@ -3,6 +3,17 @@
 <div class="card">
     <h2>Swarm</h2>
 
+    <p-message *ngIf="!canScan && protocol === 'http:'" severity="warn" styleClass="w-full mb-3 py-4 border-round-xl"
+        text="Network scanning is disabled when accessed via hostname or reverse proxy. Add devices manually using their IP address or hostname.">
+    </p-message>
+
+    <p-message *ngIf="!canScan && protocol === 'https:'"
+        severity="warn"
+        styleClass="w-full mb-3 py-4 border-round-xl"
+        text="Network scanning is disabled when accessed via hostname or reverse proxy. Add devices manually using their hostname.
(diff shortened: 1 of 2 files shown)


tbshfr:
Just noticed that this also breaks the swarm view for a single bitaxe - it no longer detects itself. I will look into how to fix this

tbshfr:
The swarm page should now work again, with some limitations:
- older device versions may not be addable if the ui is accessed via hostname or behind a proxy (ip should still work, if someone could test this would be great)
- if its accessed by hostname or proxy, auto scanning is disabled
- if the ui is accessed over https, you can only add other devices to the swarm if they are also accessible over https (mixed content blocking), I added warnings for this in the ui

The swarm page still needs some cleanup. I think it would be best to only allow adding devices via ip addresses when the ui is accessed by ip to keep the old behavior and avoid confusion. 

Would appreciate quick feedback / input before I proceed with cleanup.

mutatrum:
We currently have #1240, which is planned for v2.13. If that is added, is this PR then still valid?

tbshfr:
According to the last comment it only covers .local, so this wont work behind a reverse proxy with a subdomain on a normal domain. I will close this PR and wait until the other one is merged, if this is still relevant I will open a new one based on the other PR.

PR #1385, opened 2025-11-24
Show notification on dashboard when device is unreachable

duckaxe:
This pull request replaces an outdated one: https://github.com/bitaxeorg/ESP-Miner/pull/737

The objective: If the device does not respond, polling of the `system/info` endpoint must continue. Additionally, AxeOS should display a notification to alert the user that the device has become unavailable. Once the device becomes available again, the notification will disappear and AxeOS will resume normal operation.

Technically speaking, `catchError` captures the error. If an error occurs, a notification is displayed and the browser title changes. Once the endpoint request has been completed without error, the error is deleted.

Screenshot (1)


main/http_server/axe-os/src/app/pipes/date-ago.pipe.ts
```diff
@@ -7,8 +7,8 @@ import { Pipe, PipeTransform } from '@angular/core';
 export class DateAgoPipe implements PipeTransform {
   private static _this = new DateAgoPipe();
 
-  public static transform(value: number): string {
-    return this._this.transform(value);
+  public static transform(value: number, args?: any): string {
+    return this._this.transform(value, args);
```
mutatrum, line 11: 👍  Nice, I've spent some time wondering why args didn't work when this pass-through wasn't there on other pipes.

mutatrum:
Does this stop the addition of new data to the graph and updating of fields on the dashboard? As I struggled to get the loop to not die when this happened.

duckaxe:
> Does this stop the addition of new data to the graph and updating of fields on the dashboard? As I struggled to get the loop to not die when this happened.

The loop is still active. You can see that well in the screenshot (there is no page reload). Once the miner is available again, collected data will be added once more.

Screenshot (1)

WantClue:
@duckaxe please kindly resolve the conflicts

duckaxe:
@WantClue Solved.

WantClue (approved):

PR #1389, opened 2025-11-24
Don't fail SYSTEM_init_peripherals on invalid VCORE value

### LINKED ISSUE #1382: v2.11: Invalid VCORE config in NVS causes SYSTEM_init_peripherals failure and effectively soft-bricks device
After upgrading a Bitaxe Gamma 601 from **v2.10.x** to **v2.11.0**, an existing config with **VCORE = 0.900 V** causes the miner to fail during peripheral init and never reach a usable state (no UI, no hashing). The same hardware and config works on **v2.10.x**.

This looks like an unintended side effect of the new behavior in v2.11 where `SYSTEM_init_peripherals()` failure is treated as fatal in `app_main()`.

### Environment

* **Board:** Bitaxe Gamma 601
* **Firmware:**

  * Previously OK on 2.10.x
  * Problem appears after upgrade to 2.11.0
* **Config:** ASIC VCORE set to **0.900 V** (saved in NVS from earlier firmware)

---

### Symptoms

On v2.11.0, the serial log shows:

```text
I (...) vcore: Set ASIC voltage = 0.900V
E (...) TPS546: Voltage requested (0.900000 V) is out of range
E (...) vcore: VCORE_set_voltage(...): TPS546 set voltage failed!
E (...) system: SYSTEM_init_peripherals(...): VCORE set voltage failed!
E (...) bitaxe: Failed to init peripherals
```

After this:

* The device does **not** bring up AxeOS (no screen UI / web UI).
* It appears “dead” or stuck, and there is no way to correct the bad VCORE via the normal interface.

Downgrading to **2.10.x** on the same hardware with the same stored config **does** boot to a usable state again.

---

### Likely cause in 2.11

In 2.11, two changes seem relevant:

1. Extended TPS546 detection / startup failure handling.
2. `app_main()` now explicitly checks the return value of `SYSTEM_init_peripherals()` and logs `"Failed to init peripherals"` then returns if it is not `ESP_OK`.

This means a previously “just bad but survivable” VCORE config (e.g. 0.900 V) now causes the whole app to abort at startup, leaving the user with no way to fix the config without low-level intervention.

---

### Expected vs actual

**Expected:**

* If VCORE in NVS is out of range for the board:

  * Either clamp it to a safe default and continue, or
  * Boot in a “safe mode” where hashing is disabled but the UI/API is available so the user can correct settings.

**Actual in 2.11.0:**

* Out-of-range VCORE causes `VCORE_set_voltage` → `SYSTEM_init_peripherals` to fail.
* `app_main()` treats this as fatal and exits early.
* Device never reaches a usable UI/API state; from a user’s perspective it looks soft-bricked after the upgrade.

---

### Suggestions

A few possible mitigations:

1. **Clamp VCORE on load:**
   When reading VCORE from NVS, clamp values outside the board’s allowed range to `[MIN, MAX]` and log a warning instead of failing init.

2. **Safe-mode fallback on init failure:**
   If `SYSTEM_init_peripherals()` fails due to VCORE/TPS546 issues, reset VCORE to a default and retry once, or boot a minimal mode with ASIC disabled but UI/API available.

3. **Validate VCORE on write:**
   Reject out-of-range VCORE values in the UI/API/config writer instead of allowing them to be stored and only failing at next boot.

---

* Is this fatal-on-`SYSTEM_init_peripherals` behavior in 2.11 intended even for purely config-driven issues like invalid VCORE?
* Would it be acceptable to provide a safe-mode path so that an out-of-range user setting cannot render the device effectively unusable after an upgrade?

mutatrum:
This reverts commit 2535ad4451eb51e966ab1c4036826939eb679ec3.

Fixes #1382

### DIFF
--- a/main/system.c
+++ b/main/system.c
@@ -105,7 +105,6 @@ esp_err_t SYSTEM_init_peripherals(GlobalState * GLOBAL_STATE) {
 
     // Initialize the core voltage regulator
     ESP_RETURN_ON_ERROR(VCORE_init(GLOBAL_STATE), TAG, "VCORE init failed!");
-    ESP_RETURN_ON_ERROR(VCORE_set_voltage(GLOBAL_STATE, nvs_config_get_u16(NVS_CONFIG_ASIC_VOLTAGE) / 1000.0), TAG, "VCORE set voltage failed!");
 
     ESP_RETURN_ON_ERROR(Thermal_init(&GLOBAL_STATE->DEVICE_CONFIG), TAG, "Thermal init failed!");
 


mutatrum:
From #1382:

> If I now look at the code of SYSTEM_init_peripherals, it exits on every error, so depending on what goes wrong, only a part of the peripherals are initialized, hence the idea that it's a fatal error. So if the VCORE value is invalid, the temp sensor, display and buttons are not initialised, but it will continue to boot.
> 
> Ideally, it should try to at least initialism all peripherals, and if that works, set the VCORE (and maybe other configurations). Secondary, if VCORE (and maybe other NVS configuration values) are out of spec, they should be clamped to valid values.

PR #1391, opened 2025-11-25
Coinbase transaction parser

mutatrum:
This PR adds parsing of the coinbase transaction outputs. Needs testing on as many pools as possible.

It incorporates `segwit_address.c` from pwuille and `base58.c` from luke-jr. These are a big part of all the new code (558 lines). They are copied without any changes.

One limitation is that the network (mainnet/testnet/signet) is not known, so if you mine on testnet/signet, the addresses are shown as mainnet. Suggestions welcome here.

The  are shown on the dashboard, if the address matches the address part of the pool user, it's printed bold with a star:
image

It also works with the `sensitive-data` flag:
image

If the payout is above 95%, it shows the pool fee, otherwise it will show the share percentage and a warning banner

Example log from solo ckpool (address masked, first output is the same as my pool address:
```
I (16952) stratum_task: Coinbase outputs: 3, total value: 315205815 sat
I (16962) stratum_task:   Output 0: 312253878 sat -> bc1q**************************************  (Your payout address)
I (16971) stratum_task:   Output 1: 6372528 sat -> bc1q**************************************
I (16980) stratum_task:   Output 2: 0 sat -> OP_RETURN: .!.._..Vv0......d...rqz..S..6.*D..k{
```
<hr>
For the dashboard, the address abbreviation has changed to split the address in 4 segments, removing middle segments until it falls below a certain length. This was done after review by the bitcoin design community, which you can read the thread in the Bitcoin Design Discord at https://discord.com/channels/903125802726596648/1443584935305678878.

We came to this mechanism because that's also usable on the display and yields the best results for verification for all address types:
image


0xf0xx0:
idea for the network detection, why not use the active pools mining address? assuming the user is actually configured correctly, their address will reveal which network the pool is on.

WantClue:
@mutatrum please resolve conflicts

shufps:
love that! I was thinking about adding some solo-mining verification that checks if all coins go to the own address - like I had on the piaxe python code but it seems you are almost finished :raised_hands:

mutatrum:
> love that! I was thinking about adding some solo-mining verification that checks if all coins go to the own address - like I had on the piaxe python code but it seems you are almost finished 🙌

Need to think how to show this on the front-end. Especially with pools like OCEAN that can have dozens of payout addresses.

shufps:
> > love that! I was thinking about adding some solo-mining verification that checks if all coins go to the own address - like I had on the piaxe python code but it seems you are almost finished 🙌
> 
> Need to think how to show this on the front-end. Especially with pools like OCEAN that can have dozens of payout addresses.

IMHO a real verficiation would need a % field that allows a certain value to not be sent to the own address because ckpool is solo mining with 2% fee.

A more easy approach that would work without any settings (besides a "solo verification" checkbox) would be to check if the address is amongst the payout addresses and maybe just accept it on >95% or so.

A little bit of a safety net, to prevent the biggest scams.

*edit*: hmm ... or a checkbox on the stratum settings with just "Solo Pool" and enable a slight check automatically. This could be tuned later but just to make sure it's not complete scam.

I would like that :smiling_face_with_three_hearts: 

IMHO this is important because every solo pool could steal some hashrate by just letting it mine on its own address now and then and nobody would ever notice

johnny5OO (approved):

johnny5OO (approved):

WantClue:
please resolve conflicts

mutatrum:
> please resolve conflicts

Resolved. Only thing I can't get to work is the sensitive-data attribute on the payout address.

duckaxe:
Please add demo data to system.service.ts

mutatrum:
Tested on Ocean:

image

~There is a limit of 100 coinbase tx outputs.~ There have been blocks with more than 100, even close to [1000](https://mempool.space/block/133471), but that's not recent.

Edit: I've reduced the number of decoded tx outputs to 6. This should cover most solo pools. If there are more outputs, they are still checked if it's the user's address, but they are not sent to the front-end.

mutatrum:
Also fixes #1446

kmk142789 (approved):

WantClue:
will be reviewed today

mutatrum:
> will be reviewed today

Looking forward to it. Take note of this:

> It incorporates segwit_address.c from pwuille and base58.c from luke-jr. These are a big part of all the new code (558 lines). They are copied without any changes.

That'll save you some reviewing. I think that code is batlle-tested.

main/http_server/openapi.yaml
```diff
@@ -435,6 +438,28 @@ components:
...
+              - value
+              - address
+            properties:
+              value:
+                type: integer
+                description: Output value in satoshis
+              address:
+                type: string
+                description: Output address
+        coinbaseValueTotalSatoshis:
```
WantClue, line 457: the indent doesn't look right

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -634,6 +642,9 @@ export class HomeComponent implements OnInit, OnDestroy {
     updateMessage(!info.frequency || info.frequency < 400, 'FREQUENCY_LOW', 'warn', 'Device frequency is set low - See settings');
     updateMessage(!!info.isUsingFallbackStratum, 'FALLBACK_STRATUM', 'warn', 'Using fallback pool - Share stats reset. Check Pool Settings and / or reboot Device.');
     updateMessage(info.version !== info.axeOSVersion, 'VERSION_MISMATCH', 'warn', `Firmware (${info.version}) and AxeOS (${info.axeOSVersion}) versions do not match. Please make sure to update both www.bin and esp-miner.bin.`);
+    let percentage = this.getPayoutPercentage(info);
+    updateMessage(percentage > 0 && percentage < 95, 'NOT_SOLO_MINING', 'warn', `Your share of the coinbase reward is only ${percentage.toFixed(1)}%`);
+    updateMessage(percentage === 0, 'NOT_SOLO_MINING', 'warn', `You don't have a share in the coinbase reward`);
```
WantClue: Bug: Second updateMessage call removes the message added by the first.

The warning will never display for users with a payout share between 0% and 95%.


    const notSoloCondition = percentage >= 0 && percentage < 95;
    const notSoloText = percentage === 0
      ? `You don't have a share in the coinbase reward`
      : `Your share of the coinbase reward is only ${percentage.toFixed(1)}%`;
    updateMessage(notSoloCondition, 'NOT_SOLO_MINING', 'warn', notSoloText);

main/screen.c
```diff
@@ -559,7 +559,7 @@ static void screen_update_cb(lv_timer_t * timer)
         lv_label_set_text_fmt(mining_network_difficulty_label, "Difficulty: %s", GLOBAL_STATE->network_diff_string);
     }
 
-    if (GLOBAL_STATE->scriptsig != NULL && strcmp(lv_label_get_text(mining_scriptsig_label), GLOBAL_STATE->scriptsig) != 0) {
+    if (strcmp(lv_label_get_text(mining_scriptsig_label), GLOBAL_STATE->scriptsig) != 0) {
```
WantClue, line 562: NULL pointer dereference: strcmp() with potentially NULL scriptsig is undefined behavior.

restore the NULL check

main/tasks/stratum_task.c
```diff
@@ -337,78 +334,68 @@ void stratum_primary_heartbeat(void * pvParameters)
...
-
-    if (GLOBAL_STATE->scriptsig == NULL || strcmp(scriptsig, GLOBAL_STATE->scriptsig) != 0) {
-        ESP_LOGI(TAG, "Scriptsig: %s", scriptsig);
-
-        char * previous_miner_tag = GLOBAL_STATE->scriptsig;
-        GLOBAL_STATE->scriptsig = scriptsig;
-        free(previous_miner_tag);
-    } else {
-        free(scriptsig);
+    // Update coinbase outputs
```
WantClue, line 376: memcmp and memcpy use result->output_count to determine the size, but there's no validation that result->output_count doesn't exceed the capacity of GLOBAL_STATE->coinbase_outputs. If coinbase_process_notification returns more outputs than the destination array can hold, this will cause memory corruption.

main/http_server/axe-os/src/app/components/home/home.component.ts
mutatrum: Changed the type of the 2nd one, simpler solution. Well spotted.

main/screen.c
mutatrum: scriptsig is a char[128] now, so it's never null.

main/tasks/stratum_task.c
mutatrum: Added safeguard

PR #1393, opened 2025-11-25
Add HttpClient default timeout

duckaxe:
On almost all AxeOS views, we use data pooling with a 5-second interval. Angular's `HttpClient` does not have a default timeout, which is not suitable for our purposes as we repeat a request after five seconds.

This PR sets a global timeout of 5 seconds. This value can be overwritten for any request if needed.


duckaxe:
Closed. The OTA update requires more than 5s. Needs a better solution.

PR #1395, opened 2025-11-26
Bitaxe SupraHex 701/702

mutatrum:
- Add configurations for the SupraHex 701/702
- Fiix hashrate register asic_nr for BM1368
- Add missing board configs (400, 302)
- Add factory images for all available boards

Many thanks to @eduionita for testing and verification!

image

image

image


WantClue:
awesome thanks, we need some more tester on this one I guess

PR #1396, opened 2025-11-27
Add timeout to GET requests

duckaxe:
On almost all AxeOS views, we use data pooling with a 5-second interval. Angular's HttpClient does not have a default timeout, which is not suitable for our purposes as we repeat a request after five seconds.

This PR adds a timeout of 5 seconds to GET requests.


mutatrum:
Can be done generally with an Interceptor? I saw a fairly old SO post: https://stackoverflow.com/a/49562579 No idea if this is still valid.

duckaxe:
I already created and closed a PR for a default timeout https://github.com/bitaxeorg/ESP-Miner/pull/1393

I would only set a timeout for GET requests, as it is impossible to predict how long POST and PATCH requests might take. Furthermore, it is not a problem if a POST request takes a while. However, if we pool GET requests every five seconds, then a timeout would make sense.

WantClue:
> I already created and closed a PR for a default timeout #1393
> 
> I would only set a timeout for GET requests, as it is impossible to predict how long POST and PATCH requests might take. Furthermore, it is not a problem if a POST request takes a while. However, if we pool GET requests every five seconds, then a timeout would make sense.

I do agree, at some point we just want to ignore a GET if it takes too long

PR #1398, opened 2025-11-27
Fix: Decrease z-index of loading overlay

duckaxe:
In some cases, the loading overlay and toaster notifications may be displayed simultaneously. Currently, notifications are masked by the layer. This PR ensures that the layer is assigned a lower (z-)index.

**Before**
Before

**After**
After


main/http_server/axe-os/src/app/components/loading/loading.component.scss
```diff
@@ -1,10 +1,9 @@
 #loading {
     background-color: var(--maskbg);
-    left: calc(4rem + var(--sidebar-width));
```
duckaxe, line 3: Removed because duplicate.

PR #1403, opened 2025-11-27
Identify Device moved to screen overlay

mutatrum:
Moving the `Hi!` to the overlay reduces complexity, as it was biting with the connection screen.

Also added the option to cancel the `Hi!` screen by pressing the Idenity button again on the System page, or by pressing the boot button.

### DIFF
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.html
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.html
@@ -133,7 +133,7 @@
                                     severity="secondary"
                                     class="inline-block"
                                     styleClass="button-icon-small"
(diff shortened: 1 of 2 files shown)


mutatrum:
> Thank you. Awesome code. Tested on my device.

I was playing QA monkey, and pressed the button several times in a row. That gave some interesting behavior, as with each button click, a new timer was added while the previous timers were also running. So I had a very quick blinking `Hi!` screen as each subsequent timer ran out. When trying to fix that, and trying to cancel the `Hi!` with the boot button, I got into trouble with the carousel and the connectivity screen. So the change grew slightly bigger than I anticipated.

PR #1404, opened 2025-11-29
Measure Job Submit Latency

mutatrum:
Adds logging:
```
I (70779) asic_result: Share submit latency: 1095 µs
```

Also moved some static fields to stack allocation and limited the scope of stratum latency measurement (not counting JSON parsing).


mutatrum:
Parts are implemented in #1422

PR #1407, opened 2025-11-29
Add chance indicator on dashboard

WantClue:
image

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.ts
+++ b/main/http_server/axe-os/src/app/components/home/home.component.ts
@@ -630,7 +630,7 @@ export class HomeComponent implements OnInit, OnDestroy {
     if (!info.networkDifficulty || info.networkDifficulty === 0) return '0';
     const percentage = (info.bestDiff / info.networkDifficulty) * 100;
     // Show 2 significant digits
-    return percentage.toPrecision(2);
+    return percentage < 10 ? percentage.toPrecision(2) : percentage.toFixed(1);
   }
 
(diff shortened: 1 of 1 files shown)


mutatrum (approved):
🚀

Travetown:
A nice daily reminder of how bad the odds are. It reminds me of P-Pool, where a large part of the chart is dedicated to displaying the 13,000 years it takes to mine a block. 
These figures are beyond normal imagination and can hardly be properly ranked. 

What good does it do me to be constantly reminded that I might as well quit because it's just a waste of energy? Looking at a number that doesn't change for years can get really boring. 

PR 1368 is actually sufficient for classification. 

Things like the scoreboard (PR 1236) are much more interesting, because over time you get a real idea of how rare “good” shares are, and we are then moving in the areas where a small miner actually operates. 
Sorry if that sounds a bit harsh.

main/http_server/axe-os/src/app/components/home/home.component.html
```diff
@@ -155,6 +155,13 @@
                                 {{info.bestDiff | diffSuffix}}
                             </span>
                             <span class="text-500 text-lg">all-time best</span>
+                            <span class="text-lg font-medium mb-3" *ngIf="info.networkDifficulty">
+                                <tooltip-text-icon
+                                    [text]="'(' + getNetworkDifficultyPercentage(info) + '%)'"
```
duckaxe, line 160: Regarding https://github.com/bitaxeorg/ESP-Miner/discussions/1085 we need a space before `%`

PR #1408, opened 2025-11-30
Option for display-wake & flashing for alerts

Distortions81:
I like the display sleep/off function, but want to know if there is an important alert.
This adds a [default-off] option in the display settings to wake the display and toggle-invert flash to add a visual alert.
The alert is canceled when the user presses a button.

The important alerts for this are currently: SCR_OVERHEAT, SCR_ASIC_STATUS and block_found.
The display will also wake for IDENTIFY.

Feedback welcome!


mutatrum:
Thank you for the PR. It's a good idea, but maybe overly complicated. It has quite a large overlap with #1366. These alerts shouldn't be cancellable, except for IDENTIFY, which already is. The idea there is that any non-carousel screen will show, even if the display is sleeping or on always off. So also when the Wi-Fi status changes, and anything else that might pop up that's not a carousel screen.

Maybe you can test that PR and see if that also suits the need. It uses a fairly short timeout (5 seconds), but as the user is already in front of the device, that should be sufficient to f.e. check the IP address or hashrate.

PR #1422, opened 2025-12-03
Misc. set of improvements and error checking

### LINKED ISSUE #1053: "Flatline of Death" - hashrate stops changing, no new shares are found. Pool is still alive and sending work
Sometimes Bitaxes will crash and not show any initials signs of it. The hashrate still shows it's hashing, but does not change and no new shares are found. On the AxeOS graph, this shows as a flatline (hence people calling it the Flatline of Death).  

Sometimes it can crash in minutes, sometimes hours and occasionally days for this to occur. A restart will typically fix it.

It would be good if the firmware could detect this crash and recover from it automatically.

Thank you again amazing team!
Steve :-)

mutatrum:
Hopefully fixes #1053:
 * Improve `STRATUM_V1_receive_jsonrpc_line`:
   * Improve `json_rpc_buffer` handling;
 * Add more error handling in `mining.notify` parsing;
 * No-copy logging in `debug_stratum_tx`;
 * Stratum latency measurement improvements:
   * Measure directly after receive, log at the end of processing;
   * Use `send_uid` directly;
 * Use `snprintf` instead of `sprintf`;

Thanks to @Mecanix-live for pointing out several of these issues.

### DIFF
--- a/main/tasks/stratum_task.c
+++ b/main/tasks/stratum_task.c
@@ -7,6 +7,7 @@
 #include "work_queue.h"
 #include "esp_wifi.h"
 #include <esp_sntp.h>
+#include "esp_timer.h"
 #include <sys/time.h>
 #include <stdbool.h>
 #include <string.h>


main/tasks/create_jobs_task.c
```diff
@@ -26,7 +26,7 @@ void create_jobs_task(void *pvParameters)
     {
         mining_notify *mining_notification = (mining_notify *)queue_dequeue(&GLOBAL_STATE->stratum_queue);
         if (mining_notification == NULL) {
-            ESP_LOGE(TAG, "Failed to dequeue mining notification");
+            ESP_LOGW(TAG, "stratum_queue is empty");
```
adammwest: pretty sure the log doesnt make sense here
```c 
ESP_LOGW(TAG, "stratum_queue is empty");
```


because 
```c
(mining_notify *)queue_dequeue(&GLOBAL_STATE->stratum_queue);
```

is blocking when the queue is 0

due to https://github.com/bitaxeorg/ESP-Miner/blob/master/main/work_queue.c#L37
```c
void *queue_dequeue(work_queue *queue)
{
    pthread_mutex_lock(&queue->lock);

    while (queue->count == 0)
    {
        pthread_cond_wait(&queue->not_empty, &queue->lock);
    }
```
if it returns there the queue can only be count > 0, the message itself could be null, but the queue wasnt empty

same for asic_job_queue both use the same mechanism

mutatrum: Good point. Not sure how to deal with a potentially empty queue then, we should never get there, but seeing the flatline of death, it might look like it.

A few options:
1. The use of both queues is different, `stratum_queue` is mostly empty, while `ASIC_job_queue` should be mostly full. If we make both queues to be mostly empty, maybe the structure could be simplified as well, f.e. we don't need to do a queue empty on a mining.notify with the Clean Jobs flag set. The timing of sending new jobs shifts to the create_jobs_task. Not sure if this will work out. 
2. Go queue-less, at least for the ASIC_job_queue. Have one 'current' job, and cycle extranonce_2 on demand, sending it directly to the asic. If a new mining.notify comes in, also send directly to the asic. It's an ever further simplified version of 1.
3. Make a non-waiting version of dequeue, for ASIC_job_queue. Not sure if that makes sense, it adds complexity instead of reducing it.

components/stratum/stratum_api.c
```diff
@@ -282,16 +265,47 @@ void STRATUM_V1_parse(StratumApiV1Message * message, const char * stratum_json)
...
         // new_work->difficulty = difficulty;
         cJSON * params = cJSON_GetObjectItem(json, "params");
-        new_work->job_id = strdup(cJSON_GetArrayItem(params, 0)->valuestring);
+        if (!params || !cJSON_IsArray(params)) {
+            ESP_LOGE(TAG, "Invalid params in mining.notify");
+            free(new_work);
+            goto done;
+        }
+        int params_count = cJSON_GetArraySize(params);
+        if (params_count < 8) {
```
terratec, line 334: clean jobs item is missing or is it optional?

components/stratum/stratum_api.c
terratec, line 380: use params_count

terratec, line 381: check >8

components/stratum/stratum_api.c
mutatrum: It's not specified as optional on [Stratum mining protocol](https://en.bitcoin.it/wiki/Stratum_mining_protocol#mining.notify), but better be safe than sorry.

Furthermore, the old code assumed it could be on position further down, and there's a test guarding that. Going back in history I didn't find any rationale for that, but for this also: better safe than sorry.

main/tasks/create_jobs_task.c
mutatrum: Ok, I went with the queue-less approach. It has quite some benefits, code complexity and chances on deadlocks reduce. Less management, etc.

adammwest:
@mutatrum 
I updated https://github.com/adammwest/test-stratum-server
it should cause recv=0, at the second notify message (15s) it closes the connection.
maybe useful as a test.

main/tasks/create_jobs_task.c
adammwest: >Ok, I went with the queue-less approach. It has quite some benefits, code complexity and chances on 
> deadlocks reduce. Less management, etc.

I like this personally, however I think a no queue approach requires a New clean  PR

mutatrum:
> @mutatrum I updated https://github.com/adammwest/test-stratum-server it should cause recv=0, at the second notify message (15s) it closes the connection. maybe useful as a test.

Tested, I made the test-server not exit, so it was available immediately:
```
I (40185) stratum_api: Connection closed by peer
E (40190) stratum_task: Failed to receive JSON-RPC line, reconnecting...
E (40198) stratum_task: Shutting down socket and restarting...
I (40206) stratum_task: Clean Jobs: clearing queue
I (40981) bm1370: Job ID: 48, Asic nr: 0, Core: 33/6, Ver: 0556C000
I (40982) asic_result: ID: job123, ASIC nr: 0, ver: 2556C000 Nonce A2120542 diff 630.4 of 1024.
I (41210) stratum_task: Connecting to: stratum+tcp://192.168.1.229:4446 (192.168.1.229)
I (41511) stratum_task: Socket created, connecting to 192.168.1.229:4446
I (41515) stratum_task: Resetting stratum uid
I (41516) stratum_task: Clean Jobs: clearing queue
I (41518) stratum_api: tx: {"id": 1, "method": "mining.configure", "params": [["version-rolling"], {"version-rolling.mask": "ffffffff"}]}
```
Good tool to have in the toolbox!

main/tasks/create_jobs_task.c
mutatrum: I can split it out, but as a PR on top of the previous work here. So basically keep this PR up to the point of the log message above, and split the rest of the commits off to a new branch. Makes sense.

mutatrum:
Split out #1424

WantClue:
Please resolve conflicts

mutatrum:
> Please resolve conflicts

Done

Issue #310, opened 2024-08-28
Phase out support for bitaxeMax in future firmware updates


skot:
we can do this with a json file in the releases that specifies the minimum HW revision required to run the update. Then AxeOS needs a change to implement this.

This is paired with the esp-miner firmware migration the bitaxeorg GitHub org

jddebug:
Why phase out support? I have a lot of these. They have only recently become stable.

benjamin-wilson:
> Why phase out support? I have a lot of these. They have only recently become stable.

The Max was one of the earliest working bitaxes and did not have a large number of sales when compared to the Ultra and Supra. We believe it has reached a state of stability and now is a good time to sunset any more development. Support for older ASICs adds complexity, bloat and maintenance costs that burdens development of new Bitaxe versions.

skot:
We may want to phase out new update support for some of the earlier Ultras too.. So this JSON should allow for hardware revision number based filtering

Kubaaaaaaaaaa:
> > Why phase out support? I have a lot of these. They have only recently become stable.
> 
> 
> 
> The Max was one of the earliest working bitaxes and did not have a large number of sales when compared to the Ultra and Supra. We believe it has reached a state of stability and now is a good time to sunset any more development. Support for older ASICs adds complexity, bloat and maintenance costs that burdens development of new Bitaxe versions. 

But there could be some UI enhancements in the future that even owners of old MAXs could benefits from.

PR #1424, opened 2025-12-04
Go Queueless

mutatrum:
Eliminate asic_task and ASIC_jobs_queue.

Now there is only one queue, the stratum_queue. When a new stratum message is put in the stratum_queue, it's taken out and in case of a mining.notify with clean_jobs, sent directly to the ASIC. If there's no new message, the dequeue times out based on asic_job_frequency, extranonce_2 is incremented and the jobs is sent to the ASIC.

This eliminated rejected shares completely in testing, the only rejections left are stale jobs.


0xf0xx0:
testing, runnin for 9 hours without any failed submissions

0xf0xx0 (approved):

mutatrum:
Tested it on the Max, 10 hours in, I got 2 duplicate shares:

image

[duplicate.log](https://github.com/user-attachments/files/24326117/duplicate.log)

My suspicion is that the submit is taking too long, and starves the ASIC from new work which wraps around as the job time is only 15ms. I'll test it more with a lower diff pool, to see if it happens more often then.

WantClue:
might be a good time to retire the max finally. #310 
We can remove with that a good chunk of other code finally

eandersson:
Been running this on a Ultra and Supra since it was uploaded and getting the same amount (~0.03%) as on my other devices.

mutatrum:
Fix for duplicate shares on Max is in. After 15 hours no more duplicates:

image

main/main.c
```diff
@@ -97,7 +97,7 @@ void app_main(void)
     if (xTaskCreate(stratum_task, "stratum admin", 8192, (void *) &GLOBAL_STATE, 5, NULL) != pdPASS) {
         ESP_LOGE(TAG, "Error creating stratum admin task");
     }
-    if (xTaskCreate(create_jobs_task, "stratum miner", 8192, (void *) &GLOBAL_STATE, 10, NULL) != pdPASS) {
+    if (xTaskCreate(create_jobs_task, "stratum miner", 8192, (void *) &GLOBAL_STATE, 20, NULL) != pdPASS) {
```
terratec, line 100: Can this change increase the probability of orphan blocks for all bitaxe models when the result task has a lower priority?

mutatrum: I see the same percentage of Stale shares on different devices. I would say the Max is most sensitive to this, because of the short asic job frequency, and it's not significantly higher than the other devices I've tested on.

Issue #1355, opened 2025-11-16
Uptime back on the dashboard


EricDimitri:
Just that.

Having the UpTime visible, helps Understand and know for how long a specific Bitaxe was runnning.

THANK YOU VERY MUCH!

Travetown:
You can still see the exact time. 

<img width="407" height="221" alt="Image" src="https://github.com/user-attachments/assets/c199c361-83d9-47ed-b6ba-ecc9dadee528" />

EricDimitri:
> You can still see the exact time.
> 
> <img alt="Image" width="407" height="221" src="https://private-user-images.githubusercontent.com/66588710/514890544-c199c361-83d9-47ed-b6ba-ecc9dadee528.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjMzMTEwODcsIm5iZiI6MTc2MzMxMDc4NywicGF0aCI6Ii82NjU4ODcxMC81MTQ4OTA1NDQtYzE5OWMzNjEtODNkOS00N2VkLWI2YmEtZWNjOWRhZGVlNTI4LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTExMTYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUxMTE2VDE2MzMwN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTY5NjQyZWRjOGZlYjU3ODgxNTU2Y2FhNmJjMzgwZjI3MjZiM2Y2YzZhMDE5ODc2YzQ1YzA0NmZjZWIxZDNlYTYmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.9uIldJN9mDugN_skEGjC_GmM6PWzhrlJ8q560FkJOOs">

Yes, but not in the dashboard...

WantClue:
[...]

Issue #1402, opened 2025-11-27
Make AxeOS adjustable


WantClue:
We need to make axeos a bit more adjustable and have a minimilized presentation of information.
In order to stay in line with "grannyproof" we need to reduce the information we display.

The current design is phenomenal but lacks in grasp on first sight. Users will be overloaded with information they either don't need or don't understand.
With tools like a draggable and resizeable dashboard we can make AxeOS less busy for the user.

Working on a Draft RN

PR #1428, opened 2025-12-05
Bring back uptime

mutatrum:
Due to popular request, uptime is back on the dashboard. Found a little nook to put it that made.

image

Also slightly increased the font size of the tiny statistics:
image

#1427.
#1356.
#1355.

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -173,7 +173,7 @@
                         {{info.bestSessionDiff | diffSuffix}}
                     </span>
                     <span class="text-500 text-lg">session best</span>
-                    <span class="text-sm font-medium mb-3"> ({{ info.uptimeSeconds | dateAgo: { intervals: 2, strict:true } }})</span>
+                    <span class="text-sm font-medium mb-3"> ({{ info.uptimeSeconds | dateAgo: { intervals: 2, strict:true, short:true } }})</span>
                 </div>
             </div>
         </div>
--- a/main/http_server/axe-os/src/app/pipes/date-ago.pipe.ts
+++ b/main/http_server/axe-os/src/app/pipes/date-ago.pipe.ts
@@ -32,15 +32,9 @@ export class DateAgoPipe implements PipeTransform {
         if (args?.intervals && shownIntervals >= args.intervals) break;
         const counter = Math.floor(seconds / intervals[i]);
         if (counter > 0) {
-          if (counter === 1) {
-            if (result) result += ', '
-            result += counter + ' ' + i + ''; // singular (1 day ago)
-            seconds -= intervals[i]
-          } else {
-            if (result) result += ', '
-            result += counter + ' ' + i + 's'; // plural (2 days ago)
-            seconds -= intervals[i] * counter
-          }
+          if (result) result += args?.short ? ' ' : ', ';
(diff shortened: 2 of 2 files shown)


duckaxe:
On smaller screens, the uptime is broken down into several lines. Perhaps `intervals: 1` is the better choice.

mutatrum:
1200px wide is the worst-case. Maybe it should switch to 2x2 cards earlier:

image


1920 is ok:

image

WantClue:
I think a nicer addition would be to put it into app.topbar.component. That way we have a clean overview of the uptime every time and now only on the dashboard @mutatrum

EricDimitri:
@mutatrum Amazing!! Really useful. Thanks a lot!! IS it in any beta version?

mutatrum:
Will move to top bar, good suggestion.

EricDimitri:
> Will move to top bar, good suggestion.

Excelent!! You also have more space there, for some other critical info, or that you want to show no matter the MEnu you are in.

Nice!!

duckaxe:
I'm not sure it's a good idea to include uptime in the top bar. Uptime isn't important enough to warrant such a prominent place. Besides, we don't have room for it on mobile. I would reserve the top bar exclusively for important buttons.

WantClue:
> I'm not sure it's a good idea to include uptime in the top bar. Uptime isn't important enough to warrant such a prominent place. Besides, we don't have room for it on mobile. I would reserve the top bar exclusively for important buttons.

I do agree, but I would also argue it's not a good place in the containers. A potential option would be to show it only on desktop mode.

duckaxe:
> I do agree, but I would also argue it's not a good place in the containers. A potential option would be to show it only on desktop mode.

On mobile there is no space for a string. 
As long as we can use `intervals: 1`, we can place the uptime in the sidebar below the menu. It will also work on mobile devices. 

Screenshot

duckaxe:
Personally, I would put the uptime information back on the card and allow users to select it https://github.com/bitaxeorg/ESP-Miner/issues/1402 if they find it important. In my opinion, we are heading in the wrong direction if we try to fill the dashboard with unimportant information. This would not be good for readability and usability.

EricDimitri:
> Personally, I would put the uptime information back on the card and allow users to select it #1402 if they find it important. In my opinion, we are heading in the wrong direction if we try to fill the dashboard with unimportant information. This would not be good for readability and usability.

I understand this might not be a priority for the dev team, but labeling it as 'unimportant information' seems dismissive, especially since several users have stated it is extremely useful to them. Utility is subjective; just because it isn't critical for one developer doesn't mean it lacks value. The dashboard already includes various metrics that many users might not use, but are retained for the specific subset that needs them, and pronbably because some devs find those "useful"

Cheers 8-)

mutatrum:
Alternatively, it could be an abbreviated uptime: `1h 5m`.

EricDimitri:
> Alternatively, it could be an abbreviated uptime: `1h 5m`.

OF course!!

WantClue (approved):
LGTM

EricDimitri:
@mutatrum, when will this be relased? Thanks!!

mutatrum:
> @mutatrum, when will this be relased? Thanks!!

Somewhere in Januari most likely.

EricDimitri:
@mutatrum 

Even Beta?

EricDimitri:
THANK YOU VERY MUCH!!!!! @mutatrum It looks awesome, and pretty useful!!!

PR #1432, opened 2025-12-05
show units in self test

### LINKED ISSUE #1390: Add units to self-test logs
Currently in self-test values are logged without units, which can be confusing.

WantClue:
fixes #1390


ghost:
Factory firmware was flashed to the Gamma 601 using the Bitaxetool and the .cvs file for the 601, capture was collected via the Web Flasher after the flash was completed.

Adding the logs here for reference for this PR.

[bitaxe-logs-PR1432-self-test-2025-12-06T15-57-45-845Z.txt](https://github.com/user-attachments/files/23997536/bitaxe-logs-PR1432-self-test-2025-12-06T15-57-45-845Z.txt)

PR #1433, opened 2025-12-05
Add missing keys to openapi.yaml, split responses into schemas

0xf0xx0:
we definitely need something to ensure the openapi.yaml is in sync with the backend x3

i think i got all the keys, but more than likely missed one or two


duckaxe (approved):
Wonderful.

PR #1434, opened 2025-12-07
enable vid pid and serial logging to tinyusb

WantClue:
With this we can write the https://pid.codes/org/Bitaxe/ to the esp. 
This will have the result that new flashing via usb connection needs to be done via boot mode. 

Users now need to press the BOOT button before starting the device in order to switch into BOOT mode. 
This will need to be communicated.


ghost:
Tested this PR.

I got no issues with this where the "BOOT" button needs to be held down and plugging in the power cable so that the device can be flashed with the Web Flasher or the Bitaxetool. 

If this PR gets approved, the "readme.md" file will need updating to go with it.

PR #1443, opened 2025-12-11
Turn on the screen for identify mode

terratec:
There is no separate screen for the identify image to trigger the display activity.
In timeout mode the display remains on as long as the identify image is shown.


mutatrum:
I think this is also covered in #1366

terratec:
> I think this is also covered in #1366

No, because the identify mode is just an overlay?

mutatrum:
> > I think this is also covered in #1366
> 
> No, because the identify mode is just an overlay?

You are correct.

PR #1449, opened 2025-12-12
feat: warning if default address is been used

### LINKED ISSUE #1411: Show banner if default payout address is used
Show a warning banner when the default stratum user is being used for hashing. Could be a mix between a thank you and a warning, as it's possible people do this on purpose, but we don't want people to continue hashing to an address that's not theirs.

WantClue:
fixes #1411

### DIFF
--- a/main/http_server/axe-os/src/app/components/pool/pool.component.ts
+++ b/main/http_server/axe-os/src/app/components/pool/pool.component.ts
@@ -21,6 +21,8 @@ export class PoolComponent implements OnInit {
   public form!: FormGroup;
   public savedChanges: boolean = false;
 
+  public readonly DEFAULT_BITCOIN_ADDRESS = 'bc1qnp980s5fpp8l94p5cvttmtdqy8rvrq74qly2yrfmzkdsntqzlc5qkc4rkq';
+
   public pools: PoolType[] = ['stratum', 'fallbackStratum'];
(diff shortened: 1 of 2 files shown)


duckaxe:
Does it make sense to display the notification on the dashboard, because we show all notifications there?

main/http_server/axe-os/src/app/components/pool/pool.component.ts
```diff
@@ -231,4 +233,13 @@ export class PoolComponent implements OnInit {
   trackByFn(index: number, option: ITlsOption): number {
     return option.value;
   }
+
+  isUsingDefaultAddress(pool: PoolType): boolean {
```
duckaxe, line 237: Better `isPoolUsingDefaultAddress`?

WantClue:
> Does it make sense to display the notification on the dashboard, because we show all notifications there?

I don't think so. Might be an initial catch but do we want to load everything into the dashboard ? I don't think so

main/http_server/axe-os/src/app/components/pool/pool.component.ts
WantClue: same same

duckaxe:
I see. But imagine someone gets their Bitaxe and never visits the pools page. They will never receive the notification.

PR #1460, opened 2025-12-16
Fix possible out-of-bounds write

terratec:
The util function bin2hex writes a null-terminated string and should check the length of the destination buffer including the null-termination.


WantClue:
Please add a description <3

mutatrum:
Good find, lgtm

PR #1468, opened 2025-12-23
Add name resolve for public IPv6 addresses

mutatrum:
This should fix resolving of IPv6 addresses


main/tasks/stratum_task.c
```diff
@@ -68,95 +69,124 @@ typedef struct {
...
-                    if (esp_netif) {
-                        int netif_index = esp_netif_get_netif_impl_index(esp_netif);
-                        if (netif_index >= 0) {
-                            addr6->sin6_scope_id = (u32_t)netif_index;
-                            ESP_LOGI(TAG, "Set scope_id to interface index: %lu", (unsigned long)addr6->sin6_scope_id);
-                        }
-                    }
-                }
+    // Preferred order: IPv4 first, then IPv6
+    const int preferred_families[] = { AF_INET, AF_INET6 };
```
0xf0xx0, line 105: this just puts us in the same place x3 ipv6 should be furst, but only if the network supports ipv6

mutatrum: If both are supported IPv4 has precedence. If it's an IPv6 only domain name, it should now resolve. The latter was broken.

0xf0xx0: > If both are supported IPv4 has precedence. If it's an IPv6 only domain name, it should now resolve. The latter was broken.

and resolution of a domain thats both v4 and v6 x3 if the pool has separate v4 and v6 domains like ckpool then this might work (i tested with atlas), but ipv6 devices should prefer v6 unless theres no AAAA record.

mutatrum: We dont know if IPv6 routing works only based on name resolution, to make that properly work it needs more work That's outside the scope of this PR. This is only for restoring the resolve of IPv6 hostnames.

0xf0xx0: fair, ill test on ck in a bit.

0xf0xx0 (approved):
resolves solo6.ckpool.org correctly

main/tasks/stratum_task.c
ffrediani: Understand this is to fix an existing issue, but 0xf0xx0 is correct that if there is IPv4 and IPv6 should always take precedence. That's how any modern operating system works now a days and always prefers IPv6 over IPv4 which is legacy internet protocol.
Not to be forgotten and revised at a new issue at some point.

PR #1478, opened 2026-01-01
801 Strapless working

benjamin-wilson:
These changes allow proper configuration for the strapless two phase TPS546 found in the Bitaxe GT 801


benjamin-wilson:
Self test passing on fresh 801 GT

PR #1480, opened 2026-01-01
Self test was prevously not rolling ntime and the search space was rolling over. Improve logging, misc

benjamin-wilson:
Self test was prevously not rolling ntime and the search space was rolling over. Improve logging, misc

### DIFF
--- a/main/main.c
+++ b/main/main.c
@@ -57,6 +57,12 @@ void app_main(void)
         return;
     }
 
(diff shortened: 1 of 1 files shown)


benjamin-wilson:
In EMC2103 the offset is only applied in get_external_temp
Hm. Lets set the offset to 0 for now and we can figure out how to handle this later. I don't think the offset should be applied to the 2101 internal temp?

PR #1483, opened 2026-01-02
improve empty string on null init + formatting

WantClue:
formatting of main.c

the strdup can produce a NULL if there is no safety check properly. 
with this it's a proper check


ghost:
Factory firmware from the repo pulled down and flashed to the Gamma 601 using the Bitaxetool to test `SELF-TEST`.


https://github.com/user-attachments/assets/4984b3bd-afe9-4649-b192-459579cf73e0

PR #1490, opened 2026-01-05
Tps phase registers

WantClue:
Phase register 0xFF is not correct for a single phase. This new device config writes the proper phase to the TPS if it is in single or multi mode.

0x00 should be usually used when targeting single phase device configurations.
https://www.ti.com/lit/ds/symlink/tps546d24a.pdf?ts=1767579305045

### DIFF
--- a/main/power/vcore.c
+++ b/main/power/vcore.c
@@ -16,76 +16,75 @@ static const char *TAG = "vcore";
 
 static TPS546_CONFIG get_tps546_config(const FamilyConfig * family)
 {
+    TPS546_CONFIG config = {0};
 
     // Set family-specific parameters
     switch (family->id) {
     case GAMMA_TURBO:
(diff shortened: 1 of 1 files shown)


main/power/vcore.c
```diff
@@ -1,81 +1,101 @@
...
+/**
+ * @brief Generate TPS546 configuration based on family and phase flag
+ * @param family Pointer to family configuration
+ * @return Complete TPS546_CONFIG structure
+ */
+static TPS546_CONFIG get_tps546_config(const FamilyConfig * family)
+{
+    TPS546_CONFIG config = {
+        // Always set phase based on device config flag
+        .TPS546_INIT_PHASE = family->tps546_multi_phase ? TPS546_INIT_PHASE_MULTI : TPS546_INIT_PHASE_SINGLE,
```
mutatrum: This can be moved down into the 3 big case statements. No need to add it to `device_config.h` as it is derived from family.

PR #1495, opened 2026-01-09
docs: stratumExtranonceSubscribe is returned as an int

adamdecaf:
The value is being written as a number, so either the docs are wrong or the code should be write a boolean.

https://github.com/bitaxeorg/ESP-Miner/blob/v2.12.2/main/http_server/http_server.c#L877


0xf0xx0:
all the bools are ints from axeos iirc, one of the goals of the api refactor im pushing for is to make those truly bools. come comment on #945 :3

PR #1496, opened 2026-01-09
Some small bugs and cleanups

mutatrum:
Found some bugs and did some cleanup while working on something else:

 * Moved init of `version` and `axeOSVersion` to `system.c`;
 * `Power_get_current` and `VCORE_get_voltage_mv` are now only called from PowerManagementTask;
 * Fix a bug in the formatting of share numbers in BAP;

### DIFF
--- a/main/http_server/http_server.c
+++ b/main/http_server/http_server.c
@@ -36,7 +36,6 @@
 #include "global_state.h"
 #include "nvs_config.h"
 #include "vcore.h"
-#include "power.h"
 #include "connect.h"
 #include "asic.h"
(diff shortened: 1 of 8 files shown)


main/http_server/http_server.c
```diff
@@ -36,7 +36,6 @@
 #include "global_state.h"
 #include "nvs_config.h"
 #include "vcore.h"
```
mutatrum, line 38: `vcore.h` should also not be here, but that's for another day.

WantClue (approved):
code lgtm

PR #1500, opened 2026-01-11
Customize screens

### LINKED ISSUE #711: Dynamic Screen Information Settings, including adding Voltage Regulator Temperatures
Currently the Bitaxe switches between 2 screens with particular information on them. I find some of the stats helpful, and some of them no longer useful once it is configured. It would be nice to have a way to select the options that would be shown on the screen in a check list.

Due to space considerations the user could pick so many items per screen, specify which screen they would be included on, and how many different screens would be shown. 

For example if a user wanted a lot of information they could specify 3 screens of data, including more information with the new suggestions below, or others if more are considered.

Some other options that would be nice to offer on the screens that are not currently available:
1) The wallet address for the mining reward
2) The Voltage Regulator temperature (higher priority)
3) The long term average hashrate over a period of time

For my personal usage I would like to be able to setup a single screen showing Hashrate, energy usage per TH, Asic temp and Voltage Regulator Temperature, and skip showing the current IP information and SSID info.

In the case a user has issues connecting and cannot remember the SSID or the IP of the unit, we could have a way to show these by pressing one of the two buttons quickly (or holding) on the front to show this information, but for the most part the IP and SSID are not needed once configured, and scrolling to these screens is not helpful.

### LINKED ISSUE #1465: Be able to dismiss certain notifications
### Discussed in https://github.com/bitaxeorg/ESP-Miner/discussions/1462

<div type='discussions-op-text'>

<sup>Originally posted by **mediageni** December 17, 2025</sup>
Would be nice if there is a close button to remove this message from the dashboard and the miner display on the device
Today is your lucky day! Your device has found a block 🎉 </div>

mutatrum:
Replace the fixed screens with a way for the user to customize them. This enables users to add or remove information they deem useful, have more or less screens. It also opens the door for leveraging the the biggest screens, as the templates can be made to size for 8 or even 16 lines.

image

An example what you can do:

image

It's not completely finished, there are a few open points:
 * The 'restore defaults' functionality is a bit iffy and probably not completely working yet. I'm not sure how to migrate into this functionality.
 * I want to add a 12px and 16px high font, so people can add big text to the screen. Didn't get round to it yet. The idea is to use a simple markdown-like syntax, such as adding `#` and `##` in front of a line to set the font. Ideas welcome.
 * `display_config.c` is a bit unwieldy with all the handle functions. Not sure if it gets better with a different structure though. I can make it smaller but that would also make it more complex.


ghost:
Tested this on an Ultra 207 only.

Failed to boot up properly, boot log enclosed.

[1500-207-boot-log.txt](https://github.com/user-attachments/files/29448467/1500-207-boot-log.txt)

mutatrum:
Sorry, still working on the merge. Should've mentioned that.

mutatrum:
image

PR #1501, opened 2026-01-11
Single options request handler for API

mutatrum:
No need to define separate options handlers for each API path.


WantClue:
awesome shit

PR #1442, opened 2025-12-11
Generate frontend api services from openapi spec

0xf0xx0:
revival of #832  

just needs the system and theme services replaced, im not sure how to go around that xmx could someone experienced in the frontend take a look?

closes: #832

PR #1503, opened 2026-01-12
Frontend openapi

WantClue:
this fixes the necessary frontend changes required so that #1442 does work

### DIFF
--- a/main/http_server/axe-os/src/app/services/system.service.ts
+++ b/main/http_server/axe-os/src/app/services/system.service.ts
@@ -8,7 +8,9 @@ import {
   SystemInfo as ISystemInfo,
   SystemStatistics as ISystemStatistics,
   SystemASIC as ISystemASIC,
-  SystemASICASICModelEnum
+  SystemASICASICModelEnum,
+  SystemService as GeneratedSystemService,
+  Settings
 } from 'src/app/generated';
 
 import { environment } from '../../environments/environment';
--- a/main/http_server/axe-os/src/app/components/system/system.component.ts
+++ b/main/http_server/axe-os/src/app/components/system/system.component.ts
@@ -6,7 +6,7 @@ import { SystemApiService } from 'src/app/services/system.service';
 import { LoadingService } from 'src/app/services/loading.service';
 import { DateAgoPipe } from 'src/app/pipes/date-ago.pipe';
 import { ByteSuffixPipe } from 'src/app/pipes/byte-suffix.pipe';
-import { SystemInfo as ISystemInfo, SystemASIC as ISystemASIC, } from 'src/app/generated';
+import { SystemInfo as ISystemInfo, SystemASIC as ISystemASIC, GenericResponse, } from 'src/app/generated';
 
 type TableRow = {
   label: string;


main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -461,6 +462,8 @@ export class SwarmComponent implements OnInit, OnDestroy {
         return { color: 'red', msg: 'Overheated' };
       case !!axe.power_fault:
         return { color: 'red', msg: 'Power Fault' };
+      case axe.isIdentifying === 1:
```
mutatrum: should `isIdentifying` be a bool?

main/http_server/axe-os/src/app/services/system.service.spec.ts
```diff
@@ -1,16 +1,16 @@
 import { TestBed } from '@angular/core/testing';
 
-import { SystemService } from './system.service';
+import { SystemApiService } from './system.service';
 import { provideHttpClient } from '@angular/common/http';
 
-describe('SystemService', () => {
-  let service: SystemService;
+describe('SystemApiService', () => {
```
mutatrum, line 6: Is this rename necessary? If we keep it SystemService the PR would be quite a bit smaller.

main/http_server/axe-os/src/app/services/system.service.ts
```diff
@@ -196,34 +207,54 @@ export class SystemService {
   }
 
   public restart(uri: string = '') {
-    return this.httpClient.post(`${uri}/api/system/restart`, {}, {responseType: 'text'});
+    if (environment.production && this.generatedSystemService && !uri) {
```
mutatrum, line 211: What's the function of the two if's here? Is this one for this device, and the second one (with `uri`) for swarm?

main/http_server/axe-os/src/app/services/system.service.ts
```diff
@@ -267,10 +308,18 @@ export class SystemService {
   }
 
   public getSwarmInfo(uri: string = ''): Observable<{ ip: string }[]> {
-    return this.httpClient.get<{ ip: string }[]>(`${uri}/api/swarm/info`).pipe(timeout(5000));
+    if (!uri) {
+      // Local device swarm info - this might not be implemented in the generated API
```
mutatrum: Why wouldn't it be implemented in the generated API?

main/http_server/axe-os/src/app/services/system.service.spec.ts
WantClue: renamed it to SystemApiService because the OpenAPI generator creates a SystemService class. Our wrapper service adds functionality that isn't in the generated API - specifically swarm support (calling other devices via URI), development mock data, and consistent timeout handling. The generated API only handles calls to the local device, but we need to be able to call any device in the swarm network.

main/http_server/axe-os/src/app/services/system.service.ts
WantClue: check if generatedSystemService exists because it's marked @Optional() in the constructor - this provides graceful degradation. The !uri check is crucial because the generated API only knows how to call the local device, not arbitrary remote URIs in network.

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
WantClue: true it's gone

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -226,9 +226,10 @@ export class SwarmComponent implements OnInit, OnDestroy {
         this.toastr.error(errorMsg, `Device at ${axe.IP}`);
         return of(null);
       })
-    ).subscribe(res => {
+    ).subscribe((res: any) => {
       if (res !== null) {
-        this.toastr.success(res, `Device at ${axe.IP}`);
+        this.toastr.success(res.message, `Device at ${axe.IP}`);
```
mutatrum, line 238: This will fail when used on an older swarm device, as it won't have the `message` field. Unfortunately, it needs a backwards compatibility check.

PR #1504, opened 2026-01-12
Add support for indexed strings in NVS config

mutatrum:
Code extracted from both #1236 and #1500, so those PRs can be smaller.


main/nvs_config.c
```diff
@@ -226,6 +247,17 @@ esp_err_t nvs_config_init(void)
         ESP_LOGW(TAG, "Could not open nvs");
         return err;
     }
+        
+    nvs_stats_t stats;
+    err = nvs_get_stats(NULL, &stats);
+    if (err == ESP_OK) {
+        ESP_LOGI(TAG, "Used entries: %lu", stats.used_entries);
```
WantClue, line 254: might a LOGD be more useful here instead of a LOGI ?

Issue #1476, opened 2025-12-31
ASIC Temp chart misleading on multi chip units


bonifacio123:
On multi-chip units it seems the graph is charting the ASIC Temps for ASIC #1 (800 GT in this example). For this metric I would think the highest temperature value of all the ASIC's should be used. Thank you.

<img width="1452" height="637" alt="Image" src="https://github.com/user-attachments/assets/cd620573-c8ac-4126-86a9-d46a9d1c805e" />

<img width="506" height="649" alt="Image" src="https://github.com/user-attachments/assets/e14405ce-8619-41ac-acd9-702a78cd525a" />

mutatrum:
See #1730

PR #1509, opened 2026-01-15
Graph the hottest ASIC temp for multi chip units

bonifacio123:
This PR addresses issue "ASIC Temp chart misleading on multi chip units #1476". Currently the graph chart ASIC #1 on a multi chip unit. The PR charts the hottest ASIC temp. This change was tested on a 601 and 801

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.ts
+++ b/main/http_server/axe-os/src/app/components/home/home.component.ts
@@ -760,7 +760,13 @@ export class HomeComponent implements OnInit, OnDestroy {
       case eChartLabel.hashrate_10m:       return info.hashRate_10m;
       case eChartLabel.hashrate_1h:        return info.hashRate_1h;
       case eChartLabel.errorPercentage:    return info.errorPercentage;
-      case eChartLabel.asicTemp:           return info.temp;
+      case eChartLabel.asicTemp:
+        switch (true) {                    // we want to show to hottest asic temp on the graph
+          case info.temp2 > info.temp:
+            return info.temp2;
+          default:
+            return info.temp; 
+        }
       case eChartLabel.vrTemp:             return info.vrTemp;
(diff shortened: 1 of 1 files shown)


main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -760,7 +760,13 @@ export class HomeComponent implements OnInit, OnDestroy {
       case eChartLabel.hashrate_10m:       return info.hashRate_10m;
       case eChartLabel.hashrate_1h:        return info.hashRate_1h;
       case eChartLabel.errorPercentage:    return info.errorPercentage;
-      case eChartLabel.asicTemp:           return info.temp;
+      case eChartLabel.asicTemp:
+        switch (true) {                    // we want to show to hottest asic temp on the graph
```
mutatrum, line 764: This is a bit convoluted. Could be much simpler:
```
return info.temp2 > info.temp ? info.temp2 : info.temp;
```

But besides the code, maybe we should add `asic2Temp` directly, similar to `fan2rpm`. Or, the neat solution would be to add both lines if someone selects `asicTemp`, but that's more complicated.

bonifacio123: That was my initial solution but somehow thought Switch would make future support of additional ASICs easier. Wrote this late - looking now you're right. I can change it. For me I want to see the hottest ASIC temp. I'm not versed enough in the code base to add another like to the graph.

bonifacio123: Replaced with PR 1514

mutatrum: Oh you didn't have to reopen a PR, you can just continue committing to the same branch and the PR will update.

PR #1523, opened 2026-01-22
Add support Webhook

quyendang:
Allow periodically sending Bitaxe information to a webhook so that the Bitaxe miner’s status can be monitored remotely; this feature is intended for developers.
###  Change setting via Web UI
Screenshot 2026-01-22 at 14 43 05

### Enable Webhook via API

Use the following API request to enable the webhook feature and configure webhook settings on Bitaxe.

#### Endpoint
PATCH `/api/system`

#### Headers
Content-Type: application/json

#### Example Request
```bash
curl -X PATCH http://bitaxeip/api/system \
  -H "Content-Type: application/json" \
  -d '{
    "webhookEnabled": true,
    "webhookUrl": "http://webhook.com/bitaxe",
    "webhookInterval": 60
  }'

### CODE REVIEW

File: main/tasks/webhook_task.c [hunk 1]
```diff
@@ -0,0 +1,341 @@
...
+    double d_value = round((double)number * FACTOR) / FACTOR;
+    return cJSON_AddNumberToObject(object, name, d_value);
+}
+
+static cJSON* cJSON_CreateFloat(float number) {
+    double d_value = round((double)number * FACTOR) / FACTOR;
+    return cJSON_CreateNumber(d_value);
+}
+
+// Helper function to create system info JSON (similar to GET_system_info but without httpd_req_t)
```
mutatrum: This should be possible without the code duplication. Maybe extract this code into a function that just creates the response json that's used both here as well in http_server?

mutatrum (changes requested):
Need code deduplication

mutatrum:
I think this is superceded by #1623

PR #1524, opened 2026-01-22
Clear power fault on swarm

### LINKED ISSUE #1519: power fault doesn't get's cleared
i think the power fault doesn't get's cleared on the swarm page

mutatrum:
Fixes #1519


WantClue:
thank you

PR #1537, opened 2026-02-04
Add support for 650

benjamin-wilson:
Confirmed working

https://github.com/bitaxeorg/bitaxeGamma/tree/650-duo


mutatrum:
Am I understanding this correctly that the 650 is running with the same TPS546 config as the Gamma, within the same power limits?

benjamin-wilson:
> Am I understanding this correctly that the 650 is running with the same TPS546 config as the Gamma, within the same power limits?

Correct

benjamin-wilson:
Released 650
https://github.com/bitaxeorg/bitaxeGamma/releases/tag/gamma-duo-650

PR #1544, opened 2026-02-07
Add option to disable coinbase tx decode

### LINKED ISSUE #1540: Handle coinbase tx parser for other chains
#1391 raised some concerns of confusion when people are mining other pools. I wrote this on Discord, but for continuity, I copied it here:

This (1391) was added so people can verify what the payout address is. There have been a [few instances of scam pools ](https://github.com/mweinberg/stratum-speed-test/tree/main/findings) that say they mine bitcoin, but replace the work and payout for something completely different. I'm not talking about other chain pools, or shared pools, these were advertised as bitcoin pools with the intention of stealing the miners work. This gives the user details on what the actual work for the pool is, so they verify this work. It automates what other people have created to verify the ⁨`mining.notify⁩` stratum messages. IMO this is a valuable addition.

Having said that, this is a bitcoin only solution. Part of this is indeed philosophical: OSMU is focused on bitcoin mining, and while these miners work on other chains, there is no specific support for them in the firmware. However, besides this philosophical point is also a technical point: there is no information in the stratum protocol on what the job actually is. The addresses in the coinbase transaction are stored as hashes, not in base58 or bech32. The prefixes of the address are derived from the output script. These are the same for all forks, but each fork has their own address prefixes. Instead of `⁨1`⁩, `⁨3`⁩, `⁨bc1q`⁩ and `⁨bc1p`⁩ they use different prefixes. We don't know - and will never know - what these are if the user is mining something different.

A small bit of information is the address format of the payout address in the pool user field. However, there are pools that pay out in bitcoin, but mine something else. So assuming the chain is or is not bitcoin cannot be derived solely from the user address. Other detection methods (block height, network diff) are approximate as well, and are suboptimal.

This leaves little options left to do:
1) Do nothing and have confused users;
2) Have a checkbox at the pool configuration to disable this;
3) Be able to disable the warning; (This might be tricky with fallback pools).
4) Anything else?...

mutatrum:
Added option to disable coinbase tx decoding.

Fixes #1540

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -496,7 +496,7 @@
                         Amount
                     </div>
                     <ng-container *ngIf="info.coinbaseValueTotalSatoshis && info.coinbaseValueTotalSatoshis > 0">
-                        {{ info.coinbaseValueTotalSatoshis }}
+                        {{ info.coinbaseValueTotalSatoshis / 100_000_000 }}
                     </ng-container>
                 </div>
             </div>


0xf0xx0:
non-blocking nit: i expected the coinbase amount to still be formatted in btc
image

works for me uwu

mutatrum:
> non-blocking nit: i expected the coinbase amount to still be formatted in btc

Do you know if all forks with the 8 decimal format?

0xf0xx0:
btc forks should keep at least 8, dont see why we should care about em tho x3

mutatrum:
> btc forks should keep at least 8, dont see why we should care about em tho x3

We just want to minimize the number of issues raised and discord questions on incorrect data for non-bitcoin forks.

mutatrum:
> code lgtm, but maybe this should be opt in and not enabled by default ?

Nobody will enable this. I'd rather have people complaining than 95% of the people never seeing this. I added a tooltip, that might nudge people towards looking at the config instead of opening issues.

PR #1545, opened 2026-02-07
statistics_task: use uint64_t timestamps to prevent 32-bit overflow

r3mko:
The currentTime is in **ms** and it overflows after about 24 days, 20 hours, 31 minutes with int32_t.

Screenshot 2026-02-07

Making it an uint64_t should fix that.

### DIFF
--- a/main/tasks/statistics_task.c
+++ b/main/tasks/statistics_task.c
@@ -119,12 +119,12 @@ void statistics_task(void * pvParameters)
     TickType_t taskWakeTime = xTaskGetTickCount();
 
     while (1) {
-        const int32_t currentTime = esp_timer_get_time() / 1000;
+        const uint64_t currentTime = esp_timer_get_time() / 1000;
         const uint16_t configStatsFrequency = nvs_config_get_u16(NVS_CONFIG_STATISTICS_FREQUENCY);
-        const uint32_t statsFrequency = configStatsFrequency * 1000;
+        const uint64_t statsFrequency = (uint64_t)configStatsFrequency * 1000;
 
         if (0 != statsFrequency) {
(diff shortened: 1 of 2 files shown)


mutatrum:
Good find. There are a few other places in the code that use `esp_timer_get_time()` and `uint32_t` that might have the same problem.

r3mko:
**Current 32-bit ms sites:**

hashrate_monitor_task.c (line 176)
`uint32_t time_ms = esp_timer_get_time() / 1000;`

bap_handlers.c (line 40)
`uint32_t current_time = esp_timer_get_time() / 1000;`

bap_subscription.c (line 108)
`uint32_t current_time = esp_timer_get_time() / 1000;`

bap_subscription.c (line 213)
`uint32_t current_time = esp_timer_get_time() / 1000;`

self_test.c (line 454) and self_test.c (line 468)
`uint32_t start_ms / duration_ms based on esp_timer_get_time() / 1000`

PID.c (line 5)
`millis() returns unsigned long from esp_timer_get_time() / 1000 (32-bit on ESP32)`

screen.c (line 650)
stores uptime seconds in uint32_t, that would roll after ~136 years

_**Related storage fields tied to those 32-bit ms values:**_

bap_handlers.c (line 26) (last_message_time)
bap_subscription.h (line 23) (last_response)
bap_subscription.h (line 24) (update_interval_ms)
bap_subscription.h (line 25) (last_subscribe)

I think most are used for timedeltas / elapsed-time checks, so should be "safe".
But I could update those as well if you want.

mutatrum (approved):

mutatrum:
Thanks for the analysis, the rest are fine as they use unsigned wraparound and only work on short timescales.

PR #1546, opened 2026-02-10
Check if the address is in the same /24 network as the device's IP

V45370:
Allow non-standard private networks other than 10.0.0.0 - 10.255.255.255, 172.16.0.0 - 172.31.255.255 and 192.168.0.0 - 192.168.255.255


WantClue:
RFC 1918 defines the standard IP ranges we do not move away from them.

PR #1555, opened 2026-02-17
Add total uptime and hashes

### LINKED ISSUE #1551: Store total uptime and total hashes
Updating a counter every second will wear the NVS out. However, assuming a 24kb NVS partition:

Image

once a minute should be doable actually.

As for the total work, we could use `log2_work`:

> It is the 2-logarithm of the expected number of block header hash attempts were necessary the build the chain up to that point.

mutatrum:
Measures total device uptime, as well as total hashes and calculated log2 work. Values are once an hour persisted to NVS, and stored in `noinit` fields, so they are preserved over soft reboots. Only on a power cycle, some data will be lost.

image

image

Note: pressing the reset button is a power-on boot.


0xf0xx0:
related, mayb we should stick the total work in the misc card from #1687?

mutatrum:
> related, mayb we should stick the total work in the misc card from #1687?

Done.

WantClue (approved):
ACK

PR #1567, opened 2026-02-20
Fix matching in QuicklinkService and add M45Core.com

Distortions81:
Regex so *.m45core.com/user and m45core.com/user function.
Site will handle this correctly next pool reboot.

### DIFF
--- a/components/stratum/include/utils.h
+++ b/components/stratum/include/utils.h
@@ -27,7 +27,7 @@ double networkDifficulty(uint32_t nBits);
 
 void suffixString(uint64_t val, char * buf, size_t bufsiz, int sigdigits);
 
-float hashCounterToGhs(uint32_t duration_ms, uint32_t counter);
+float hashCounterToGhs(uint64_t duration_us, uint32_t counter);
 
 #define STRATUM_DEFAULT_VERSION_MASK 0x1fffe000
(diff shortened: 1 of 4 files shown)


main/http_server/axe-os/src/app/services/quicklink.service.ts
```diff
@@ -33,6 +33,7 @@ export class QuicklinkService {
       { search: 'parasite.wtf', url: `https://parasite.space/user/${user}` },
       { regex: /^(eu|au)?solo[46]?.ckpool\.org/, url: `https://$1solostats.ckpool.org/users/${user}` },
       { search: 'atlaspool.io', url: `https://atlaspool.io/dashboard.html?wallet=${user}` },
+      { regex: /^((?:[a-z0-9-]+\.)*)m45core\.com$/i, url: `https://$1m45core.com/user/${user}` },
```
0xf0xx0: nit: better regex would be `^((eu|tinyminer)\.)?m45core\.com$` if you want to explicitly define each subdomain, or `^(.+?\.)?m45core\.com$` (thisll also match all sub-sub domains)  
no need for `i` flag either

0xf0xx0 (changes requested):

PR #1575, opened 2026-02-23
Add m45core pool quick link

Distortions81:
Updates quicklink matching to be safer (case-insensitive + exact or suffix host matching) and adds m45core.com quicklink, including subdomains and sub-sub domains. I reverted changes and that automatically closed the PR it seems. I hate github.

### DIFF
--- a/main/http_server/axe-os/src/app/services/quicklink.service.ts
+++ b/main/http_server/axe-os/src/app/services/quicklink.service.ts
@@ -34,7 +34,7 @@ export class QuicklinkService {
       { search: 'parasite.wtf', url: `https://parasite.space/user/${user}` },
       { regex: /^(eu|au)?solo[46]?.ckpool\.org/, url: `https://$1solostats.ckpool.org/users/${user}` },
       { search: 'atlaspool.io', url: `https://atlaspool.io/dashboard.html?wallet=${user}` },
-      { regex: /^((?:[a-z0-9-]+\.)*)m45core\.com$/, url: `https://$1m45core.com/user/${user}` },
+      { regex: /^(eu\.|tinyminer\.)?m45core\.com$/, url: `https://$1m45core.com/user/${user}` },
     ];
 
     const normalizedStratumURL = stratumURL.toLowerCase();


main/http_server/axe-os/src/app/services/quicklink.service.ts
```diff
@@ -33,14 +34,18 @@ export class QuicklinkService {
       { search: 'parasite.wtf', url: `https://parasite.space/user/${user}` },
       { regex: /^(eu|au)?solo[46]?.ckpool\.org/, url: `https://$1solostats.ckpool.org/users/${user}` },
       { search: 'atlaspool.io', url: `https://atlaspool.io/dashboard.html?wallet=${user}` },
+      { regex: /^((?:[a-z0-9-]+\.)*)m45core\.com$/, url: `https://$1m45core.com/user/${user}` },
```
mutatrum: https://github.com/bitaxeorg/ESP-Miner/pull/1567#discussion_r2836695449

main/http_server/axe-os/src/app/services/quicklink.service.ts
```diff
@@ -33,14 +34,18 @@ export class QuicklinkService {
...
     ];
 
+    const normalizedStratumURL = stratumURL.toLowerCase();
+
     for (const pool of pools) {
-      if ('search' in pool && stratumURL.includes(pool.search)) {
+      if ('search' in pool
+        && (normalizedStratumURL === pool.search || normalizedStratumURL.endsWith(`.${pool.search}`))) {
```
mutatrum, line 44: Why change this from includes to endsWith? That'll break the search entries.

Distortions81: Example: public-pool.io.atlaspool.io would send user to public-pool.io

Distortions81: Mismatches like parasite.wtf.com would also be possible with `search`

Distortions81: image

The matching still works, unless the actual domain and TLD not matching are desired behavior.

mutatrum: Ah, it verifies on URL, not the complete string. That makes sense, missed that.

mutatrum:
> Updates quicklink matching to be safer (case-insensitive + exact or suffix host matching) and adds m45core.com quicklink, including subdomains and sub-sub domains. I reverted changes and that automatically closed the PR it seems. I hate github.

I think you can keep committing to the branch you created, GitHub will pick it up. Didn't know it closed the PR if you revert everything, but then I would think you can just re-open it if you added changes again. In case it happens again :-)

main/http_server/axe-os/src/app/services/quicklink.service.ts
```diff
@@ -20,6 +20,7 @@ export class QuicklinkService {
   public getQuickLink(stratumURL: string, stratumUser: string): string | undefined {
     const user = stratumUser.split('.')[0];
 
+    // Match entries against a lowercased stratum URL.
```
WantClue, line 23: comment not needed there

PR #1588, opened 2026-03-02
Show BIP-110 signal in Block Header

mutatrum:
image


components/stratum/coinbase_decoder.c
```diff
@@ -153,6 +153,9 @@ esp_err_t coinbase_process_notification(const mining_notify *notification,
     result->user_value_satoshis = 0;
     result->decoding_enabled = decode_outputs;
 
+    // Detect BIP-110 signaling: check if bit 4 (0x00000010) is set in version
+    result->bip110_signaling = (notification->version & 0x00000010) != 0;
```
luke-jr: idk about ESP-Miner policies, but bit 4 is only RDTS for 7 more months - after that, it may get repurposed for another softfork.

mutatrum: Thanks, missed the signal expiry. Added.

PR #1593, opened 2026-03-04
fix: add default-jre to devcontainer for OpenAPI codegen

jayrmotta:
## Summary
Adds `default-jre` to the devcontainer Dockerfile so OpenAPI code generation can run inside the container.

## Problem
OpenAPI code generation requires Java. The firmware build was failing in the Docker devcontainer because the JRE was not installed.

## Solution
Install `default-jre` in the base image dependency list in `.devcontainer/Dockerfile`.

## Testing
- Build the devcontainer: `docker build -t espminer-build .devcontainer`
- Run a full build inside the container: `idf.py build`
- Confirm OpenAPI/codegen steps complete and the build succeeds.

### DIFF
--- a/components/stratum/stratum_api.c
+++ b/components/stratum/stratum_api.c
@@ -258,6 +258,8 @@ void STRATUM_V1_parse(StratumApiV1Message * message, const char * stratum_json)
                         message->error_str = strdup(cJSON_GetStringValue(error_msg));
                     }
                 }
+            } else if (cJSON_IsString(error_json)) {
+                message->error_str = strdup(cJSON_GetStringValue(error_json));
             }
 
         // if the result is a boolean, then parse it
@@ -274,6 +276,8 @@ void STRATUM_V1_parse(StratumApiV1Message * message, const char * stratum_json)
                 message->error_str = strdup("unknown");
                 if (cJSON_IsString(reject_reason_json)) {
                     message->error_str = strdup(cJSON_GetStringValue(reject_reason_json));
+                } else if (cJSON_IsString(error_json)) {
+                    message->error_str = strdup(cJSON_GetStringValue(error_json));
                 }                
             }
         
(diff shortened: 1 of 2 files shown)


mutatrum:
Should also be added to README.md

https://github.com/bitaxeorg/ESP-Miner?tab=readme-ov-file#prerequisites-1

mutatrum:
Alternatively, change to a non-Java based generator?

jayrmotta:
> Alternatively, change to a non-Java based generator?

I did some research on [OpenAPI's website](https://openapi-generator.tech/docs/installation/) and all the alternatives also rely on Java. The only option that doesn't rely on Java is the Docker one, which internally also depends on Java, and then our Dockerfile would have to install Docker which would then install Java, not sure if that's better, I feel it's worse.

I also checked the bash launcher, that's just a shell script wrapping the same Java executable.

Some other alternatives like `openapi-typescript` or `orval` also generate compatible types, but they don't generate the clients like `ApiModule`, `SystemService`, etc, so that would require changing the frontend to use the generated types with wrappers to communicate over the network.

mutatrum:
> I did some research on [OpenAPI's website](https://openapi-generator.tech/docs/installation/) and all the alternatives also rely on Java. The only option that doesn't rely on Java is the Docker one, which internally also depends on Java, and then our Dockerfile would have to install Docker which would then install Java, not sure if that's better, I feel it's worse.
> 
> I also checked the bash launcher, that's just a shell script wrapping the same Java executable.
> 
> Some other alternatives like `openapi-typescript` or `orval` also generate compatible types, but they don't generate the clients like `ApiModule`, `SystemService`, etc, so that would require changing the frontend to use the generated types with wrappers to communicate over the network.

I tried to refactor the project to use `ng-openapi-gen`, which is a `node.js` based generator. The generated code is fairly close to what we have now, as far as I can tell. See #1597.

0xf0xx0:
superseded by #1597

PR #1594, opened 2026-03-04
Add support for fractional difficulty in mining.set_difficulty

### LINKED ISSUE #1592: mining.set_difficulty with fractional values are unsupported
Summary (v2.13.0, and below)

When a stratum host sends mining.set_difficulty with a fractional value (e.g., [0.001], [0.5], [2.5]), AxeOS truncates it to an integer because the entire difficulty pipeline uses uint32_t.

Affected code path:

Parsing (components/stratum/stratum_api.c:387):

uint32_t difficulty = cJSON_GetArrayItem(params, 0)->valueint;
Uses valueint (integer accessor) instead of valuedouble. A pool difficulty of 0.001 becomes 0.

Message struct (components/stratum/include/stratum_api.h:68):

uint32_t new_difficulty;
Global state (main/global_state.h:133):

uint32_t pool_difficulty;
Suggest difficulty (components/stratum/stratum_api.c:454):

int STRATUM_V1_suggest_difficulty(esp_transport_handle_t transport, int send_uid, uint32_t difficulty)
Formats with %ld — integer only.

NVS config (main/global_state.h:66):

uint16_t pool_difficulty;
Impact: Any pool that uses fractional difficulty values — common for BTC/BCH pools serving very low-hashrate miners like NerdMiner or small solo setups — will have difficulty truncated to 0 for sub-1 values, or lose the fractional component for values like 2.5 (becomes 2). A difficulty of 0 could cause undefined behavior or share rejection depending on how downstream code handles it.

Suggested fix: Change the difficulty type from uint32_t to double (or float) across the pipeline — parsing should use valuedouble from cJSON, the struct fields and global state should store as double, and formatting should use %f or %g.

mutatrum:
Fixes #1592

### DIFF
--- a/components/asic/asic.c
+++ b/components/asic/asic.c
@@ -21,13 +21,13 @@ uint8_t ASIC_init(GlobalState * GLOBAL_STATE)
 
     switch (GLOBAL_STATE->DEVICE_CONFIG.family.asic.id) {
         case BM1397:
-            return BM1397_init(GLOBAL_STATE->POWER_MANAGEMENT_MODULE.frequency_value, GLOBAL_STATE->DEVICE_CONFIG.family.asic_count, GLOBAL_STATE->DEVICE_CONFIG.family.asic.difficulty);
+            return BM1397_init(GLOBAL_STATE);
         case BM1366:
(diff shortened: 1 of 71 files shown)


mutatrum:
> tested ack, we should support fractional diffs in the user-configured suggested diff as well (right now it gets truncated)

For the hashrates we have here, fractions for the suggested diff is not useful. We can't go below 256 anyways due to how the ASIC is controller, and for any sensible suggested diff the fraction is irrelevant. This PR is only for pools that have a vardiff that end up in fractionals, all the main implementations only do integers, so it's a bit of an edge-case PR.

PR #1596, opened 2026-03-05
self_test domain monitoring

benjamin-wilson:
This enhances the self test to check the health of chip domains. If a domain is wildly out of tolerance (-33% or +300%) the test will fail. Currently it is possible for chips with a failing domain to pass the self test due to the nature of the probabilistic measurements.


benjamin-wilson:
Confirmed working on 602,650 and 801

mutatrum:
This looks like quite a bit of code duplication, isn't it possible to re-use the code from hashrate_monitor?

PR #1601, opened 2026-03-09
Improve factory self-test flow for Bitaxe Touch accessories

SoloSatoshi:
## Summary

- change the post-pass behavior from requiring a manual reset to a 10-second auto-restart countdown after a
    successful factory self-test
- leave failed self-tests latched on-screen for diagnosis
- publish self-test state over BAP while self-test is active

## Why

On Bitaxe Touch hardware, the Bitaxe reset button is enclosed and not easily accessible during the factory-flash and self-test flow.

The auto-restart change removes the need for a hidden manual reset after a passing self-test. The added BAP self-test status lets accessories distinguish:

- self-test is actively running
- self-test passed and the miner is about to restart
- self-test failed and is intentionally staying on the fail screen

Without this, external accessories only see the Bitaxe disappear and reappear and can only infer a generic reconnect state.

## Scope

This PR is only the ESP-Miner side of the change. Accessory UI handling for the emitted `self_test` BAP status is implemented separately.

### DIFF
--- a/main/bap/bap_uart.c
+++ b/main/bap/bap_uart.c
@@ -111,6 +111,9 @@ void BAP_send_init_message(GlobalState *state) {
 
 void BAP_send_ap_message(GlobalState *state) {
     BAP_send_message(BAP_CMD_CMD, "mode", "ap_mode");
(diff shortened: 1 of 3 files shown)


SoloSatoshi:
Closing this to restart from upstream v2.13.1 with a clean fork.

Issue #1587, opened 2026-03-02
esp-miner is unable to handle difficulty change following with "mining.notify" when `cleanJob` set to false


blackmennewstyle:
Hello beautiful dev(s),

Why is `esp-miner` unable to handle properly the following scenario:

- The stratum sends a `mining.set_difficulty` following with `mining.notify`  with the`cleanJob` indicator set to `false`.

`esp-miner` then starts to send duplicate shares, the same solution with the same `nonce`, over and over again until the stratum sends another `mining.notify` with the `cleanJob` indicator set to `true`.

The behavior is clearly visible in my attachment file.

It is way more easy to reproduce that behavior on BTC `testnet4` by the way.
```
2026-03-02 16:52:49.4134] [I] [btc1] [0HNJOFCT4H9CS] VarDiff update to 9427.961
[2026-03-02 16:52:49.4134] [D] [btc1] [0HNJOFCT4H9CS] Sending: {"jsonrpc":"2.0","method":"mining.set_difficulty","params":[9427.960695338843],"id":null} 
[...]

PR #1595, opened 2026-03-05
Remove clean_jobs check in create_jobs_task

DankMiner:
Remove unnecessary check for clean_jobs in job creation task.

When a `mining.notify` with `cleanJob=false` was received (e.g. after a `mining.set_difficulty` or new mempool transactions), `create_jobs_task` skipped calling `generate_work()` due to an early `continue`. This meant the ASIC never received the new job and kept mining the previous nonce space. Once the ASIC exhausted and wrapped its nonce range, it resubmitted previously found solutions, causing persistent duplicate share rejections until the next `cleanJob=true` notify arrived.

The `cleanJob` flag is already handled correctly in `stratum_task.c`, which flushes the stratum queue when `cleanJob=true`. The flag should only control whether queued jobs are discarded — not whether the ASIC receives new work.

Removed the `cleanJob=false` guard in `create_jobs_task.c` so that every `mining.notify` results in `generate_work()` being called.

Fixes #1587

PR #1603, opened 2026-03-11
fix: reset extranonce2 only if clean_jobs == true

blackmennewstyle:
This PR improves the extranonce2 reset.
It is only performed if clean_jobs == true.
The previous code was triggering a cascade of duplicate shares after a varDiff change:
- mining.set_difficulty
- mining.notify with clean_jobs is set to false The behavior is described here: https://github.com/bitaxeorg/ESP-Miner/issues/1587#issuecomment-4035872469


main/tasks/create_jobs_task.c
```diff
@@ -63,10 +63,10 @@ void create_jobs_task(void *pvParameters)
                 GLOBAL_STATE->new_stratum_version_rolling_msg = false;
             }
 
-            extranonce_2 = 0;
-
-            if (!current_mining_notification->clean_jobs) {
-                continue;
+            // reset extranonce2 only if clean_jobs == true
+            // avoid a cosmic festival of duplicated shares
```
adammwest: `// reset extranonce2 only if clean_jobs == true` The code already shows this.
I think its better to explain a decision, so the cause of the duplicates that led to this PR
so in the future when you look the code again its obvious.

blackmennewstyle: So should i close this PR and make a new one?

I changed it, hopefully it is better.

blackmennewstyle:
@mutatrum Hi there, should i close my PR? It is obvious that you guys don't want to fix the issue with the `extranonce2` reset according to the proper `cleanJob` flag.

At least be honest and give a reason for the lack of interest. I'm not the only one affected with that issue, that pool operator is also impacted https://github.com/bitaxeorg/ESP-Miner/pull/1595#issue-4029458532 and literally everyone using the stratum software `miningcore`.

That 8 lines of code does not hurt the other stratum software either in any single way, i tested on `CKPOOL` and SOLOPOOL.

blackmennewstyle:
New PR https://github.com/bitaxeorg/ESP-Miner/pull/1694 :  Based on your StratumV2 which allegedly was supposed to fix the issue but unfortunately, the root of the problem `extranonce2 = 0` which naively resets for every single job without taking in account the `cleanJob` state, has NOT been tackled within the `STRATUM V2` changes.

mutatrum:
> @mutatrum Hi there, should i close my PR? It is obvious that you guys don't want to fix the issue with the `extranonce2` reset according to the proper `cleanJob` flag.
> 
> At least be honest and give a reason for the lack of interest. I'm not the only one affected with that issue, that pool operator is also impacted [#1595 (comment)](https://github.com/bitaxeorg/ESP-Miner/pull/1595#issue-4029458532) and literally everyone using the stratum software `miningcore`.
> 
> That 8 lines of code does not hurt the other stratum software either in any single way, i tested on `CKPOOL` and SOLOPOOL.

It's more that we haven't seen it on the pools we normally test, and nobody has tried to reproduce it on your pool yet. Lack of time, lack of users reporting the issue and it's unclear if it is the correct way of fixing it, if there is anything to fix. Ckpool also does vardiff and it doesn't happen there.

It's good you reopened it against master with SV2 merged, and now that's out of the way we can have a look at this.

PR #1606, opened 2026-03-12
Bump up fan frequency to 19kHz on EMC2103

raivisr:
Default EMC2103 PWM frequency is 2.441 kHz, some fans emit high pitched noise when driven with such low PWM frequency. This bumps up PWM frequency to 19.531 kHz, eliminates high pitched noise on those specific fans and should still work well with any other fan.

### DIFF
--- a/main/thermal/EMC2101.c
+++ b/main/thermal/EMC2101.c
@@ -30,6 +30,7 @@ esp_err_t EMC2101_init(int temp_offset_param)
     // set the TACH input
     ESP_RETURN_ON_ERROR(i2c_bitaxe_register_write_byte(emc2101_dev_handle, EMC2101_REG_CONFIG, 0x04), TAG, "Failed to set TACH input");
 
+
     //! deprecated bit was never set successfully
     // --- Configure Fan Settings (Register 0x4A) ---
     // Start with a base configuration:
@@ -42,7 +43,6 @@ esp_err_t EMC2101_init(int temp_offset_param)
     // //if (invertPolarity) {
         ESP_RETURN_ON_ERROR(i2c_bitaxe_register_write_byte(emc2101_dev_handle, EMC2101_FAN_CONFIG, 0b00100011), TAG, "Failed to configure fan settings");
     // //}
-    ESP_RETURN_ON_ERROR(i2c_bitaxe_register_write_byte(emc2101_dev_handle, EMC2101_PWM_FREQ, 0x07), TAG, "Failed to set PWM frequency");
     
 
 
@@ -74,7 +74,7 @@ esp_err_t EMC2101_set_fan_speed(float percent)
 {
     uint8_t speed;
 
-    speed = (uint8_t) (14.0 * percent);
+    speed = (uint8_t) (63.0 * percent);
     ESP_RETURN_ON_ERROR(i2c_bitaxe_register_write_byte(emc2101_dev_handle, EMC2101_REG_FAN_SETTING, speed), TAG, "Failed to set fan speed");
     return ESP_OK;
 }


raivisr:
On second thought, it is probably even better to set the value to 0x00, or 26kHz. Intel 4 pin fan spec calls for 25Khz PWM frequency, with values 21-28kHz being acceptable.

WantClue:
A problem that this might face is that we don't know if all fans will work with such a high pwm signal. Manufacturers use all sorts of fans so it's impossible for us to figure out an ideal frequency. But this PR will be tested

raivisr:
Is it safe to assume that most if not all manufacturers make their fans to Intel spec? Also, we tested everything that we had at hand and all fans work just fine. Also, it is a PWM, a digital way to produce approximation of analog voltage, the higher the frequency, the closer the approximation.

WantClue:
Sounds fair, I'll give this a shot with 3 - 4 fan brands that I have laying around.

raivisr:
if that works for this fan controller, it is probably good idea to update all the others as well. might improve commanded vs observed fan speed tracking.

WantClue:
I was not able to reduce any high pitch noise. Can you elaborate a bit further on the brands of fan's you're using and at what fan speed you get a high pitch for reproducablity ?

jzpm:
Hi! The fan which made high pitch noise was from Gelid brand (6025, dual ball bearing, 4000 RPM fan). High pitch was there for any RPM except when running at 100%. Best to hear at 80-90% of max RPM. We can send a sample for you to see.

mutatrum:
It would make sense to put the base frequency inside the actual spec range, instead of at 19 Khz. And it might make sense to investigate the other fan controllers as well. Question is if we should make it configurable, in case the updated config messes with someones device?

WantClue:
Please adjust it to a more broadly used frequency which should be in the range of 21-25khz

raivisr:
If there are no objections, I propose the following course of action. I will close this PR, go through all the fan controllers, update all of them to Intel spec PWM frequency range (if they support it) and create a new PR.
At this time I do not think option to configure is necessary. If anyone has any problems with their fans acting weird, that can be added later.
@WantClue @mutatrum ACK/NAK?

mutatrum:
> If there are no objections, I propose the following course of action. I will close this PR, go through all the fan controllers, update all of them to Intel spec PWM frequency range (if they support it) and create a new PR. 

No need to close this, just commit to this branch and it'll be fine. You can convert the PR to draft until it's ready for review again. That way it stays visible.

> At this time I do not think option to configure is necessary. If anyone has any problems with their fans acting weird, that can be added later. @WantClue @mutatrum ACK/NAK?

Agree. The limited research I did basically pointed to the fact that virtually all fans follow the Intel spec, so they _should_ be ok without configuration options.

raivisr:
Updated all 3 of controllers to work within the Intel spec PWM frequency. EMC2101 is a tricky one, changing the PWM frequency changed effective fan setting resolution to roughly 7% per step, so I changed the mapping of fan % as well. Unfortunately I don't have anything on hand to test this.

mutatrum:
> Updated all 3 of controllers to work within the Intel spec PWM frequency. EMC2101 is a tricky one, changing the PWM frequency changed effective fan setting resolution to roughly 7% per step, so I changed the mapping of fan % as well. Unfortunately I don't have anything on hand to test this.

Hmm, that's not good, 7% is way too coarse. I have EMC2101 devices so I can test that.

raivisr:
That's roughly 14 steps, so not very fine granularity. Please test, if it is total garbage, I will revert that one back.

main/thermal/EMC2101.c
```diff
@@ -74,7 +74,7 @@ esp_err_t EMC2101_set_fan_speed(float percent)
 {
     uint8_t speed;
 
-    speed = (uint8_t) (63.0 * percent);
+    speed = (uint8_t) (14.0 * percent);
```
mutatrum: This change is not correct, it seems changing the base frequency doesn't alter this.

raivisr: according to datasheet it should, but since I have no device with EMC2101 on hand, I'll revert this particular driver and only update the other two.

mutatrum: I don't fully understand the datasheet either, but the fan was definitely off with this change and it's fine without this.

PR #1597, opened 2026-03-05
Migrate openapitools to ng-openapi-gen

mutatrum:
This eliminates the build time dependency on Java

PR #1608, opened 2026-03-13
Pause/resume mining

### LINKED ISSUE #734: Enable / Disable TPS - Start / Stop mining
Following OSMU townhall #29 discussion 

It would be very useful to have the ability to enable/ disable the onboard TPS that feeds the asic. 

this would allow for mining to be commanded by an external source to start/ stop mining whilst keeping the ESP and AxeOs alive , so the user can maintain connectivity, stats, swarm etc. 

a use case scenario is the upcoming solarbit $0/khw project. Currently i will cut the entire power to the bitaxe but would be better to just instruct the bitaxe via the BAP port to stop mining when sun is down.. and start mining when sun is up. a cleaner way that killing the entire miner. 

Summary: 

1) API command to start stop TPS 
2) BAP port ability to accept such commands from external esp32 hardwired via 6pin BAP 
3) Not sure if possible: but if no mining is being conducted .. we dont need active cooling on the asic.. therefore is it possible when TPS is commanded to stop power to asic to also terminate power to the fan ? worst case pwm to minimum ? thinking how we could maximize energy efficiency by only running essentials on bitaxe to maintain heartbeat. fan running for no reason ruins the efficiency and wastes power.

b-rowan:


b-rowan:
Tested on BitAxe 401, everything seems to work locally.

main/http_server/http_server.c
```diff
@@ -1308,8 +1369,8 @@ esp_err_t start_rest_server(void * pvParameters)
     httpd_config_t config = HTTPD_DEFAULT_CONFIG();
     config.uri_match_fn = httpd_uri_match_wildcard;
     config.stack_size = 8192;
-    config.max_open_sockets = 20;
-    config.max_uri_handlers = 24;
+    config.max_open_sockets = 13;
```
johnny9: I'm guessing you didn't mean to change these. Change these back to the original values

main/http_server/http_server.c
```diff
@@ -1393,8 +1454,16 @@ esp_err_t start_rest_server(void * pvParameters)
     };
     httpd_register_uri_handler(server, &system_restart_options_uri);
 
+    httpd_uri_t system_mining_pause_uri = {
+        .uri = "/api/system/mining",
+        .method = HTTP_POST,
+        .handler = POST_mining_pause,
+        .user_ctx = rest_context
+    };
+    httpd_register_uri_handler(server, &system_mining_pause_uri);
```
johnny9, line 1447: any reason to not just use the system PATCH api and PATCH the miningPaused value?

johnny9 (commented):
Concept ACK

Whether you decide to continue with the additional POST endpoint or make use of the system api PATCH, you should update the openapi.yaml (https://github.com/bitaxeorg/ESP-Miner/blob/master/main/http_server/openapi.yaml)

Tested on my 201 by hitting pause/resume in the AxeOS web you a couple of times and it works well.

main/http_server/http_server.c
b-rowan: For some reason having this on the initial values causes the web UI to fail to start.  The error message complains about these variables, but I'll change it back and replicate it so I can send a more full log here.

main/http_server/http_server.c
b-rowan: I'm OK with that, but I think of patch more for configuration rather than direct control.  For example, restart uses post, not patch, and I think about pausing and resuming as more akin to that than changing pool settings for example.

Let me know what your preference is.

b-rowan:
RE d09a1ee07eea4890615a7a9bae4083402702e536:

Getting this error when building in the devcontainer:

```
Error: /bin/sh: 1: java: not found

    at /workspace/main/http_server/axe-os/node_modules/@openapitools/openapi-generator-cli/main.js:2:49316
    at ChildProcess.exithandler (node:child_process:424:5)
    at ChildProcess.emit (node:events:519:28)
    at maybeClose (node:internal/child_process:1101:16)
    at Socket.<anonymous> (node:internal/child_process:456:11)
    at Socket.emit (node:events:519:28)
    at Pipe.<anonymous> (node:net:346:12)
```

Added the default JRE to the devcontainer to fix this.

main/http_server/http_server.c
b-rowan: Here is the exact error:

```
I (2305) http_server: Starting HTTP Server
E (2305) httpd: Config option max_open_sockets is too large (max allowed 13, 3 sockets used by HTTP server internally)
        Either decrease this or configure LWIP_MAX_SOCKETS to a larger value
E (2325) http_server: start_rest_server(1378): Start server failed
```

johnny9: your sdkconfig is not right. remove the file, rebase, then rebuild to generate a new one

main/http_server/http_server.c
johnny9: I think I agree. My first intuition was that this was a configuration but that isn't the case.

main/http_server/http_server.c
```diff
@@ -792,6 +792,66 @@ static esp_err_t POST_dismiss_block_found(httpd_req_t * req)
...
+    }
+
+    if (set_cors_headers(req) != ESP_OK) {
+        httpd_resp_send_500(req);
+        return ESP_OK;
+    }
+
+    int total_len = req->content_len;
+    int cur_len = 0;
+    char * buf = ((rest_server_context_t *) (req->user_ctx))->scratch;
```
johnny9: Extract a method to handle this parsing the json request body. This is duplicated in the PATCH endpoint

b-rowan: Done!

main/http_server/http_server.c
b-rowan: Yep, that fixed it.  Changed those settings back.

main/power/asic_init.c
```diff
@@ -42,6 +43,12 @@ uint8_t asic_initialize(GlobalState *GLOBAL_STATE, asic_init_mode_t mode, uint32
         vTaskDelay(100 / portTICK_PERIOD_MS);
     }
 
+    // Reset the frequency transition tracker so the ramp starts from
+    // the chip's post-reset default (50 MHz) rather than the stale
+    // pre-reset frequency, which would cause do_frequency_transition
+    // to skip the ramp entirely (seeing current == target).
+    reset_frequency_transition_state();
```
mutatrum: Shouldn't this be done in mining_control.c, somewhere in mining_stop? asic_init doesn't do anything with frequencies.

Or maybe even better in `power_management_task.c`, as that's the guardian of frequency :-)

b-rowan: power_management_task.c calls the mining control functions, so I think putting it in mining_control.c makes sense.  I'll have it fixed shortly.

main/http_server/openapi.yaml
```diff
@@ -819,6 +823,45 @@ paths:
         '500':
           description: Internal server error
 
+  /api/system/mining:
```
mutatrum: Not sure if a post body follows the rest of the API design here. Maybe `/api/system/pause` and `/api/system/resume` are cleaner? Or `/api/system/mining/pause` and `resume`, or stop/start.

Also add new endpoint(s) to README.md.

b-rowan: Yeah that's fine by me, should be simple enough to change.

johnny9 (approved):
ACK b789bca09c6581c3264777e58e4ed741be236835

Tested on 201

.devcontainer/Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN apt-get update \
   libglib2.0-dev \
   libnuma1 \
   libpixman-1-dev \
+  default-jre \
```
mutatrum: Please remove from PR, this is fixed in either #1593 or #1597.

mutatrum:
It works, but a few remarks:

 * When paused, the log prints a warning every 500ms:
   `W (39219) create_jobs_task: ASIC not initialized, skipping job send`
   This is unwanted;
 * As the temp sensor is no longer functioning, the fan defaults to 70%:
   `I (50739) fan_controller: Temperature sensor starting up, setting fan to 70%`
   I think the fan can just be set to 0% I guess? Not sure if this is ok on all models though, that needs verification;
 * Not sure what to do with the stratum connection. IMO it makes sense to disconnect, as most pools will probably close the connection anyways if no shares are coming in, and my guess is it'll either go to the fallback pool or retry the stratum connect. Anyways, it's nicer to disconnect and not tie up pool resources for a paused miner;
 * Paused state needs a bit more feedback on the dashboard and on the swarm page. Pause/play button could also be added on swarm entries.

Nice feature addition, thanks for contributing!

main/http_server/axe-os/src/app/services/system.service.ts
```diff
@@ -10,7 +10,7 @@ import {
   SystemASIC as ISystemASIC,
   SystemASICASICModelEnum,
   SystemService as GeneratedSystemService,
-  Settings
+  Settings,
```
mutatrum, line 13: Try to limit unrelated changes, to limit merge conflicts in the future.

main/http_server/http_server.c
```diff
@@ -632,37 +632,44 @@ bool check_settings_and_update(const cJSON * const root)
     return result;
 }
 
-static esp_err_t PATCH_update_settings(httpd_req_t * req)
+static esp_err_t http_server_read_request_body(httpd_req_t * req)
```
mutatrum: The extraction of this method also seems unrelated? Other than the return value and removal of comment the function is the same, IINM.

There are quite a few PRs with changes in `http_server.c`, so here as well, try to limit the changes.

main/power/mining_control.c
```diff
@@ -0,0 +1,69 @@
...
+    asic_hold_reset_low();
+
+    // Mark uninitialized immediately so tasks stop issuing UART commands
+    GLOBAL_STATE->ASIC_initalized = false;
+
+    // Reset the frequency transition tracker so the ramp starts from
+    // the chip's post-reset default (50 MHz) rather than the stale
+    // pre-reset frequency, which would cause do_frequency_transition
+    // to skip the ramp entirely (seeing current == target).
+    reset_frequency_transition_state();
```
mutatrum: Instead of `reset_frequency_transition_state()`, maybe it's possible to handle this in `power_management_task.c` itself, to call `ASIC_set_frequency` with 50 to wind down the frequency. I wonder if we actually need these new files then?

Or maybe move the functions into power_management_task.c, as that's the only file that's calling these methods.

main/http_server/openapi.yaml
```diff
@@ -819,6 +823,56 @@ paths:
...
+    post:
+      summary: Pause mining
+      description: Pauses mining activity
+      operationId: pauseMining
+      tags:
+        - system
+      responses:
+        '200':
+          description: Mining paused
+          content:
```
mutatrum, line 836: If you return a GenericResponse with a string, the code can be simpler and slightly more in line with the other action calls. You can add pause/resume on swarm entries, where the `message` field from the response is shown in a toastr.

mutatrum (changes requested):

main/http_server/axe-os/src/app/layout/app.topbar.component.html
```diff
@@ -37,6 +37,15 @@
                     <wifi-icon [rssi]="info.wifiRSSI" />
                 </a>
             </li>
+            <li>
+                <a class="block py-2 text-white cursor-pointer"
+                    tooltipPosition="bottom"
+                    [pTooltip]="isMiningPaused ? 'Resume Mining' : 'Pause Mining'"
+                    (click)="toggleMiningPaused()"
+                >
+                    <i class="pi text-xl block" [ngClass]="isMiningPaused ? 'pi-play' : 'pi-pause'"></i>
```
mutatrum: Can you try to see how it looks with `pi-play-circle` and `pi-pause-circle`? The pause icon is a bit skinny 😆

main/http_server/http_server.c
b-rowan: https://github.com/bitaxeorg/ESP-Miner/pull/1608#discussion_r2937762659

b-rowan:
Seems like there is still some work to be done here.  I'll do what I can, thanks for the feedback!

main/http_server/http_server.c
b-rowan: Happy to revert this if you think it is necessary, LMK.

b-rowan:
@mutatrum Fixed all comments but the one, let me know what's best to do there.  Swarm page now shows more information, a pause/resume button, and the icons for letting a user know its paused.  All the log spamming should be gone, and the pool stratum mining task is killed when mining is paused, then restarted when it is resumed.

mutatrum:
> @mutatrum Fixed all comments but the one, let me know what's best to do there. Swarm page now shows more information, a pause/resume button, and the icons for letting a user know its paused. All the log spamming should be gone, and the pool stratum mining task is killed when mining is paused, then restarted when it is resumed.

Thanks, I'll have a look. Sorry to harp on about this, but please only force-push as a last resort of getting yourself out of a git cesspit. A force-push delinks previous review comments and makes it harder to review what has changed. For a running PR, it's probably easier to keep committing and hold off on merging master until review comments are fully resolved. (Only exception is if there is a specific PR on master that needs to be added to the PR, but I don't think that's the case here.)

main/http_server/http_server.c
mutatrum: Oh didn't see johhny9's comment. Not sure what happened then, as the extracted method is only used once now?

Background is #1240, that'll impact `http_server.c` quite a bit so if the change surface can be as small as possible here, that would make either PRs easier to merge. 

Adding new endpoints or handlers is fine of course, as otherwise you can't make this PR, but those are also easy to merge in. Moving code around is tricky.

b-rowan:
> Sorry to harp on about this, but please only force-push as a last resort of getting yourself out of a git cesspit.

This seems to be repo by repo, I am very used to force pushing to try to keep the history clean and make each commit make sense so it can be rebated instead of squashed.  I suppose in this case it will be squashed so it doesn't matter.  I have just been doing `fixup`commits to fix things I find to try to combine commits, but I suppose it makes sense in this case to just leave the history messy until it is squashed.

main/http_server/http_server.c
b-rowan: Should be fixed now, I removed the body parsing function.

b-rowan:
Ok, all comments should be resolved now.  Let me know if there is anything else!

components/asic/include/frequency_transition_bmXX.h
```diff
@@ -27,4 +27,13 @@ typedef void (*set_hash_frequency_fn)(float frequency);
...
 void do_frequency_transition(float target_frequency, set_hash_frequency_fn set_frequency_fn);
 
+/**
+ * @brief Reset the tracked frequency state to the initial value (50 MHz)
+ *
+ * Call this before re-initializing an ASIC after a reset so that
+ * do_frequency_transition ramps from the chip's post-reset default
+ * instead of the stale pre-reset value.
+ */
+void reset_frequency_transition_state(void);
```
mutatrum: Function is no longer used.

mutatrum:
> Ok, all comments should be resolved now. Let me know if there is anything else!

Code looks really clean now, well done & thanks! I'll have another test run tomorrow. 

Only other tiny (nice to have) thing I can think of is to add a pause indicator to the display as well. Take a look at `notification_label` in `screen.c` (line 599-619). Not sure what character to use though for pause. This is the font used: https://int10h.org/oldschool-pc-fonts/fontlist/font?portfolio_6x8, so maybe double line (║) or 2x half block (▐▐) or something like that?

b-rowan:
> > Ok, all comments should be resolved now. Let me know if there is anything else!
> 
> Code looks really clean now, well done & thanks! I'll have another test run tomorrow.
> 
> Only other tiny (nice to have) thing I can think of is to add a pause indicator to the display as well. Take a look at `notification_label` in `screen.c` (line 599-619). Not sure what character to use though for pause. This is the font used: [int10h.org/oldschool-pc-fonts/fontlist/font?portfolio_6x8](https://int10h.org/oldschool-pc-fonts/fontlist/font?portfolio_6x8), so maybe double line (║) or 2x half block (▐▐) or something like that?

Added it, I'll test it out on my 401 when I'm home later, but should be fine I think.

mutatrum:
Related: #911

b-rowan:
> Related: #911


Not really sure what would need to be done to work with that, but my thought is that that can just be built on top of what I've added here?  The change to stratum_task.c was pretty small, so hopefully should be simple enough, but I think creating a general status task for handling all this might be a bit outside my abilities...

mutatrum:
> Not really sure what would need to be done to work with that...

Sorry, my comment was unclear. I just added it as reference for that issue, no action needed for this PR. This PR makes implementing that issue quite a bit easier.

b-rowan:
Tested and working on 401, the little pause icon shows up in the top right section of the screen.

mutatrum (approved):
Code lgtm, will do one more round of testing.

main/http_server/http_server.c
```diff
@@ -792,6 +792,58 @@ static esp_err_t POST_dismiss_block_found(httpd_req_t * req)
...
+        return ESP_OK;
+    }
+
+    GLOBAL_STATE->SYSTEM_MODULE.mining_paused = true;
+    ESP_LOGI(TAG, "Mining paused by API request");
+
+    httpd_resp_set_type(req, "application/json");
+    cJSON * resp = cJSON_CreateObject();
+    if (resp == NULL) {
+        httpd_resp_send_err(req, HTTPD_500_INTERNAL_SERVER_ERROR, "Memory allocation failed");
```
WantClue: change this to internal error not "Memory allocation failed"

main/http_server/http_server.c
```diff
@@ -792,6 +792,58 @@ static esp_err_t POST_dismiss_block_found(httpd_req_t * req)
...
+        return ESP_OK;
+    }
+
+    GLOBAL_STATE->SYSTEM_MODULE.mining_paused = false;
+    ESP_LOGI(TAG, "Mining resumed by API request");
+
+    httpd_resp_set_type(req, "application/json");
+    cJSON * resp = cJSON_CreateObject();
+    if (resp == NULL) {
+        httpd_resp_send_err(req, HTTPD_500_INTERNAL_SERVER_ERROR, "Memory allocation failed");
```
WantClue: same here

main/tasks/fan_controller_task.c
```diff
@@ -52,7 +52,15 @@ void FAN_CONTROLLER_task(void * pvParameters)
...
-        if (nvs_config_get_bool(NVS_CONFIG_OVERHEAT_MODE)) {
+        if (GLOBAL_STATE->SYSTEM_MODULE.mining_paused) {
+            if (fabs(power_management->fan_perc - 30) > EPSILON) {
+                ESP_LOGI(TAG, "Mining paused, setting fan to 30%%");
+                power_management->fan_perc = 30;
+                if (Thermal_set_fan_percent(&GLOBAL_STATE->DEVICE_CONFIG, 0.3f) != ESP_OK) {
+                    exit(EXIT_FAILURE);
+                }
+            }
+        } else if (nvs_config_get_bool(NVS_CONFIG_OVERHEAT_MODE)) {
```
mutatrum: This should be the first clause, overheat is more important than pausing.

main/tasks/fan_controller_task.c
```diff
@@ -52,7 +52,15 @@ void FAN_CONTROLLER_task(void * pvParameters)
     TickType_t taskWakeTime = xTaskGetTickCount();
 
     while (1) {
-        if (nvs_config_get_bool(NVS_CONFIG_OVERHEAT_MODE)) {
+        if (GLOBAL_STATE->SYSTEM_MODULE.mining_paused) {
```
WantClue: overheat mode is more important

main/http_server/http_server.c
b-rowan: All of the messages covering that same clause are `"Memory allocation failed"`, so this was set to be consistent with them, but it's changed now.

jkl2e:
Is it possible to send the parameter miningPaused by /api/system ?
I tried to send true/false and 'true'/'false' and 1/0 but it doesn't work.
So it is only available in the webui?

I want to enable /disable automatically depending on, Solar power is available or not.

mutatrum:
> Is it possible to send the parameter miningPaused by /api/system ? I tried to send true/false and 'true'/'false' and 1/0 but it doesn't work. So it is only available in the webui?

Nothing is webui only, that uses the API for everything.

> I want to enable /disable automatically depending on, Solar power is available or not.

See readme:
```
# Pause mining
curl -X POST http://YOUR-BITAXE-IP/api/system/pause

# Resume mining
curl -X POST http://YOUR-BITAXE-IP/api/system/resume
```

jkl2e:
Ahh thank you, I build "Profile settings", und picked the parameters from /api/system/info.
For this parameter, I will have to change the way my scirpt is working.

jkl2e:
Fanspeed jumps to 30% when paused. How can I set it to 0% ?

b-rowan:
> Fanspeed jumps to 30% when paused. How can I set it to 0% ?

You can't, that value is hard coded when paused.

Also, please open an issue or discussion for this, this PR has already been merged, so this is not the place for discussion.

PR #1609, opened 2026-03-14
Show BIP-54 signal in Block Header

mutatrum:
Now BIP-54 has been signalled in the wild, adding this to the dashboard, similar to BIP-110 signalling.


WantClue (approved):

luke-jr:
Uh, this is incorrect. Having a given locktime does NOT signal BIP54...

mutatrum:
> Uh, this is incorrect. Having a given locktime does NOT signal BIP54...

The test is for `nLockTime` = block height - 1 and `nSequence` != `0xFFFFFFF`. Is that not correct?

PR #1620, opened 2026-03-20
Add streaming of display screen to dashboard

mutatrum:
Add a new http endpoint `/api/screen/stream` that serves an Animated PNG stream of the display:


mutatrum:
Probably needs to be revisited after #1623. Current solution is overly simple regarding disconnects.

bonifacio123:
Tested this on a 801, no issues observed. Worked as described.

PR #1621, opened 2026-03-20
Add cpu usage task and asic share processing time

mutatrum:
Add CPU Usage % on the system page.

image

It also measures the processing time for a share produced by the asic up to the submit to the pool.

There is also a more extensive `task_monitor_task` which is not started by default as this shouldn't run in production but that will show per task load.

This task monitor showed a relatively high usage in the `asic_result_task`, and by moving some log statement out of the critical path, the processing time dropped from 28ms to 2.7ms.

`I (353364) asic_result: Processing time: 2.7 ms`


main/task_monitor.c
```diff
@@ -0,0 +1,106 @@
...
+#define MAX_TASKS 40
+#define INTERVAL_MS 60000
+
+static const char* TAG = "task_monitor";
+
+void task_monitor_task(void *pvParameters) {
+    TaskStatus_t *task_array1 = malloc(MAX_TASKS * sizeof(TaskStatus_t));
+    TaskStatus_t *task_array2 = malloc(MAX_TASKS * sizeof(TaskStatus_t));
+    if (task_array1 == NULL || task_array2 == NULL) {
+        ESP_LOGE(TAG, "Failed to allocate memory for task arrays");
```
WantClue, line 16: Memory leak on partial allocation failure

add a free for task_array1 and 2

main/task_monitor.c
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    // Cleanup (though loop is infinite)
+    free(task_array1);
+    free(task_array2);
+    vTaskDelete(NULL);
+}
+
+void cpu_monitor_task(void *pvParameters) {
+    GlobalState *GLOBAL_STATE = (GlobalState *)pvParameters;
```
WantClue, line 90: If the task is created without passing a valid GlobalState pointer, dereferencing GLOBAL_STATE on line 102 will cause a crash

components/stratum/stratum_api.c
```diff
@@ -509,17 +509,23 @@ int STRATUM_V1_pong(esp_transport_handle_t transport, int message_id)
 /// @param version_bits The hex-encoded version bits set by miner (BIP310).
 int STRATUM_V1_submit_share(esp_transport_handle_t transport, int send_uid, const char * username, const char * job_id,
                             const char * extranonce_2, const uint32_t ntime,
-                            const uint32_t nonce, const uint32_t version_bits)
+                            const uint32_t nonce, const uint32_t version_bits, uint64_t arrival_time_us, uint64_t *out_sent_time_us)
```
WantClue: Unused parameter: arrival_time_us is never referenced

main/main.c
```diff
@@ -30,6 +31,9 @@ void app_main(void)
 {
     ESP_LOGI(TAG, "Welcome to the bitaxe - FOSS || GTFO!");
 
+    // xTaskCreate(task_monitor_task, "task_monitor", 8192, NULL, 1, NULL);
+    xTaskCreate(cpu_monitor_task, "cpu_monitor", 4096, (void *)&GLOBAL_STATE, 1, NULL);
```
WantClue: cpu_monitor_task is created immediately at startup and receives a pointer to GLOBAL_STATE, but the structure isn't initialized until later (e.g., SYSTEM_init_system at line 83, device_config_init at line 75). The task may read uninitialized or partially initialized fields, leading to undefined behavior.

add an if statement:
`if (xTaskCreate(cpu_monitor_task, "cpu_monitor", 4096, (void *)&GLOBAL_STATE, 1, NULL) != pdPASS) {
        ESP_LOGE(TAG, "Error creating cpu monitor task");
    }`

mutatrum: Good catch! Although it's static here, as well as substructures, so that's not an issue. `SYSTEM_init_system` only populates starting values, which in the case of `cpu_usage` is 0, which is fine.

main/task_monitor.c
mutatrum: This follows the rest of the code, we never guard against that and shouldn't be needed as GlobalState is static in `main.c`.

0xf0xx0:
not building for me currently

```
ESP-Miner/main/task_monitor.c:95:31: error: implicit declaration of function 'ulTaskGetIdleRunTimePercentForCore'; did you mean 'ulTaskGetIdleRunTimePercent'? [-Wimplicit-function-declaration]
   95 |         float idle_percent = (ulTaskGetIdleRunTimePercentForCore(0) + ulTaskGetIdleRunTimePercentForCore(1)) / 2.0f;
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                               ulTaskGetIdleRunTimePercent
[1604/1633] Building C object esp-idf/main/CMakeFiles/__idf_main.dir/http_server/http_server.c.obj```

mutatrum:
Delete `sdkconfig` and do a full rebuild.

0xf0xx0:
sdkconfig strikes again! will merge in a bit if no more nits pop up

bonifacio123:
I see the CPU % now - the only odd thing I noticed is that whenever I restart I see very large number for efficiency at first - after a little bit they settle to something more normal

Screenshot From 2026-04-07 20-46-24

0xf0xx0:
> I see the CPU % now - the only odd thing I noticed is that whenever I restart I see very large number for efficiency at first - after a little bit they settle to something more normal

im not able to reproduce this, can you give logs?

bonifacio123:
I reapplied the firmware and the large numbers happened again. By the time I screen captured the numbers settled some. I've been using this miner to test various PR's. Maybe something got smashed in nvram.

[debug.log](https://github.com/user-attachments/files/26604003/debug.log)

Screenshot From 2026-04-09 08-17-32

mutatrum:
It's an artefact from the startup procedure, which is now visible due to #1634, where there's a moment of very low hashrate:

image

This is unrelated, this also happens in `master` at the moment.

WantClue:
please resolve conflicts @mutatrum

PR #1622, opened 2026-03-21
Customizable dashboard

WantClue:


WantClue:
image

bonifacio123:
Tested this on a 601 - works great!  Thank you for doing this. The only odd thing I see is the menu is truncated and not all options are viewable on a mobile phone

Screenshot_20260406-082327~2

main/http_server/axe-os/src/app/components/home/home.component.html
```diff
@@ -24,227 +24,248 @@
...
-                            [options]="dataSourceLabels(info)"
-                            optionLabel="name"
-                            optionValue="value"
-                            formControlName="chartY2Data"
-                            styleClass="w-10rem border-100 p-dropdown--small">
-                        </p-dropdown>
+        <!-- Chart -->
+        <div class="grid-stack-item" gs-id="chart">
+            <div class="grid-stack-item-content">
+                <ng-container *ngIf="!info.power_fault">
```
mutatrum, line 203: What's the rationale of switching off the chart on power fault?

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -44,6 +46,22 @@ interface ISystemInfoError {
...
+const DASHBOARD_LAYOUT_KEY = 'DASHBOARD_LAYOUT_V1';
+const HIDDEN_WIDGETS_KEY = 'DASHBOARD_HIDDEN_WIDGETS';
+
+const WIDGET_DEFAULTS: WidgetDef[] = [
+  { id: 'hashrate',    label: 'Hashrate',            x: 0, y: 0,  w: 3,  h: 5,  minW: 2, minH: 4 },
+  { id: 'efficiency',  label: 'Efficiency',          x: 3, y: 0,  w: 3,  h: 5,  minW: 2, minH: 4 },
+  { id: 'shares',      label: 'Shares',              x: 6, y: 0,  w: 3,  h: 5,  minW: 2, minH: 4 },
+  { id: 'bestdiff',    label: 'Best Difficulty',     x: 9, y: 0,  w: 3,  h: 5,  minW: 2, minH: 4 },
+  { id: 'chart',       label: 'Chart',               x: 0, y: 5,  w: 12, h: 13, minW: 4, minH: 8 },
+  { id: 'power',       label: 'Power',               x: 0, y: 16, w: 4,  h: 7,  minW: 2, minH: 5 },
```
mutatrum: y should be 18?  (5 + 13 from the row above)

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -44,6 +46,22 @@ interface ISystemInfoError {
...
+  { id: 'heat',        label: 'Heat',                x: 4, y: 16, w: 4,  h: 7,  minW: 2, minH: 5 },
+  { id: 'fan',         label: 'Fan',                 x: 8, y: 16, w: 4,  h: 7,  minW: 2, minH: 4 },
+  { id: 'pool',        label: 'Pool',                x: 0, y: 23, w: 4,  h: 6,  minW: 2, minH: 4 },
```
mutatrum: This y should be 25 then

main/http_server/axe-os/src/app/layout/app.topbar.component.html
```diff
@@ -19,6 +19,36 @@
...
+                    </a>
+                </li>
+                <li class="hidden md:block">
+                    <a class="block py-2 text-white cursor-pointer"
+                        tooltipPosition="bottom"
+                        [pTooltip]="(dashboardEdit.editMode$ | async) ? 'Save Layout' : 'Edit Layout'"
+                        (click)="dashboardEdit.toggleEditMode()"
+                    >
+                        <i class="pi text-xl block"
+                            [ngClass]="(dashboardEdit.editMode$ | async) ? 'pi-lock' : 'pi-pencil'"></i>
```
mutatrum: Icons are confusing. I would use `pi-th-large` as the starting one instead of `pi-pencil`.

Maybe use `pi-list-check` for the widgets one, and `pi-save` for the save layout.

What's also tricky is that reset layout is on the same spot as the widgets one. Clicking the same spot restored to default, and that shouldn't happen, that should be a no-op.

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -133,15 +188,142 @@ export class HomeComponent implements OnInit, OnDestroy {
...
+
+    this.updateChartHeightForViewport();
+  }
+
+  @HostListener('window:resize')
+  onWindowResize(): void {
+    this.updateChartHeightForViewport();
+  }
+
+  private updateChartHeightForViewport(): void {
```
mutatrum: This doesn't work for me? Not sure why.

main/http_server/axe-os/src/app/components/home/home.component.scss
```diff
@@ -23,3 +19,110 @@
...
+        }
+
+        .ui-resizable-handle,
+        .gs-resizable-handle {
+            display: block !important;
+        }
+    }
+}
+
+@container widget (max-width: 300px) {
```
mutatrum: There is something weird going on the the padding of the card. On the default layout, the top row has very narrow padding, which shouldn't happen. On the default layout, padding should be 2rem, not 0.75rem, so the padding reduce kicks in too quickly, and this is noticaly because it's happening on default.

Without gridstack: 
Image

With gridstack:
Image

mutatrum:
> * Chart should adapt to height as well. Looks like the bottom of the chart is now gone?

This doesn't work for me yet.

On default layout it looks like this:

image

So, padding on the top row cards is too narrow, graph is cut off. What I also noticed is that if I only make the graph card taller, it doesn't remember this?

I also noticed that vertical and horizontal space between cards is uneven:

image

main/http_server/axe-os/src/app/components/home/home.component.html
WantClue: to display the user that something is broken, but maybe this should not be part of that pr

0xf0xx0:
just noticed latest commits make the graph not remember its size, changing it and then reloading reverts to the default

bonifacio123:
Tested latest build (adj icons, padding & margin fix) on a 601 and mobile device - LGTM

0xf0xx0:
minor nit cause idk if thisll ever be encountered in the wild, but zooming out and back in breaks the chart scaling 
image

0xf0xx0:
and now the card is at least 2x the size of the graph waa

main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -44,6 +46,22 @@ interface ISystemInfoError {
...
 const HOME_CHART_DATA_SOURCES = 'HOME_CHART_DATA_SOURCES';
+const DASHBOARD_LAYOUT_KEY = 'DASHBOARD_LAYOUT_V1';
+const HIDDEN_WIDGETS_KEY = 'DASHBOARD_HIDDEN_WIDGETS';
+
+const WIDGET_DEFAULTS: WidgetDef[] = [
+  { id: 'hashrate',    label: 'Hashrate',            x: 0, y: 0,  w: 3,  h: 5,  minW: 2, minH: 4 },
+  { id: 'efficiency',  label: 'Efficiency',          x: 3, y: 0,  w: 3,  h: 5,  minW: 2, minH: 4 },
+  { id: 'shares',      label: 'Shares',              x: 6, y: 0,  w: 3,  h: 5,  minW: 2, minH: 4 },
+  { id: 'bestdiff',    label: 'Best Difficulty',     x: 9, y: 0,  w: 3,  h: 5,  minW: 2, minH: 4 },
+  { id: 'chart',       label: 'Chart',               x: 0, y: 5,   w: 12, h: 21, minW: 4, minH: 8 },
```
0xf0xx0, line 57: default height should be 9 or 10, 9 fit best for me

Issue #1328, opened 2025-11-05
Reset error register after frequency change


mutatrum:
On some ASICs, the error register increases quite a bit whenever the frequency changes, including the initial frequency ramp-up at boot, sometimes in resulting in an error counter above 1000 immediately. Resetting the error counter after a frequency change will reduce user confusion.

PR #1623, opened 2026-03-21
Websocket api

### LINKED ISSUE #1212: Add share notification icons to AxeOS
Add the display share notification icons to the icon bar of AxeOS.

For this to work fluidly, we would need a secondary websocket stream, for data. In a future change, we can then move away from polling the info endpoint and subscribe to a info feed.

mutatrum:
This adds a new websocket endpoint: `/api/ws/live`. It streams partial updates in the format of the `/api/system/info` endpoint. The frontend is completely switched over to use this stream, so all fields are updated as soon as they are available. To improve this experience, the following things have also been changed:

 * There is only one websocket connection that's used on all screens, so no longer multiple calls to the info endpoint;
 * Share submit and the block header cards now have a blinking up and down arrow icon, similar as on the display (#1212);
 * Device disconnect is smoother now, and reconnects quicker when the device comes online again;
 * The hashrate registers now poll every second and have smooth color transitions, so they pulsate;
 * The vertical axis has been stabilised, the scale should now jump around a lot less;
 * Average efficiency is now calculated on 1m hashrate and current power, instead of front-end datapoints;
 * The websocket log handler is unified into the same code;
 * Switching axis on the chart is now almost instant, without redrawing the complete chart;
 * The chart also has a smarter way of handling historical data. The chart always starts filling with 1 second data, and slowly reduce datapoints to the statistics setting.

A recording of loading the dashboard while the device is still starting:

[Screencast from 2026-03-21 23-32-44.webm](https://github.com/user-attachments/assets/111eed6f-f454-4599-931b-6dc9deae2c34)

You can see the frequency ramp-up, settle into hashrate, receive work and submit shares.


WantClue:
@mutatrum can do a check with the cpu usage task? for comparison

0xf0xx0 (approved):
tested ack

0xf0xx0:
image
 cpu usage with #1621

mutatrum:
> @mutatrum can do a check with the cpu usage task? for comparison

These are the tasks that are involved, which is old vs new, and I only selected the tasks that are doing something with the network:

image

`tiT` is the TCP/IP task, `websocket_task` was the log websocket, which is renamed to `ws_log_task`. `ws_api_task` is the new websocket API task.

It's not a complete apples-to-apples, as IIRC in the before run the log window was open for a bit as well. But as expected, the httpd task drops to 0 as there are no more REST requests. The TCP/IP task is also lower. The WS API task is at 0.5%. Overall the CPU load on the ESP is in the same ballpark I would say.

bonifacio123:
Have tested this on an 801 and it has been running without issues. Everything seems to work as expected. The GUI data updating is much smoother.

0xf0xx0:
something i noticed that may be caused by this pr is the power draw bar will rarely drop to 0 before popping back to the correct wattage. can anyone else reproduce?

mutatrum:
> something i noticed that may be caused by this pr is the power draw bar will rarely drop to 0 before popping back to the correct wattage. can anyone else reproduce?

Power never drops to 0, there's a fixed amount added to the measurement to account for the rest of the components (display, ESP, etc). See `power_offset` in `device_config.h`.

0xf0xx0 (approved):
tested, acked, no issues on my axe

bonifacio123:
Running the latest build. Looks good. One minor annoyance is the error spikes when asics are ramping up which throws off viewing the actual smaller error values. Would it be possible to record a value of zero until asics are up to speed?

Screenshot From 2026-04-23 08-59-08

ghost:
Same.

image

mutatrum:
Related: #1328

I can take a look how to mitigate this, without going super specific. Changing frequency will trigger the same, but erasing the errors doesn't feel like the right thing to do here.

0xf0xx0:
good to merge? havent found any more issues on my end

bonifacio123:
Congrats mutatrum on this merge. I know you worked hard on it and it will make a big difference in how the UI looks and performs !!!

PR #1635, opened 2026-03-26
cutting the blocking time, http server responsive

WantClue:
Currently we have an issue with the DNS resolution that is blocking the lwIP of the Bitaxe if it tries to connect to a pool over the DNS resolution and it's not answering in an appropiate amount of time. Causing the WebUI to stale and be not responsive at all.

esp_getaddrinfo() was called once explicitly in resolve_stratum_address(), then esp_transport_connect() was called with the hostname — which triggers a second internal DNS resolution. When the pool is unreachable, each DNS query blocks for 10-20+ seconds (lwIP retries). With 3 retry attempts and double DNS per attempt, that's potentially 2+ minutes of heavy DNS activity serialized through lwIP's single tcpip_thread, which also handles the HTTP server's socket operations.          

What Changed

stratum_task.c — two locations (main connect loop + heartbeat):                                                                                                     
- Pass the already-resolved IP (conn_info.host_ip) to esp_transport_connect() instead of the hostname
- For TLS connections, set the common name (SNI hostname) separately via esp_transport_ssl_set_common_name() so certificate validation still works                  
- Added #include "esp_transport_ssl.h"                                                                                                                            

This eliminates the redundant DNS resolution, cutting the blocking time roughly in half and reducing contention on the lwIP network stack so the HTTP server stays responsive during pool failover.

### DIFF
--- a/main/tasks/stratum_task.c
+++ b/main/tasks/stratum_task.c
@@ -15,6 +15,7 @@
 #include "coinbase_decoder.h"
 #include <esp_heap_caps.h>
 #include "hashrate_monitor_task.h"
+#include "esp_transport_ssl.h"
 
 #define MAX_RETRY_ATTEMPTS 3
 #define MAX_CRITICAL_RETRY_ATTEMPTS 5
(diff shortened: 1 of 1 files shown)


mutatrum:
I can't reproduce this on #3035b7deeb895614bb558314e70355007a72141a (pre this PR).

If I use a non-existing domain name, it takes about 18 seconds from boot until fallback pool, with a non-existing IP address it takes 44 seconds. TLS on or off doesn't make a difference in timing.

Maybe I tested it with a wrong domain name?

If I test again with this PR, I get the same results, so not sure what's different now?

A compounding issue is the fact the dashboard doesn't load directly at startup, it is dependent on some information to be there. This is fixed in #1634.

It would help to have a `primary_stratum_status` field or something to show connection status at startup or when it uses the fallback pool.

PR #1637, opened 2026-03-28
Refactor theme mechanism

mutatrum:
There is a lot of duplicate code in the theme system. Refactored it to only use a single RGB value, which makes it easy to do this while reducing line count:

image


main/http_server/axe-os/src/app/components/design/theme-config.component.html
```diff
@@ -0,0 +1,43 @@
...
+        <div class="flex align-items-center">
+          <p-radioButton name="colorScheme" [value]="'light'" [(ngModel)]="selectedScheme"
+            (onClick)="changeColorScheme('light')" inputId="light"></p-radioButton>
+          <label for="light" class="ml-2">Light</label>
+        </div>
+      </div>
+    </div>
+
+    <div class="col-12 mt-4">
+      <h5>Theme Color</h5>
```
mutatrum, line 25: This file is extracted from `theme-config.component.ts‎`, only change is `Theme Colors` -> `Theme Color`

0xf0xx0 (approved):
ack, should look into switching the text color on lighter colors
image
image

mutatrum:
image

0xf0xx0 (approved):
tested ack

0xf0xx0:
two last things before merging, the buttons/checkboxes should also use contrast-color, and the hint indicators should be the fg color

image
image
image

WantClue:
just resolved the minor conflict another quick test would be ideal

mutatrum:
> just resolved the minor conflict another quick test would be ideal

Makes more sense to do again after #1651. Keep open for now, to see what the conflicts will be.

WantClue:
after 2.15

WantClue (approved):
lgtm

PR #1646, opened 2026-04-02
Add log download button with log preservation over soft reboots

mutatrum:
Adds a download button for historical logs, preserving logs over soft reboots.

Fixes #1643
Fixes #1644


STSMiner1:
Works well.

Just one thing, the device might need to be power cycled after flashing the firmware, if you don't the logs might contain corrupted data when downloaded.

Power cycle clears the logs from the device.

Example of the corrupted data in the log file.

image

mutatrum:
Hmm, that's not supposed to happen. I'll retest that.

STSMiner1:
I'm able to reproduce it (different device), here is another example of the corrupted data in the logs.

image

mutatrum:
What are the steps you take to get this reproduced?

STSMiner1:
Flash the firmware (OTA or Factory), download the logs after it's rebooted (repeat the download, it will do it eventually).

The first example above, self test manually ran once, wait for it to finish and it reboots, then access it and download the logs, corrupted data for the self test here in the log, might take a few attempts here as well.

image


Don't power cycle the devices as that will wipe the logs and you will have to start all over again to try and reproduce it.

STSMiner1:
Reproduced it on the Gamma 602 here as well.

image

Firmware - PR1646

image

mutatrum:
Ok, got it reproduced. I saw some bad data stricktly between reboots, but indeed also when running self-test. I didn't realise self-test logs were captures as well, that's a surprising benefit! The hardware reset button is a power-cycle by the way (`Reset due to power-on event`), so the log buffer is empty then.

mutatrum:
Fixed a few things:
 * Force cache flush to prevent zeroed data in the buffer
 * Add checksum to NO_INIT struct to detect write tearing
 * Use static buffer for log lines up to 2kb to reduce memory fragmentation
 * Limit scan for newline
 * Add a restart marker in the log buffer

It's still possible the log buffer is not recovered as it's a best effort solution, but it shouldn't product corrupted logs anymore.

STSMiner1:
Tested - all good now, need others to test it as well.

bonifacio123:
Tested on a 601 - LGTM

main/http_server/axe-os/src/app/components/logs/logs.component.ts
```diff
@@ -91,6 +91,30 @@ export class LogsComponent implements OnInit, OnDestroy, AfterViewChecked {
     this.logs.length = 0;
   }
 
+  public downloadLogs() {
+    const url = `http://${window.location.host}/api/system/logs`;
```
0xf0xx0: we should eventually make an axeos lib to interface with these

main/log_buffer.c
```diff
@@ -0,0 +1,229 @@
...
+
+static const char * TAG = "log_buffer";
+
+// Increment this when log_buffer_header_t struct changes
+#define LOG_BUFFER_MAGIC 0xB17A5E10 
+
+typedef struct {
+    uint32_t magic;
+    uint64_t total_written;
+    uint32_t checksum;
```
0xf0xx0, line 25: nit: 64+32+32 or 32+32+64

0xf0xx0 (approved):
ack, a nit

e: missed a nit, the download button should be disabled until the log is returned from the axe, otherwise you can spam-download and have a stream of logfiles lmao

mutatrum:
> e: missed a nit, the download button should be disabled until the log is returned from the axe, otherwise you can spam-download and have a stream of logfiles lmao

Not sure if this is possible, but I'll give it a try. It does take a short bit before it starts, especially if the buffer is full.

main/http_server/axe-os/src/app/components/logs/logs.component.ts
0xf0xx0: wait im dumb i forgot we already did this KEK, just add the log endpoint to our openapi.yaml

mutatrum: Added it to `openapi.yaml` and used as a service.

mutatrum:
> e: missed a nit, the download button should be disabled until the log is returned from the axe, otherwise you can spam-download and have a stream of logfiles lmao

Done, button is disabled during download.

PR #885, opened 2025-05-02
[WIP] Angular 19 Upgrade

eandersson:

PR #1651, opened 2026-04-08
Upgrade AxeOS to Angular 19

mutatrum:
This is an experimental migration to Angular 19. One of the biggest changes is that it's based off a default theme, instead of a copy-drop into the project. Not all elements are styled the same, so it needs polish, but this is mainly a feasibility exercise, and probably needs to be done again after #1622 and maybe other pending PRs.

The line count of this PR is massive, however, if you break it down:
- `package-lock.json`: +8958/-7,850
- theme files: 0/-8,675 (94 files deleted, 226kB)
- Actual changes: +689/-490

Rationale was that Angular 18 is no longer supported and the increasing number of security issues in the npm build part, which for a large part were coming from the Angular 18 dependencies:

Before: `60 vulnerabilities (9 low, 11 moderate, 39 high, 1 critical)`
After: `8 vulnerabilities (1 moderate, 7 high)`

Following this, the next step would be to migrate from PrimeFlex to Tailwind. But that can be done gradually.

Prior art: #885 

It would be prudent to move forward to Angular 20 or latest, as Angular 19 is also almost EOL.


WantClue:
heck of a PR, will start the review this weekend

mutatrum:
> heck of a PR, will start the review this weekend

It's smaller than it looks, see comment above. It just deletes a gazillion files and `package-lock.json` is updated.

ghost:
Network section - Missing the 'X' top right to close / cancel the Network selection box.
  

image

WantClue:
@mutatrum please resolve conflicts

WantClue (approved):
lgtm, nice migration

mutatrum:
Lots of small adjustments to the theme, should be 99% the same as with Angular 18.

ghost:
Looking good.

Just one small thing I found and that's with the Swarm page and the buttons on the right for the devices, when they are clicked on the popup messages are looking a bit odd with the  {"message":  and the 'backslash'  } as shown in the screenshot here. 

image

  
v2.14.1 for comparison below.

image

WantClue (approved):

PR #1655, opened 2026-04-10
Implement automatic hashrate anomaly detection and ASIC recovery

cniweb:
This pull request introduces an automatic hashrate anomaly detection and recovery mechanism to the `hashrate_monitor_task`. The main goal is to improve system robustness by monitoring for abnormal hashrate drops or spikes and automatically reinitializing the ASICs if anomalies persist, thus reducing the need for manual intervention or full system reboots. The detection thresholds are dynamically computed based on hardware configuration to minimize false positives.

**Hashrate anomaly detection and recovery:**

* Added a new `check_hashrate_anomaly` function that monitors for low or high hashrate anomalies and triggers ASIC live recovery if anomalies persist for a configurable number of consecutive polls. The function is documented in `hashrate_monitor_task.h` and called from the main monitoring loop before updating the highest observed hashrate. [[1]](diffhunk://#diff-f479b069b1d89257ff9436f39ee0d5c33a26e2a17983ea05e4538bbd43ce74c3R152-R206) [[2]](diffhunk://#diff-38d3bbe4d74e4b947af481760131957f2b2107ff4c6062b7d08f047e5b51cbf5R26-R38) [[3]](diffhunk://#diff-f479b069b1d89257ff9436f39ee0d5c33a26e2a17983ea05e4538bbd43ce74c3R267-R277)
* Introduced state variables and thresholds for anomaly detection, including dynamic calculation of the lower threshold based on ASIC and hash domain configuration. [[1]](diffhunk://#diff-f479b069b1d89257ff9436f39ee0d5c33a26e2a17983ea05e4538bbd43ce74c3R25-R29) [[2]](diffhunk://#diff-f479b069b1d89257ff9436f39ee0d5c33a26e2a17983ea05e4538bbd43ce74c3R39-R45) [[3]](diffhunk://#diff-f479b069b1d89257ff9436f39ee0d5c33a26e2a17983ea05e4538bbd43ce74c3R216-R227)

**Integration and robustness improvements:**

* Ensured that after ASIC recovery, measurements are reset to avoid reporting spikes from stale counters, and UART buffers are flushed to clear stale data.
* Added necessary includes for `asic_init.h` and `driver/uart.h` to support the new recovery logic.

### DIFF
--- a/main/tasks/hashrate_monitor_task.c
+++ b/main/tasks/hashrate_monitor_task.c
@@ -22,6 +22,11 @@
 #define DIV_10M (HASHRATE_1M_SIZE)
 #define DIV_1H (HASHRATE_10M_SIZE * DIV_10M)
 
+// Number of consecutive anomalous readings before triggering ASIC recovery
+#define ANOMALY_CONSECUTIVE_THRESHOLD 3
+// Stabilization delay after ASIC recovery to let tasks settle before frequency changes
+#define RECOVERY_STABILIZATION_DELAY_MS 2000
+
 static unsigned long poll_count = 0;
 static float hashrate_1m[HASHRATE_1M_SIZE];
 static float hashrate_10m_prev;
@@ -154,9 +159,15 @@ void check_hashrate_anomaly(void *pvParameters, float current_hashrate)
         return;
     }
 
-    if (current_hashrate < highest_hashrate
-        && (current_hashrate < expected_hashrate * lower_threshold_hashrate_pct
-            || current_hashrate > expected_hashrate * upper_threshold_hashrate_pct)) {
+    // Detect two types of anomaly independently:
+    // - Low hashrate: only after we've previously achieved a good hashrate (prevents
+    //   false triggers during initial ramp-up), current drops below the lower threshold
+    // - High hashrate spike: always anomalous regardless of previous highest
+    bool is_low_anomaly = current_hashrate < highest_hashrate
+                          && current_hashrate < expected_hashrate * lower_threshold_hashrate_pct;
+    bool is_high_anomaly = current_hashrate > expected_hashrate * upper_threshold_hashrate_pct;
+
+    if (is_low_anomaly || is_high_anomaly) {
         low_hashrate_count++;
         ESP_LOGW(TAG, "Abnormal hashrate detected: %.3f Gh/s (expected: %.3f Gh/s). Count: %d",
                  current_hashrate, expected_hashrate, low_hashrate_count);
@@ -165,7 +176,7 @@ void check_hashrate_anomaly(void *pvParameters, float current_hashrate)
         return;
     }
 
-    if (low_hashrate_count >= 3) {
+    if (low_hashrate_count >= ANOMALY_CONSECUTIVE_THRESHOLD) {
         reinitiate_count++;
         ESP_LOGW(TAG, "Reinitiating ASICs due to sustained abnormal hashrate. Reinitiate count: %d", reinitiate_count);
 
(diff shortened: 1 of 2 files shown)


main/tasks/hashrate_monitor_task.c
```diff
@@ -183,6 +264,17 @@ void hashrate_monitor_task(void *pvParameters)
...
+            // Check for sustained hashrate anomalies and auto-recover if needed.
+            // Must run before updating highest_hashrate so the low-anomaly guard
+            // (current < highest) correctly ignores the initial ramp-up phase.
+            check_hashrate_anomaly(pvParameters, current_hashrate);
+
+            // Track highest observed hashrate (after anomaly check)
+            if (current_hashrate > highest_hashrate) {
+                highest_hashrate = current_hashrate;
+                ESP_LOGI(TAG, "New highest hashrate: %.3f Gh/s", highest_hashrate);
+            }
```
0xf0xx0: why? if we're looking for abnormal hashrates we shouldnt be tracking the highest, and this isnt used anywhere

main/tasks/hashrate_monitor_task.c
```diff
@@ -144,6 +213,18 @@ void hashrate_monitor_task(void *pvParameters)
...
+    // of a single hash domain on a single ASIC. Multiplying by 2.0 provides a
+    // margin so that losing one domain triggers detection, while normal variance
+    // (which is less than one full domain) does not cause false positives.
+    float expected_hr = GLOBAL_STATE->POWER_MANAGEMENT_MODULE.expected_hashrate;
+    if (expected_hr > 0.0f && asic_count > 0 && hash_domains > 0) {
+        float per_domain_contribution = expected_hr / asic_count / hash_domains;
+        lower_threshold_hashrate_pct = 1.0f - (per_domain_contribution * 2.0f / expected_hr);
+        ESP_LOGI(TAG, "Hashrate anomaly lower threshold: %.0f%% of expected", lower_threshold_hashrate_pct * 100.0f);
+    }
+
```
0xf0xx0: why not a simpler 75% threshold? the more chips you add the higher the threshold becomes, the gamma turbos threshold would be 0.875 and a nerdoctaxe would be 0.992. the gammas threshold alone is 0.5, meaning this code actually needs 2 domains to go down before it triggers the reset.

main/tasks/hashrate_monitor_task.c
```diff
@@ -135,6 +149,61 @@ static void update_hashrate_averages(SystemModule * SYSTEM_MODULE)
...
+        // Perform live recovery with stabilization delay
+        uint8_t chip_count = asic_initialize(GLOBAL_STATE, ASIC_INIT_RECOVERY, RECOVERY_STABILIZATION_DELAY_MS);
+
+        if (chip_count > 0) {
+            ESP_LOGI(TAG, "ASIC recovery successful, resuming normal operation.");
+            // Reset measurements to avoid hashrate spike from stale counters
+            hashrate_monitor_reset_measurements(GLOBAL_STATE);
+        } else {
+            ESP_LOGE(TAG, "ASIC recovery failed - chip count 0");
```
0xf0xx0: this should be a full reboot, instead of leaving the user in a worse state than before

main/tasks/hashrate_monitor_task.h
```diff
@@ -23,4 +23,17 @@ void hashrate_monitor_reset_measurements(void *pvParameters);
...
+
+/**
+ * Check for sustained hashrate anomalies and auto-recover ASICs if needed.
+ *
+ * When the hashrate drops below a dynamic lower threshold or exceeds an upper
+ * threshold for multiple consecutive polls, this function reinitializes the
+ * ASICs using live recovery mode to restore normal operation without a full
+ * system reboot.
+ *
+ * Inspired by TinyChipHub ESP-Miner-TCH stability improvements.
```
0xf0xx0: what improvements? skimmed the tch codebase and i cant find similar functions.

0xf0xx0 (changes requested):
claude :/

main/tasks/hashrate_monitor_task.h
0xf0xx0: nvm its on the `master-TCH` branch, i see now

mutatrum:
This is overly complex. All cases of diminished hashrate are happening if one or more domains of an ASIC are dropping out, e.g. the hashcounter register is not changing anymore. This can be detected without any heuristics.

The problem however is that this is almost always a sign of a crappy power supply. If the device is powered by a decent power supply, this never happens. So just simply restarting the ASIC hides this problem. Maybe we should add a notification messages first, when one or more of the domains dropped out, so the user can try to either adjust frequency or voltage to remedy the issue.

PR #1659, opened 2026-04-12
resize for mobile view

WantClue:
resizes the chart view for mobile devices


main/http_server/axe-os/src/app/components/home/home.component.ts
```diff
@@ -210,6 +210,24 @@ export class HomeComponent implements OnInit, OnDestroy {
     clearTimeout(this.resizeTimer);
     this.resizeTimer = setTimeout(() => {
       this.grid?.cellHeight(this.getCellHeight());
+
+      if (this.grid) {
+        const isMobile = window.innerWidth < 768;
```
mutatrum: `window.matchMedia("(max-width: 768px)").matches;`

This should handle rotated devices better.

main/http_server/axe-os/src/app/components/home/home.component.ts
mutatrum: This is trouble as well. Can you try with cellheight `auto` so gridstack tries to get square cell sizes?

PR #1660, opened 2026-04-12
Add dynamic ticks on horizontal chart axis

mutatrum:
Also does a smarter way of limiting the data points in the chart.


0xf0xx0 (approved):
tested ack

PR #1663, opened 2026-04-15
Add wide screen toggle

mutatrum:
Add a toggle to the customize dashboard to switch to wide screen:

image

On a 4k monitor, goes from this:

image

To this:

image

This works on all screens. As it duplicates what was already available on the Logs screen, the toggle is removed there.


WantClue (approved):
code lgtm

EricDimitri:
Hi!! I just updated to the 0b2 version, 2.14.0b2, and I dont see it as you show, different to both windows. It looks like this:
image

WHy is that?

EricDimitri:
Oh!!! I Just realized, that you can customize the Dashboard!! Thats really good!!!!
Thanks!!!

PR #1668, opened 2026-04-17
Add API and UI to reset best difficulty

trcjr:
- Added /api/system/resetBestDifficulty endpoint to backend (C, OpenAPI)
- Integrated reset button in web UI (Angular)
- Updated SystemApiService and EditComponent for new feature
- Ensures best difficulty can be reset from UI and API

This PR implements a secure, CORS-compliant API for resetting the best difficulty, updates the OpenAPI spec, and provides a user-facing button in the web interface. All changes are backward compatible.


trcjr:
This was originally develop before the scoreboard. After playing with it rebased on the scoreboard for a bit I don't feel this is needed.

I'm considering a `/api/system/resetScoreboard` instead.

PR #1670, opened 2026-04-23
Rename API key overheat_mode to overheatMode (#621)

### LINKED ISSUE #621: bug: API formatting inconsistency

0xdeadbeefnetwork:
Closes #621.

The `overheat_mode` JSON key in the system API is the only snake_case key in an otherwise camelCase REST surface. This renames it to `overheatMode`.

## Compatibility

The reporter asked about deprecation procedure, so this PR keeps the old spelling working alongside the new one:

- **GET `/api/system/info`** emits both `overheatMode` (canonical) and `overheat_mode` (deprecated alias).
- **PATCH `/api/system`** accepts either spelling. If both are present, `overheatMode` wins.
- **OpenAPI** marks `overheat_mode` with `deprecated: true` in both schemas.
- **Frontend** reads `info.overheatMode ?? info.overheat_mode` so AxeOS keeps working if it ever talks to older firmware (and vice-versa).

The NVS storage key (`nvs_key_name`) is unchanged so existing devices keep their stored setting on upgrade.

## Test plan

- [x] `npm run generate:api && ng build --configuration=production` passes.
- [ ] CI: ESP-IDF build (no local toolchain).
- [ ] Manual: PATCH with `overheat_mode: 0` still clears the overheat flag.
- [ ] Manual: PATCH with `overheatMode: 0` clears the overheat flag.
- [ ] Manual: GET response contains both keys with the same value.

### DIFF
--- a/main/http_server/axe-os/src/app/components/edit/edit.component.html
+++ b/main/http_server/axe-os/src/app/components/edit/edit.component.html
@@ -46,7 +46,7 @@
             </ng-container>
         </div>
 
-        <div *ngIf="form.get('overheat_mode')?.value === 1" class="mt-4">
+        <div *ngIf="form.get('overheatMode')?.value === 1" class="mt-4">
             <button pButton type="button" label="Disable Overheat Mode"
                     class="p-button-danger w-full"
                     (click)="disableOverheatMode()">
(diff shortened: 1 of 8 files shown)


0xf0xx0:
nack, llm slop

see #945 if you'd like to contribute without relying entirely on claude

PR #1671, opened 2026-04-23
Only show restart warning when Wi-Fi changes (#581)

### LINKED ISSUE #581: AxeOS 2.4.1 incorrectly states restart is needed to change hostname

0xdeadbeefnetwork:
Closes #581.

The network settings page warned "you must restart" after any save, but the device applies hostname changes live (visible immediately in the swarm view, as the issue reporter noted). Limit the warning to dirty `ssid`/`wifiPass` so it only appears when a Wi-Fi reconnection is actually needed.

## Test plan

- [x] Frontend builds (`ng build --configuration=production`).
- [ ] Manual: change only hostname → no restart warning, hostname updates immediately.
- [ ] Manual: change SSID → restart warning appears.
- [ ] Manual: change Wi-Fi password → restart warning appears.
- [ ] Manual: change hostname + SSID together → restart warning appears.

### DIFF
--- a/main/http_server/axe-os/src/app/components/network-edit/network.edit.component.ts
+++ b/main/http_server/axe-os/src/app/components/network-edit/network.edit.component.ts
@@ -64,11 +64,16 @@ export class NetworkEditComponent implements OnInit {
       form.ssid = form.ssid.trim();
     }
 
+    // Hostname is applied live; only Wi-Fi changes require a restart.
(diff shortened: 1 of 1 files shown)


mutatrum:
This might change with #1131, might make sense to do this after that one.

PR #1674, opened 2026-04-23
Show actual resolved frequency

mutatrum:
Expose frequency as resolved by the PLL solver.

Example, when setting 528 Mhz:

image

The PLL cannot perfectly resolve this, so it lands to the closest frequency:

```
I (992164) power_management: New ASIC frequency requested: 528 MHz (current: 525 MHz)
I (992165) pll: Frequency: 528.125 MHz (fb_divider: 169, refdiv: 2, postdiv1: 4, postdiv2: 1)
```

Which will then be shown:

image


WantClue (approved):
thanks

Issue #1633, opened 2026-03-26
Prevent accidental reset through web interface - Feature impovement


Nexus9090:
On a number of occasions I've accidentally reset the device through the AXE-OS page, the control panel icon is placed quite close to the reset button and when using a tablet to access the web interface it is too easy to accidentally hit the reset icon which takes place immediately and without confirmation.

I propose that the reset icon include a pop-up to confirm the reset procedure to avoid inadvertent resets of the device.

Thanks

PR #1675, opened 2026-04-23
Confirm before restart from topbar (#1633)

0xdeadbeefnetwork:
## Summary
- Prompt the user via `window.confirm()` before calling `systemService.restart()` from the topbar icon.
- Addresses #1633: on tablets the restart icon sits close to the mining controls and is easy to hit by accident, rebooting the device and losing in-flight work with no warning.

## Test plan
- [x] `ng build --configuration development` — clean build, no TS errors
- [x] `ng serve` started, confirm-dialog string present in bundled `main.js`
- [ ] Click restart icon → prompt appears; Cancel aborts, OK still reboots
- [ ] No regression on the adjacent Pause/Resume icon (untouched)

### DIFF
--- a/main/http_server/axe-os/src/app/layout/app.topbar.component.ts
+++ b/main/http_server/axe-os/src/app/layout/app.topbar.component.ts
@@ -72,6 +72,9 @@ export class AppTopBarComponent implements OnInit, OnDestroy {
   }
 
   public restart() {
+    if (!window.confirm('Restart the device? Current mining work will be lost.')) {
+      return;
+    }
     this.systemService.restart().subscribe({
       next: () => this.toastr.success('Device restarted'),
       error: () => this.toastr.error('Restart failed')


main/http_server/axe-os/src/app/layout/app.topbar.component.ts
```diff
@@ -72,6 +72,9 @@ export class AppTopBarComponent implements OnInit, OnDestroy {
   }
 
   public restart() {
+    if (!window.confirm('Restart the device? Current mining work will be lost.')) {
```
mutatrum, line 75: Not sure what will be lost on restart, just asking `Restart the device?` should be sufficient. And please add a screenshot on how this looks on the PR.

WantClue (commented):
NACK there is no benefit in asking the user this question

PR #1676, opened 2026-04-23
Chart ASIC temp as hottest chip (#1476)

0xdeadbeefnetwork:
## Summary
- The ASIC Temp chart on multi-chip units was plotting only ASIC #1's temperature (`info.temp`), hiding hotter chips. Issue reporter on a GammaTurbo (two-chip 800 GT) saw ASIC #2 running hotter than charted.
- Use `Math.max(info.temp, info.temp2)` so the chart reflects the hottest chip.
- On single-chip boards `Thermal_get_chip_temp2` returns -1, so `Math.max` degenerates to `info.temp` — no change in behavior.

## Test plan
- [x] `ng build --configuration development` — clean build, no TS errors
- [ ] On a multi-chip unit (GammaTurbo / GammaDuo / Hex / SupraHex), verify chart tracks the hotter of the two reported chip temps
- [ ] On a single-chip unit (Ultra / Supra / Gamma), verify chart still shows ASIC temp as before

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.ts
+++ b/main/http_server/axe-os/src/app/components/home/home.component.ts
@@ -1153,7 +1153,7 @@ export class HomeComponent implements OnInit, OnDestroy {
       case eChartLabel.hashrate_10m:       return info.hashRate_10m;
       case eChartLabel.hashrate_1h:        return info.hashRate_1h;
       case eChartLabel.errorPercentage:    return info.errorPercentage;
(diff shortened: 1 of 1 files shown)


mutatrum:
Nack, it makes more sense to add ASIC 2 temp as a separate line. I've added that to #1677. I'm also working on showing multiple lines of the same sensor unit, so it's possible to show both asic and vr temperature on the same chart.

PR #1687, opened 2026-05-11
feat: move uptime to misc card

0xf0xx0:
removes the share luck % and moves the uptime from the diff card to its own

image
image


0xf0xx0:
no longer removes the share luck %, tho im still leaning towards it being kinda useless... thoughts?

also im not sure why it takes a couple toggles for the card to render properly, furst its invisible, then a couple toggles later it decides to render :\

WantClue:
> no longer removes the share luck %, tho im still leaning towards it being kinda useless... thoughts?
> 
> 
> 
> also im not sure why it takes a couple toggles for the card to render properly, furst its invisible, then a couple toggles later it decides to render :\

I tend to agree with you, it gives little information at all

WantClue (approved):

Issue #1584, opened 2026-03-01
Migrate notification messages to backend


mutatrum:
Currently the logic for dashboard notifications is done in the front-end. The system would be cleaner if this logic is done in the backend, and the messages are just exposed as a list with type, text, severity, dismissible, timestamp.

F.e. this would allow way easier exposing of #1565 to the front-end, and support a general version of #1465.

PR #1689, opened 2026-05-13
Pool disconnect

### LINKED ISSUE #591: Axe not realizing stratum host went down and loops old work
**Describe the bug**
If the pool goes down the axe may not notice if it hasn't found any shares to submit in some amount of time. Unclear what that is.

**To Reproduce**
1. Set high difficulty with DATUM or similar
2. Go for extended period without shares (give it an hour to be on the safe side)
3. Kill DATUM
4. Axe will continue mining without having anywhere to submit shares and without getting fresh jobs

**Expected behavior**
After 30 seconds or so the Axe should realize it's not doing anything useful any more and fail over to pool #2

**Screenshots & Photos**

 - Bitaxe HW version: The 650GH one, firmware v2.4.1
 - Bitaxe HW vendor: Booth at Nashville conf 2024
 - ESP-Miner FW version: [e.g. 2.1.1, etc]
 - Hash Frequency:
 - Voltage:
 - Pool URL, Port, User:

WantClue:
on pool failure bitaxe currently keeps the job and continues mining.
In order to safe on power consumption on a catastrophic failure of the pool connection or any downtime we make use of the newly added pause mining feature


main/tasks/fan_controller_task.c
```diff
@@ -72,6 +72,8 @@ void FAN_CONTROLLER_task(void * pvParameters)
             update_fan_speed(GLOBAL_STATE, 100.0f, "Overheat");
         } else if (GLOBAL_STATE->SYSTEM_MODULE.mining_paused) {
             update_fan_speed(GLOBAL_STATE, 30.0f, "Paused");
+        } else if (GLOBAL_STATE->SYSTEM_MODULE.pools_unavailable) {
```
0xf0xx0, line 75: something that could be resolved in a different pr is these hardcoded fan percentages  
i dont see much reason to not leave the auto fan speed to do its thing

main/tasks/power_management_task.c
```diff
@@ -150,11 +150,12 @@ void POWER_MANAGEMENT_task(void * pvParameters)
         power_management->chip_temp2_avg = Thermal_get_chip_temp2(GLOBAL_STATE);
 
         power_management->vr_temp = Power_get_vreg_temp(GLOBAL_STATE);
-        // User requested pause or hardware fault
-        if ((sys_module->mining_paused || sys_module->hardware_fault) && !is_paused) {
+        // User pause, hardware fault, or all pools unreachable
+        bool wants_stop = sys_module->mining_paused || sys_module->hardware_fault || sys_module->pools_unavailable;
```
0xf0xx0, line 146: this feels weird but i dont have suggestions atm

0xf0xx0 (commented):
will test soon

0xf0xx0:
closes: #591

main/tasks/power_management_task.c
mutatrum: Might be good to cover this in #1584

WantClue: agreed this needs to be tackled in another pr to move the backend into a better place

0xf0xx0:
tested ack, went down and came back up just fine
image

WantClue:
- Primary dies → V1/V2 task exits → coordinator counts 1 failure, swaps to fallback.
  - Fallback also dies → counter hits 2 → coordinator enters PAUSED, sets pools_unavailable = true. Power management cuts ASIC power, fan controller responds.
  - While paused, coordinator probes both pools every 30s. First to respond wins; mining resumes on that pool, counter resets, pools_unavailable clears.
  - Single-pool case still works: threshold = 1, single failure → pause → recovery probe retries forever.

WantClue:
Screenshot 2026-05-20 at 20 06 07
beautiful

WantClue:
What was broken
  
  After exhausting MAX_RETRY_ATTEMPTS, the V1 stratum task was incrementing a local consecutive_pool_failures counter, then exiting via protocol_coordinator_notify_failure(). The coordinator would spawn a fresh V1 task for the fallback pool — but the fresh task started with consecutive_pool_failures = 0 again. The "all pools unreachable" trigger (counter ≥ 2 when a fallback is configured) could therefore never fire, so pools_unavailable was never set and
   the miner never paused — it just looped between primary and fallback forever.

  Two related issues:
  - stratum_primary_heartbeat (the old recovery probe loop) survived in stratum_v1_task.c but was never spawned, and it still referenced the renamed stratum_close_connection() — a latent link error.
  - The coordinator's existing heartbeat only probed primary while running on fallback (to switch back). There was no path to recover from pools_unavailable once set.

  The fix

  Moved the "all pools down" decision and recovery into protocol_coordinator.c, which is the only component that sees both V1 and V2 task lifecycles:

  - Added s_consecutive_pool_failures (persists across task restarts) and a COORD_STATE_PAUSED state.
  - Added protocol_coordinator_notify_success(); V1 and V2 tasks call it on successful setup to reset the counter and clear pools_unavailable.
  - On PROTOCOL_FAILED: counter is incremented; threshold is 2 with a fallback configured, 1 otherwise. Hitting the threshold sets pools_unavailable = true (power management cuts ASIC power) and transitions to PAUSED.
  - While paused, the coordinator probes both pools every 30s (user-preferred pool first) and resumes mining on the first reachable one.
  - Generalized the existing heartbeat_probe_v1/sv2 helpers into probe_pool_v1/sv2(url, port, …) so they can probe either pool.
  - Stripped the now-redundant local counter / pools_unavailable toggling and the orphaned stratum_primary_heartbeat/stratum_probe_pool code from stratum_v1_task.c.

  Files changed: protocol_coordinator.{h,c}, stratum_v1_task.c, stratum_v2_task.c

PR #1699, opened 2026-05-20
Connect npm dev server to real device

mutatrum:
This makes it possible to run a local dev environment against a real device. It also supports hot module replacement, but that's limited in Angular 18.

Run with:
`/main/http_server/axe-os$ BITAXE_IP=x.x.x.x npm run start:proxy`

It starts a locally hosted front-end on `localhost:4200` which connects to an actual device on the specified IP address:

image


WantClue (approved):
nice lgtm awesome to test like this

PR #1701, opened 2026-05-20
Add miningcore reject reason parsing

mutatrum:
Adds parsing and unit test for miningcore reject reason format.


WantClue (approved):
lgtm

PR #1705, opened 2026-05-25
Clarify partial mining reward dashboard message

0xjc65eth:
## Summary

Fixes the most visible part of #1703 by changing the dashboard message for decoded partial mining rewards from an alarming warning to a neutral informational message.

Before:

- Severity: warn
- Text: `Your share of the mining reward is only X%`

After:

- Severity: info
- Text: `Mining reward is split; your decoded share is X%`

The existing zero-share warning is unchanged:

- `You don't have a share in the mining reward`

## Why

For Sv2 Extended Channel setups such as the SRI pool donation mode described in #1703, a partial decoded payout can be intentional and valid. The previous wording made expected reward splitting sound suspicious even when AxeOS successfully decoded that the user has a non-zero share.

This keeps the useful visibility while avoiding an accusatory warning for legitimate split-reward configurations.

## Scope

This is intentionally small and UI-only:

- No firmware/backend changes.
- No new settings.
- No pool-specific allowlist.
- No change to the zero-share warning.

There is also a broader existing PR (#1678) that adds a toggle and larger coinbase-output handling. This PR is meant as a minimal alternative/foundation that can be reviewed independently.

## Testing

- Ran `npm ci` in `main/http_server/axe-os`.
- Ran `npm run build` successfully.

Build completed with the existing Angular initial bundle budget warning, but exited successfully.

PR #1706, opened 2026-05-25
Reset OTA update modal after errors

### LINKED ISSUE #764: OTA update screen doesn't clear if an error occurs
**Describe the bug**
If an error occurs when sending a firmware file over the air, the screen will get stuck on the error screen and require a reset to clear, even though the chip is still hashing.

**To Reproduce**
1. Get a firmware file
2. Send a malformed OTA update request (notice the lack of POST): `curl --data-raw "@esp-miner.bin" http://bitaxe/api/system/OTA`
3. Screen is stuck on
```
Firmware update
esp-miner.bin
Write Error
```

**Expected behavior**
I expected the boot button to clear the error screen.

 - Bitaxe HW version: gamma (gekkoscience mod)
 - Bitaxe HW vendor: gekkoscience
 - ESP-Miner FW version: master 36272f63
 - Hash Frequency: irrelevant
 - Voltage: irrelevant
 - Pool URL, Port, User: irrelevant

**Additional information**
while the reproduction steps involve sending a malformed request, the same issue happens if the connection drops out mid-transfer.

0xjc65eth:
## Summary
- reset OTA upload progress when an update starts, fails, or the modal is dismissed
- show the existing Applying update state after upload completion while waiting for the response
- route Firmware and AxeOS error handling through a shared helper so non-fatal failures can be cleared by the user

## Why
When an OTA upload hits a non-fatal error, the update modal can remain in a stale progress/error state because the observable error path does not run the complete handler. This makes recovery confusing and matches the user-facing behavior reported in #764.

Fixes #764.

## Testing
- npm run build

### DIFF
--- a/main/http_server/axe-os/src/app/components/update/update.component.html
+++ b/main/http_server/axe-os/src/app/components/update/update.component.html
@@ -102,7 +102,7 @@
             <i class="pi pi-check-circle text-green-500" style="font-size: 3.5rem; margin-bottom: 1rem;"></i>
             <div class="text-lg text-center">{{ updateMessage }}</div>
             <div class="mt-4" *ngIf="updateTarget !== 'AxeOS'">
-                <p-button (onClick)="progressModal.isVisible = false">Close</p-button>
+                <p-button (onClick)="dismissProgressModal()">Close</p-button>
             </div>
         </ng-container>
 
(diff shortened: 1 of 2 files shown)


0xjc65eth:

PR #1707, opened 2026-05-25
Do not count optional Stratum setup rejections as shares

### LINKED ISSUE #1383: Filter out `Method not found` on failed setup messages as rejected share reasons
Some pools (f.e. Ocean) don't accept `mining.suggest_difficulty` and `mining.extranonce.subscribe`:
```
₿ (25228) system: Syncing clock
₿ (25231) create_jobs_task: New Work Dequeued 692374e64ec09303
₿ (25231) stratum_task: Block height 924898
₿ (25237) create_jobs_task: New pool difficulty 131072
₿ (25242) stratum_task: Scriptsig: .< OCEAN.XYZ >.Sinnx3 Mining.....@..V..A
₿ (25248) create_jobs_task: Set chip version rolls 65535
₿ (25256) stratum_api: rx: {"error":null,"id":3,"result":true}
₿ (25268) stratum_task: setup message accepted
₿ (25273) stratum_api: tx: {"id": 4, "method": "mining.suggest_difficulty", "params": [4657]}
₿ (25283) stratum_api: tx: {"id": 5, "method": "mining.extranonce.subscribe", "params": []}
₿ (25360) stratum_task: Stratum response time: 77.17 ms
₿ (25362) stratum_api: rx: {"error":[-3,"Method not found",null],"id":4,"result":null}
₿ (25365) stratum_task: setup message rejected: Method not found
₿ (25439) stratum_api: rx: {"error":[-3,"Method not found",null],"id":5,"result":null}
₿ (25440) stratum_task: message result rejected: Method not found
```

However, these are registered as failed shares with `Method not found` as reject reason, which is incorrect. These can be safely ignored.

0xjc65eth:
## Summary
- track the request IDs used for optional Stratum V1 setup calls after authorization
- handle mining.suggest_difficulty and mining.extranonce.subscribe responses separately from share submit responses
- prevent optional setup rejections such as Method not found from being reported as rejected shares

## Why
Some pools do not implement optional Stratum V1 methods and can return Method not found. Those responses are setup capability rejections, not share submit failures, so they should not pollute rejected share stats or dashboard rejection reasons.

Fixes #1383.

## Testing
- git diff --check

Full firmware build was not run locally because idf.py is not installed in this shell and the Docker daemon is not running (/var/run/docker.sock is unavailable).

### DIFF
--- a/main/tasks/stratum_v1_task.c
+++ b/main/tasks/stratum_v1_task.c
@@ -56,6 +56,11 @@ static int stratum_get_next_uid(GlobalState * GLOBAL_STATE)
     return uid;
 }
 
+static bool stratum_v1_is_optional_setup_response(int message_id, int suggest_difficulty_message_id, int extranonce_subscribe_message_id)
+{
+    return message_id == suggest_difficulty_message_id || message_id == extranonce_subscribe_message_id;
+}
+
 struct timeval tcp_snd_timeout = {
     .tv_sec = 5,
     .tv_usec = 0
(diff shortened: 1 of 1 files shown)


0xjc65eth:

PR #1708, opened 2026-05-25
Warn when ASIC hash domains stop incrementing

### LINKED ISSUE #1053: "Flatline of Death" - hashrate stops changing, no new shares are found. Pool is still alive and sending work

0xjc65eth:
## Summary
- detect stalled ASIC hash domains from the existing per-domain hash counter registers
- expose stalled domain flags through the system info API and OpenAPI schema
- show an AxeOS warning and highlight stalled domains in the hashrate register card

## Why
Issue #1053 reports devices that continue receiving pool work while one or more ASIC domains stop producing useful work. A previous recovery-only approach was rejected because it hid likely power/voltage/frequency problems behind automatic restarts. This PR adds direct observability first: the firmware reports the stalled domain so users can diagnose power supply, voltage, and frequency settings instead of only seeing a flatline.

Fixes #1053.

## Testing
- npm run build
- git diff --check

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.html
+++ b/main/http_server/axe-os/src/app/components/home/home.component.html
@@ -583,9 +583,11 @@
                                     <td *ngIf="asicAmount > 1" class="text-500 text-left">
                                         {{ i + 1 }}
                                     </td>
-                                <td *ngFor="let domain of asic?.domains; trackBy: trackByIndex"
-                                    [style.backgroundColor]="getHeatmapColor(info, domain)"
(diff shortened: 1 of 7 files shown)


0xjc65eth:

PR #1709, opened 2026-05-25
Require physical button authorization for OTA updates

### LINKED ISSUE #1220: unlock OTA
require button press to unlock OTA

0xjc65eth:
## Summary
- require a recent BOOT button press before accepting firmware or AxeOS OTA uploads
- add a 5-minute physical authorization window shared by both OTA endpoints
- add an NVS-backed otaAuthRequired setting, enabled by default
- require recent physical authorization before disabling the OTA button requirement
- expose OTA authorization state in system info and show it in AxeOS update/settings screens

## Why
OTA updates are powerful because they can replace device firmware. This adds a physical-presence check without forcing users to remember another password or making a default password that provides little protection. Users who deliberately trust their network can disable the requirement, but that action also requires a recent button press.

Fixes #1220.

## Testing
- npm run build
- git diff --check


0xjc65eth:

mutatrum:
Having an open REST API endpoint that can disable the OTA button authorisation defeats the purpose. This needs more though.

PR #1710, opened 2026-05-25
Report found blocks only after accepted shares

### LINKED ISSUE #1552: fw should only report a found block if share was submitted and accepted
**Describe the bug**
the firmware seems to treat valid shares as found blocks, even if the share was never submitted to the pool.

**To Reproduce**
1. Set up a regtest node
2. Set up a pool pointing to it with a high difficulty (10k is plenty, 25k at most)
3. mine on it with your axe
4. watch the "block found" banner pop up even though no shares were submitted to the pool

**Expected behavior**
I expected the block found banner to appear only with submitted shares.

0xjc65eth:
## Summary
- keep updating best/session difficulty as soon as a nonce is found
- mark block candidates as pending only after the share is successfully submitted
- increment blockFound and show the block banner only when the pool accepts a pending block candidate
- clear pending block candidates when the pool rejects the matching pending share path

## Why
A nonce can meet network difficulty but still fail to become a valid pool-accepted block candidate if it is never submitted or is rejected. The dashboard should not report a found block until the pool has accepted the submitted share.

Fixes #1552.

## Testing
- git diff --check

### DIFF
--- a/main/global_state.h
+++ b/main/global_state.h
@@ -61,6 +61,7 @@ typedef struct
     uint64_t best_session_nonce_diff;
     char best_session_diff_string[DIFF_STRING_SIZE];
     int block_found;
+    uint8_t pending_block_candidates;
     bool show_new_block;
     char * ssid;
     char wifi_status[256];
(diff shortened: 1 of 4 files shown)


0xjc65eth:

PR #1711, opened 2026-05-25
Keep overheat warning active during self protect

0xjc65eth:
## Summary
- update the in-memory overheat state immediately when self protect mode starts
- clear the in-memory overheat state only after recovery successfully restarts mining

## Why
The dashboard and display read the in-memory `SYSTEM_MODULE.overheat_mode` value. During self protect recovery, the firmware persisted the NVS flag immediately but did not update the runtime state until later in the power-management loop, so the warning could disappear while the device was still cooling/recovering.

Fixes #1638.

## Testing
- git diff --check


0xjc65eth:

PR #1712, opened 2026-05-25
Show Stratum auth failures in pool status

### LINKED ISSUE #1358: show pool stratum auth error on AxeOS dashboard
If esp-miner fails to authenticate via stratum with the pool, we should propagate this error to the AxeOS dashboard.

example, if I use an invalid mining address with public-pool;

`{"id": 3, "method": "mining.authorize", "params": ["bc1qnp980s5fpp8l94p5cvttntdqy8rvrq74qly2yrfmzkdsntqzlc5qkc4rkq.bitaxe", "x"]}`

it will return;

```
I (15123) stratum_api: rx: {"id":3,"result":null,"error":[20,"Authorization validation error",", bc1qnp980s5fpp8l94p5cvttntdqy8rvrq74qly2yrfmzkdsntqzlc5qkc4rkq"]}
E (15138) stratum_task: setup message rejected: Authorization validation error
```
 
Currently AxeOS does not show the "Authorization validation error" to the user.

0xjc65eth:
## Summary
- surface failed `mining.authorize` setup responses in the existing pool connection status field
- keep the message short enough for the current `pool_connection_info` buffer

## Why
When a pool rejects `mining.authorize`, for example because the configured wallet address is invalid, AxeOS currently does not show the pool error to the user. The firmware logs the rejection, but the dashboard can still leave users guessing why mining never starts.

Fixes #1358.

## Testing
- git diff --check

### DIFF
--- a/main/tasks/stratum_v1_task.c
+++ b/main/tasks/stratum_v1_task.c
@@ -604,6 +604,11 @@ void stratum_v1_task(void *pvParameters)
                     }
                 } else {
                     ESP_LOGE(TAG, "setup message rejected: %s", stratum_api_v1_message.error_str);
+                    if (stratum_api_v1_message.message_id == authorize_message_id) {
+                        snprintf(GLOBAL_STATE->SYSTEM_MODULE.pool_connection_info,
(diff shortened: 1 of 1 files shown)


0xjc65eth:

PR #1713, opened 2026-05-25
Clamp Stratum version rolling mask to ASIC capability

### LINKED ISSUE #1169: hardcoded version-rolling.mask might exceed what ASIC can roll
hardcoded version-rolling.mask -> 0xFFFFFFFF in https://github.com/bitaxeorg/ESP-Miner/blob/0958185217f324c8fe786cd6ef05f22e117f616d/components/stratum/stratum_api.c#L424-L434

This might not agree with what the ASIC can actually roll. This parameter should be pulled from an ASIC-specific function in components/asic/ so we don't suggest to the poll something the hardware can't do.

0xjc65eth:
## Summary
- advertise the ASIC-supported version rolling mask instead of the hardcoded `ffffffff` Stratum V1 mask
- add an ASIC helper for the supported version rolling mask
- clamp pool-provided `mining.set_version_mask` / `mining.configure` results before storing them globally

Fixes #1169

## Testing
- `git diff --check`

Firmware build was not run because `idf.py` is not installed in this environment.

### DIFF
--- a/components/asic/asic.c
+++ b/components/asic/asic.c
@@ -10,6 +10,7 @@
 #include "asic.h"
 #include "device_config.h"
 #include "frequency_transition_bmXX.h"
+#include "utils.h"
 
 static const char *TAG = "asic";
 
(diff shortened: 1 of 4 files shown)


0xjc65eth:

PR #1714, opened 2026-05-25
Reset ASIC error count after frequency changes

### LINKED ISSUE #1328: Reset error register after frequency change

0xjc65eth:
## Summary
- reset each supported ASIC error-count register after a frequency transition completes
- clear the in-memory error measurement at the same time to avoid a stale baseline or wraparound spike
- route the reset through the common ASIC facade so UI, BAP, startup ramps, and safety-mode frequency changes get the same behavior

Fixes #1328

## Testing
- `git diff --check`

Firmware build was not run because `idf.py` is not installed in this environment.


0xjc65eth:

PR #1715, opened 2026-05-25
Chart the maximum ASIC temperature by default

### LINKED ISSUE #1476: ASIC Temp chart misleading on multi chip units

0xjc65eth:
## Summary
- make the default `asicTemp` chart source use the hottest ASIC temperature when a second sensor is present
- store historical `asicTemp` samples as the max of ASIC temp 1 and 2
- rename the chart label to `Max ASIC Temp` while keeping the existing `asicTemp` key compatible

Fixes #1476

## Testing
- `git diff --check`
- `npm run build`


0xjc65eth:

PR #1716, opened 2026-05-25
Avoid restart warning for hostname-only network changes

### LINKED ISSUE #581: AxeOS 2.4.1 incorrectly states restart is needed to change hostname

0xjc65eth:
## Summary
- only show the network restart warning when Wi-Fi settings changed
- keep hostname-only saves quiet because the hostname is reflected without a restart
- only enable the Restart button after saves that actually need a restart

Fixes #581

## Testing
- `git diff --check`
- `npm run build`


0xjc65eth:

PR #1718, opened 2026-05-25
Add camelCase overheat mode API field

### LINKED ISSUE #621: bug: API formatting inconsistency

0xjc65eth:
## Summary
- add `overheatMode` to the system info response
- keep `overheat_mode` as a deprecated compatibility alias
- update OpenAPI and the frontend mock data so generated API types continue to build

Fixes #621

## Testing
- `git diff --check`
- `npm run build`


0xjc65eth:

PR #1719, opened 2026-05-26
Add local work generation via Bitcoin RPC

keegreil:
This PR provides local work generation using getblocktemplate RPC communication from a local Bitcoin node, removing the need for a separate Stratum server.  It include using submitblock to submit any found blocks back to the connected node.

This is experimental, and may need to be marked as such before wide use, since this work generation code is not nearly as battle tested as the major Stratum server implementations (CKPool, DATUM, Public Pool, etc.).  

Submitting here to provide for discussion and evaluation.


components/stratum/local_work_builder.c
```diff
@@ -0,0 +1,1789 @@
...
+#include "esp_heap_caps.h"
+#include "esp_psram.h"
+#endif
+#include "libbase58.h"
+#include "mbedtls/sha256.h"
+#include "segwit_addr.h"
+#include "utils.h"
+
+#define LOCAL_WORK_COINBASE_VERSION 2
+#define LOCAL_WORK_SEQUENCE_FINAL 0xffffffffU
```
0xf0xx0, line 22: should we care about bip-54 here? it recently got advanced to complete, though theres a bit of a discussion going on...

components/stratum/local_work_builder.c
```diff
@@ -0,0 +1,1789 @@
...
+    char *end = NULL;
+    unsigned long value = strtoul(hex, &end, 16);
+    if (!end || *end != '\0' || value > UINT32_MAX) {
+        return ESP_ERR_INVALID_ARG;
+    }
+    *out = (uint32_t)value;
+    return ESP_OK;
+}
+
+static const char *find_json_value(const char *json, const char *key)
```
0xf0xx0, line 257: why not cJSON?

main/local_work_client.c
```diff
@@ -0,0 +1,823 @@
...
+
+#define LOCAL_WORK_TEMPLATE_INITIAL_DELAY_MS 15000
+#define LOCAL_WORK_TEMPLATE_ERROR_INTERVAL_MS 30000
+#define LOCAL_WORK_TEMPLATE_UNCONFIGURED_INTERVAL_MS 30000
+#define LOCAL_WORK_TEMPLATE_REFRESH_MS 60000
+#define LOCAL_WORK_RETRY_DELAY_MS 5000
+#define LOCAL_WORK_RPC_TIMEOUT_MS 30000
+#define LOCAL_WORK_INITIAL_RESPONSE_BYTES 2048
+#define LOCAL_WORK_MAX_TEMPLATE_RESPONSE_BYTES (5 * 1024 * 1024)
+#define LOCAL_WORK_TEMPLATE_TAG "/Local Work ESP32/"
```
0xf0xx0, line 35: theres a tag also defined in local_work_builder.h

what about `/ESP-Miner/Local Work/`? gives downstream forks an opportunity to distinguish themselves

main/local_work_client.c
```diff
@@ -0,0 +1,823 @@
...
+    size_t tx_count_varint_len = encode_varint(ctx->tx_bundle->tx_count + 1, tx_count_varint);
+    size_t block_bytes = sizeof(header) + tx_count_varint_len + ctx->job.coinbase_tx_len;
+    for (size_t i = 0; i < ctx->tx_bundle->tx_count; i++) {
+        block_bytes += ctx->tx_bundle->tx_lengths[i];
+    }
+
+    char *rpc_url = nvs_config_get_string(NVS_CONFIG_LOCAL_WORK_BITCOIN_RPC_URL);
+    char *rpc_auth_mode = nvs_config_get_string(NVS_CONFIG_LOCAL_WORK_BITCOIN_RPC_AUTH_MODE);
+    char *rpc_username = nvs_config_get_string(NVS_CONFIG_LOCAL_WORK_BITCOIN_RPC_USERNAME);
+    char *rpc_password = nvs_config_get_string(NVS_CONFIG_LOCAL_WORK_BITCOIN_RPC_PASSWORD);
```
0xf0xx0, line 566: could this hammer the nvs on regtest?

main/thermal/EMC2302.c
```diff
@@ -90,7 +90,7 @@ static uint16_t get_fan_speed(uint8_t reg_addr, uint8_t multiplier)
     uint32_t rpm = 3932160UL * multiplier / tach_counter;
 
     if (rpm > UINT16_MAX) {
-        ESP_LOGW(TAG, "RPM %u exceeds uint16_t range, clamping", rpm);
```
0xf0xx0, line 93: separate these into a different pr?

main/http_server/axe-os/src/app/components/pool/pool.component.html
```diff
@@ -235,6 +235,102 @@ <h2 *ngIf="pool === 'fallbackStratum'" class="mt-5">Fallback Pool Configuration<
...
 
+        <h2 class="mt-5">Sovereign Work Generation</h2>
+        <div class="card">
+            <div class="field-checkbox grid">
+                <div class="col-1 md:col-10 md:flex-order-2">
+                    <p-checkbox name="localWorkEnabled" inputId="localWorkEnabled" formControlName="localWorkEnabled" [binary]="true"></p-checkbox>
+                </div>
+                <label for="localWorkEnabled" class="col-11 m-0 pb-0 pl-3 md:col-2 md:flex-order-1 md:pl-2">
+                    <tooltip-text-icon
+                        text="Prefer local work generation"
```
0xf0xx0, line 246: prefer -> use?

0xf0xx0 (commented):
this looks fun, ima give it a spin in a bit uwu

code nits commented, also local_work_builder could use with some doc comments on each function

mutatrum:
Related discussion: #901

components/stratum/include/mining.h
```diff
@@ -21,6 +27,9 @@ typedef struct
     double pool_diff;
     char *jobid;
     char *extranonce2;
+    bm_job_source_t source;
```
mutatrum, line 30: Is this something that can be tracked within local_work_builder, or local_work_client? I'd rather not have this leak out of the local work context.

components/stratum/local_work_builder.c
```diff
@@ -0,0 +1,1789 @@
...
+                                     : (tolower((unsigned char)hex[i * 2]) - 'a' + 10));
+        uint8_t low = (uint8_t)(isdigit((unsigned char)hex[i * 2 + 1])
+                                    ? hex[i * 2 + 1] - '0'
+                                    : (tolower((unsigned char)hex[i * 2 + 1]) - 'a' + 10));
+        out[i] = (high << 4) | low;
+    }
+    return ESP_OK;
+}
+
+static void reverse_bytes_32(const uint8_t src[32], uint8_t dest[32])
```
mutatrum, line 212: IIRC this already is available in utils?

components/stratum/local_work_builder.c
mutatrum: Agree, this is a lot of custom code for something there's already a library.

components/stratum/local_work_builder.c
```diff
@@ -0,0 +1,1789 @@
...
+    if (b58_sha256_impl == NULL) {
+        b58_sha256_impl = my_sha256;
+    }
+}
+
+static void *local_work_malloc(size_t size)
+{
+    void *ptr = NULL;
+#if CONFIG_SPIRAM
+    if (esp_psram_is_initialized()) {
```
mutatrum, line 49: We're not officially supporting the non-PSRAM hardware. If you want to play nice, just disable this whole feature if no PSRAM is available.

main/thermal/EMC2302.c
mutatrum: This is already (partially) covered in #1698

mutatrum:
Screenshots please.

For what I can see in the code, this is separate from the stratum/fallbackStratum pool config. IMO it would make more sense to have this on the same level of configuration, so the user can select SV1 / SV2 / GBT mining. But maybe I misread the code?

keegreil:
> Screenshots please.
> 
> For what I can see in the code, this is separate from the stratum/fallbackStratum pool config. IMO it would make more sense to have this on the same level of configuration, so the user can select SV1 / SV2 / GBT mining. But maybe I misread the code?

Screenshots attached!
The intent was to use a "prefer local work generation" setting.  If the node is not reachable, then it would fallback to the Stratum pools, first primary then fallback.  I'm sure I can mod this if you think a different flow is better.  

Screenshot from 2026-05-30 23-02-55
Screenshot from 2026-05-30 23-02-41

mutatrum:
> Screenshots attached! The intent was to use a "prefer local work generation" setting. If the node is not reachable, then it would fallback to the Stratum pools, first primary then fallback. I'm sure I can mod this if you think a different flow is better.

The idea is to support more than 2 pools in the future. Having a GBT option would just be an option of one or more of those entries. Having it on the same level as the current primary / secondary would make this easier.

PR #1720, opened 2026-05-26
Measure SV2 share response time per-share

warioishere:
The SV2 task measured share response time against a single timestamp (the last submit). If more than one share is in flight, the pool can ack several at once via `SubmitShares.Success` (`last_sequence_number` + `new_submits_accepted_count`), so measuring against the last submit gives a wrong round trip.

Now submit timestamps are stored per sequence number (small ring buffer) and the response time is measured against the `last_sequence_number` the pool acked. `sv2_parse_submit_shares_success` now also reads `last_sequence_number`, which was already in the payload but unused.

For pools that batch-ack, the elapsed time also includes the pool's batching window, not just the network round trip. So `new_submits_accepted_count` is exposed as `responseShareBatch` and the dashboard shows a small `(batched ×N)` next to the Time when it's >1.

SV1 unaffected, it already does 1:1 request/response timing.

### DIFF
--- a/main/global_state.h
+++ b/main/global_state.h
@@ -85,6 +85,7 @@ typedef struct
     bool pool_decode_coinbase_tx;
     bool fallback_pool_decode_coinbase_tx;
     float response_time;
+    uint16_t response_share_batch; // shares acked in the batch that produced response_time (SV2: 1 = single, >1 = batched)
     float process_time;
     float cpu_usage;
     bool use_fallback_stratum;
--- a/main/global_state.h
+++ b/main/global_state.h
@@ -85,7 +85,7 @@ typedef struct
     bool pool_decode_coinbase_tx;
     bool fallback_pool_decode_coinbase_tx;
     float response_time;
-    uint16_t response_share_batch; // shares acked in the batch that produced response_time (SV2: 1 = single, >1 = batched)
+    uint16_t response_share_batch;
     float process_time;
     float cpu_usage;
     bool use_fallback_stratum;
--- a/components/stratum_v2/include/sv2_protocol.h
+++ b/components/stratum_v2/include/sv2_protocol.h
@@ -160,7 +160,8 @@ int sv2_parse_set_target(const uint8_t *payload, uint32_t len,
                          uint32_t *channel_id, uint8_t max_target[32]);
 
 int sv2_parse_submit_shares_success(const uint8_t *payload, uint32_t len,
-                                    uint32_t *channel_id, uint32_t *accepted_count);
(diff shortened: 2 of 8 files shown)


mutatrum:
Why is measuring against the last submit not correct? I assume the pool will response as soon as the batch is complete? Having the pool batch responses makes the response time measurement somewhat iffy anyways, there's no guarantee the pool will actually respond directly.

warioishere:
Good question. SV1 doesn't hit this because it already keys submit timestamps per request id (`request_timings[]` in stratum_api.c, indexed by message_id) — each `mining.submit` has its own id, the response references it, lookup is per-id. SV2 instead used a single global `stratum_v2_last_submit_time_us` that's overwritten on every submit.

That matters because submits and ack-processing run on different tasks: `stratum_v2_submit_share` is called from `asic_result_task`, while `SubmitShares.Success` is handled in the `stratum_v2_task` receive loop, with no lock between them. So the ASIC task can overwrite the timestamp between the last share of an acked batch being sent and that ack being processed.

Illustrative:
```
seq 5 submitted t=0
seq 6 submitted t=100
seq 7 submitted t=200
seq 8 submitted t=240     <- ASIC task finds another share before the ack lands
ack t=250 for last_sequence_number=7   (covers 5,6,7 - not 8)
```
Measuring against last submit -> 250-240 = 10ms (timing seq 8, not acked). Against the acked seq 7 -> 250-200 = 50ms. When submits and acks are 1:1 they're identical; the difference only shows when a new share is submitted before the previous ack arrives.

So this PR basically brings SV2 in line with how SV1 already measures (per-request instead of a single global). You're right that batching stays inherently iffy — there's no clean round trip when the pool sits on shares — which is why it also surfaces `new_submits_accepted_count` as `responseShareBatch` and the dashboard flags `(batched x N)`, so a batched measurement is labelled rather than presented as a direct round trip.

main/global_state.h
```diff
@@ -85,6 +85,7 @@ typedef struct
     bool pool_decode_coinbase_tx;
     bool fallback_pool_decode_coinbase_tx;
     float response_time;
+    uint16_t response_share_batch; // shares acked in the batch that produced response_time (SV2: 1 = single, >1 = batched)
```
WantClue: remove the comment here, thats already explaind in openapi.yaml

WantClue (commented):
lgtm apart from my comment

main/global_state.h
warioishere: done

WantClue (approved):

PR #1722, opened 2026-05-26
Set TCP_NODELAY on pool sockets to fix SV2 submit latency

warioishere:
SV2 share submits were taking ~40-60ms on hardware where SV1 submits take ~5-10ms. Traced it with tcpdump at the pool:

- The firmware writes each `SubmitSharesExtended` as two TCP segments (encrypted header, then payload), and the pool socket had Nagle's algorithm active (TCP_NODELAY was never set).
- With Nagle on, the stack sends segment 1 and holds segment 2 until segment 1 is ACKed. The pool only has a partial message so it can't respond, and sends a standalone delayed-ACK after ~40ms. Only then does segment 2 go out, after which the pool responds in <0.1ms.
- SV1 sends its message in a single write/segment, which is why it was already fast.

Fix: set `TCP_NODELAY` on the pool socket right after connect, for both SV1 and SV2. Measured SV2 submit latency drops from 40-60ms to ~8ms, in line with SV1.

Verified with tcpdump at the pool: the submit now arrives without the ~40ms delayed-ACK gap and the pool responds in <1ms.

Before:
Bildschirmfoto_20260526_133902

After:

Bildschirmfoto_20260526_133808

### DIFF
--- a/main/tasks/stratum_v2_task.c
+++ b/main/tasks/stratum_v2_task.c
@@ -48,6 +48,13 @@ static void stratum_v2_set_socket_options(esp_transport_handle_t transport)
     setsockopt(sock, SOL_SOCKET, SO_SNDTIMEO, &snd_timeout, sizeof(snd_timeout));
     setsockopt(sock, SOL_SOCKET, SO_RCVTIMEO, &rcv_timeout, sizeof(rcv_timeout));
 
+    // Disable Nagle's algorithm. SV2 frames are written as two segments (encrypted
+    // header then payload); with Nagle on, the second segment is held until the
+    // first is ACKed, which collides with the pool's delayed-ACK and adds ~40 ms
+    // of latency per share submit.
+    int nodelay = 1;
+    setsockopt(sock, IPPROTO_TCP, TCP_NODELAY, &nodelay, sizeof(nodelay));
+
     int keepalive = 1;
     setsockopt(sock, SOL_SOCKET, SO_KEEPALIVE, &keepalive, sizeof(keepalive));
 
--- a/main/tasks/stratum_v1_task.c
+++ b/main/tasks/stratum_v1_task.c
@@ -210,6 +210,14 @@ static void set_socket_options(esp_transport_handle_t transport)
             ESP_LOGE(TAG, "Failed to set SO_RCVTIMEO");
         }
 
+        // Disable Nagle's algorithm so share submits are sent immediately instead
+        // of being held back to coalesce with later data, which interacts badly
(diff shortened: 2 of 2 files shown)


main/tasks/stratum_v2_task.c
mutatrum, line 37: Would it make sense to extract this method, as they are identical as far as I can tell?

warioishere: Agreed, they're identical apart from SV1 using globals for the timeouts and logging per-setsockopt while SV2 doesn't. Extracting it needs a shared home, and `components/stratum` currently has no transport/networking file — `utils.c` is hash/diff math, `mining.c` is job/merkle, `stratum_api.c` is the SV1 JSON-RPC protocol.

Two ways I see it:

1. Put a `STRATUM_set_socket_options(esp_transport_handle_t)` in `stratum_api.c` next to the existing `STRATUM_V1_transport_init` (both are transport setup). No new file, but it sits in the SV1-named file even though it's shared.

2. New `stratum_transport.c` / `.h` that owns shared stratum TCP transport helpers (single responsibility). Cleaner separation, costs one small file for now.

I lean towards 2 since it's the cleaner home and the SV2 task shouldn't have to pull in the SV1 API header just for socket options. Which do you prefer?

mutatrum: I would opt for `stratum_socket.c` / `.h` in `component/stratum` for `set_socket_options`. Also move `resolve_stratum_address` there, so the SV2 can leverage the non-blocking DNS resolver from SV1.

Some others that can be extracted/consolidated:
 * `is_wifi_connected` appears 3 times (v1, v2 and protocol_coordinator), that can be moved to `component/connect`.
 * `cleanQueue` / `stratum_v2_clean_queue` move to `main/system.c`

Maybe this is a follow-up PR as cleanup/code consolidation PR instead of overloading this one.

warioishere: Agreed — that's more than this PR should carry. I'd suggest merging this one as-is (just the TCP_NODELAY fix) and I'll do the consolidation in a dedicated follow-up PR: `stratum_socket.c`/`.h` for `set_socket_options` + `resolve_stratum_address` (so SV2 can use the non-blocking resolver too), `is_wifi_connected` into `component/connect`, and `cleanQueue` / `stratum_v2_clean_queue` into `main/system.c`. Keeps this fix small and the refactor reviewable on its own.

mutatrum (approved):
LGTM

PR #1724, opened 2026-05-27
Consolidate shared stratum socket, wifi and clean-queue helpers

warioishere:
Follow-up to #1722 — mutatrum suggested doing the socket/code consolidation in a separate PR instead of overloading the TCP_NODELAY change, so here it is.

The SV1 and SV2 tasks had grown several copies of the same code. This pulls them together:

- New `components/stratum/stratum_socket.{c,h}` holds the socket setup (`stratum_socket_set_options`) and the non-blocking DNS resolver (`stratum_socket_resolve`), both moved out of the SV1 task. SV2 now uses the resolver as well and connects by the resolved IP, so its DNS lookups no longer block the lwIP stack — SV1 already did this.
- `is_wifi_connected` existed three times (SV1, SV2 and protocol_coordinator). Moved into the connect component as `wifi_is_connected`.
- `cleanQueue` / `stratum_v2_clean_queue` merged into `SYSTEM_clean_jobs_queue` in system.c.

One behaviour change worth pointing out: the two clean-queue functions weren't identical. The SV1 one also reset the hashrate measurements (to avoid a spike on reconnect), the SV2 one didn't. The merged version keeps the reset, so SV2 now behaves like SV1 there.

Otherwise no functional change — same timeouts, same keepalive and TCP_NODELAY settings. Builds clean for esp32s3.

### DIFF
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -595,3 +595,9 @@ static const char *get_wifi_reason_string(int reason) {
     }
     return "Unknown error";
 }
+
+bool wifi_is_connected(void)
+{
+    wifi_ap_record_t ap_info;
+    return (esp_wifi_sta_get_ap_info(&ap_info) == ESP_OK);
+}
--- a/components/connect/include/connect.h
+++ b/components/connect/include/connect.h
@@ -4,6 +4,7 @@
 #include "lwip/sys.h"
 #include <arpa/inet.h>
 #include <lwip/netdb.h>
+#include <stdbool.h>
 
 #include "esp_err.h"
 #include "esp_wifi_types.h"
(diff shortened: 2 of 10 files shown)


mutatrum:
Surprising you end up with the same amount of code: `+281/-286`. Not sure where that comes from? There are 2 improvements as benefit, so that's at least one thing :-) Still think it's a good consolidation.

warioishere:
Yeah, the line stat is misleading. The real savings came from `set_socket_options` (was in both tasks) and the three `is_wifi_connected` copies. `resolve_stratum_address` only existed once in v1 — moving it to the shared file is roughly +120/-120 without deduplicating anything, and the new header with its doc comments plus the small SV2 wiring eat the rest. So the win is fewer copies, not fewer lines.

What's left between v1 and v2 is mostly protocol-specific (Noise/binary framing vs JSON-RPC parsing), so I don't think there's another meaningful consolidation hiding here.

mutatrum (approved):
LGTM, nice consolidation with benefits

WantClue:
@warioishere there is a merge conflict now please resolve

warioishere:
done

PR #1725, opened 2026-05-28
Improve dashboard performance

mutatrum:
* Cache values for faster rendering
 * Disable rendering completely when tab is not visible
 * Smarter fill in of statistics gaps after browser throttling

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,4 +36,6 @@ Thumbs.db
 debug.log
 .vscode
 .cache
-tools/devices.txt
+
+# Device list file for tools/upload2device.py
+devices.txt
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -36,3 +36,4 @@ Thumbs.db
 debug.log
 .vscode
 .cache
+tools/devices.txt
```
WantClue: ?

WantClue (approved):
lgtm whats the devices.txt ?

.gitignore
mutatrum: Used for the flash tool:
`python3 tools/upload2device.py --file tools/devices.txt`
I can get rid of it, it was just a convenient place to put it, as it's mentioned in the help file :laughing: 

Maybe it makes more sense in the root actually, just to follow the convention of the tool.

PR #1733, opened 2026-05-29
fix: disable browser autocomplete for pool password fields

### LINKED ISSUE #1727: autocomplete for password field
we should use autocomplete="off" for the password fields in pool.component.*

that way a browser with password manager doesn't try to fill it

mbmtbtbc:
- Added `autocomplete="off"` to pool password inputs in `pool.component.html`

fixes #1727


WantClue (approved):
thanks

PR #1736, opened 2026-05-30
Add multi-line charts

### LINKED ISSUE #1476: ASIC Temp chart misleading on multi chip units

### LINKED ISSUE #1730: Add multi-line charts
Change the chart axis dropdowns to units instead of sensors. If there are multiple sensors on the same unit (f.e. Temperature) it's possible to show multiple values.

Fixes #1676 

Proof of concept:
Image

mutatrum:
Changes the chart selection dropdowns to the unit, instead of individual sensors. This allows for example multiple temperature sensors to be shown at the same time:

image

Gamma GT:

image

Fixes #1730


main/http_server/axe-os/src/app/components/home/home.component.scss
```diff
@@ -17,6 +17,7 @@
 
     td {
         transition: background-color 2s ease-in-out;
+        background-color: oklch(from var(--primary-color) var(--lightness, 0.5) c h);
```
mutatrum: ✨ CSS magic ✨

0xf0xx0 (approved):

0xf0xx0:
hmm, nit: "Percentage (%)" should be "Fan Speed (%)"

mutatrum:
> hmm, nit: "Percentage (%)" should be "Fan Speed (%)"

That one's a bit weird, as error percentage is also under percentage.

PR #1737, opened 2026-05-31
Filter out non-hashing devices in swarm

mutatrum:
Swarm shows last values when a device is paused. But now they're greyed out and not counted at the total hashrate and power stats on top.

image

(It's warm here, most devices are turned off)


ghost:
image
 


The popup bottom right was expected every time the poll counter expires as that device was turned off (being cleaned at the time).
  

image
  

Grid view.

ghost:
Revisited - Tested, no issues found other than the visual one for the popup boxes (bottom right) when buttons are clicked on for each device (PR#1788 should resolve this).

image

ghost:
No issues found, works well.

0xf0xx0:
will merge next, pls rebase uwu

0xf0xx0:
oh wait i misread the out of date prompt KEK, any blocks before i merge?

WantClue:
Devices that are reachable, unpaused, and at 0 H/s belong to no bucket
- hashingDevices (:529) excludes them (!axe.hashRate)
- pausedDevices (:530) excludes them (not paused)
- notAccessibleDevices (:531) excludes them

The three numbers no longer sum to swarm.length, and the card still says "Total Devices" while showing only the hashing count. A device that's booting, recovering from a pool disconnect, or in overheat/power-fault just vanishes from the summary. Your own spec covers this exactly — device3 (hashRate 0, accessible, unpaused) is in the swarm but appears in none of the asserted counts.

Either keep swarm.length as "Total Devices" with hashing as a sub-line, or add a fourth bucket ("idle" / "not hashing")

WantClue:
getDeviceNotification() (:690-707) has a Paused badge but no notAccessible case. Combined with .not-hashing { opacity: 0.6 }, paused / idle / unreachable all render identically — a dimmed row of zeros with no explanation of which it is. Add a case !!axe.notAccessible: return { color: 'red', msg: 'Not accessible' } (before the miningPaused case, since it's the stronger signal). Also note the row's action buttons (restart/identify) remain clickable while looking disabled.

main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -570,6 +593,7 @@ private isIpAddress(value: string): boolean {
       overheat_mode: null,
       isUsingFallbackStratum: null,
       blockFound: null,
+      notAccessible: null,
```
WantClue: **notAccessible is never cleared — the reset lives in dead code**
notAccessible: null to mergeDeviceData(). That method has zero callers anywhere in src/ — it's dead code. The live merge path is getAllDeviceInfo() at swarm.component.ts:321-330:

const existingDevice = this.swarm.find(device => device.connectionAddress === address);
const result = {
  ...,
  ...(existingDevice ? existingDevice : {}),   
  ...info,                                     
  ...asic,

So once refreshErrorHandler (:444) stamps notAccessible: true, nothing ever removes it: the device stays dimmed, stays excluded from Total Hashrate/Power, and stays counted under "not accessible" even after it comes back online. Worse, saveSwarmData() persists it to localStorage, so a reload doesn't clear it either — one transient timeout permanently poisons the entry.

ghost:
Please resolve conflicts and check against #1887

mutatrum:
> Please resolve conflicts and check against #1887

Done.

PR #1739, opened 2026-05-31
Add camelCase overheat mode API alias

### LINKED ISSUE #621: bug: API formatting inconsistency

vortexopenclaw:
Fixes #621.

## Summary
- Add `overheatMode` to `/api/system/info`.
- Keep the existing `overheat_mode` field for backward compatibility.
- Document both fields in OpenAPI, with `overheat_mode` marked as the compatibility alias.
- Update AxeOS to prefer `overheatMode` while falling back to `overheat_mode`.

## Root cause
Most system-info response fields use camelCase, but overheat status was exposed only as `overheat_mode`. That leaves API clients and generated models with a one-off snake_case field, which is the inconsistency reported in #621.

## Fix
This keeps the API backward-compatible by adding the camelCase alias without removing the existing snake_case field. Existing UI clients and third-party API consumers that read `overheat_mode` continue to work, while new clients can use `overheatMode`.

## Verification
- `npm run test:ci` in `main/http_server/axe-os` passed: `33 SUCCESS`.
- Generated ignored `SystemInfo` model contains both `overheatMode` and `overheat_mode`.
- `git diff --check` passed.

## Hardware validation
- OTA-tested on Gamma `local test device` with firmware and web UI artifacts from this branch.
- Verified the device booted, AxeOS/API were reachable, mining resumed, and the patched `/api/system/info` response included the new `overheatMode` field while preserving `overheat_mode`.
- Rolled the device back to stock `v2.13.1` afterward; API remained healthy and mining resumed.

### DIFF
--- a/main/http_server/axe-os/src/app/components/home/home.component.ts
+++ b/main/http_server/axe-os/src/app/components/home/home.component.ts
@@ -1088,7 +1088,7 @@ export class HomeComponent implements OnInit, OnDestroy {
 
     updateMessage(!!systemInfoError.duration, 'SYSTEM_INFO_ERROR', 'error', `Unable to reach the device for ${DateAgoPipe.transform(systemInfoError.duration, { strict: true })}`);
     updateMessage(!!(info as any).miningPaused, 'MINING_PAUSED', 'warn', 'Mining is paused');
-    updateMessage(!!info.overheat_mode, 'DEVICE_OVERHEAT', 'error', 'Device has overheated - See settings');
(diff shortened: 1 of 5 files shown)


vortexopenclaw:
## Validation proof

Commands run:
- `npm run test:ci` in `main/http_server/axe-os`: 33 SUCCESS
- `git diff --check`

API and compatibility check:
- Stock v2.13.1 devices expose `overheat_mode` in `/api/system/info`.
- This branch keeps `overheat_mode` and adds `overheatMode`, so existing clients remain compatible while generated camelCase clients can use the new field.
- OTA smoke on Bitaxe Gamma passed: API/UI recovered after update, then the device was rolled back healthy.

PR #1740, opened 2026-05-31
Clamp SV1 version rolling mask to ASIC-supported bits

### LINKED ISSUE #1169: hardcoded version-rolling.mask might exceed what ASIC can roll

vortexopenclaw:
Fixes #1169.

## Summary
- Advertise `1fffe000` for SV1 version rolling on BM1366, BM1368, and BM1370 instead of `ffffffff`.
- Clamp pool-provided `mining.configure` and `mining.set_version_mask` values before storing or applying them.
- Keep BM1397 all-bits compatibility behavior unchanged.
- Add focused coverage for configure JSON and ASIC mask clamping.

## Root cause
The SV1 `mining.configure` request advertised `ffffffff` as the supported `version-rolling.mask`. BM1366, BM1368, and BM1370 program only the 16-bit roll value derived from `version_mask >> 13`, so their practical supported mask is `0x1fffe000`. Advertising or applying unsupported bits can let a pool assign version bits the miner cannot actually roll.

## Fix
This adds ASIC-level supported-mask helpers, uses the supported mask when requesting SV1 version rolling, and clamps pool-returned masks before applying them to mining state. BM1397 keeps the previous behavior because its version-mask setter is currently a placeholder and this PR should not change that runtime path.

## Verification
- `git diff --check` passed.
- `ESP-IDF environment loaded, then GITHUB_ACTIONS=true idf.py -C test-ci build` passed.

## Hardware validation
- OTA-tested on Gamma `local test device` / BM1370 with firmware from this branch.
- OTA-tested on Supra `local test device` / BM1368 with firmware from this branch.
- Both devices booted, AxeOS/API were reachable, mining resumed, and latest snapshots showed `0` rejected shares after warmup.
- Both devices were rolled back to stock `v2.13.1` afterward; API remained healthy and mining resumed.

## Remaining risk
The local `test-ci build` compiles the unit-test app but does not execute those tests on hardware/QEMU. A stronger mining-impact proof would use a controlled SV1 pool that returns masks outside `0x1fffe000` and compare accepted shares/rejects before and after on BM1368/BM1370 hardware.

### DIFF
--- a/components/asic/asic.c
+++ b/components/asic/asic.c
@@ -10,6 +10,7 @@
 #include "asic.h"
 #include "device_config.h"
 #include "frequency_transition_bmXX.h"
+#include "utils.h"
 
 static const char *TAG = "asic";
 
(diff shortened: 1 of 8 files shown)


vortexopenclaw:
## Validation proof

Commands run:
- `git diff --check`
- `ESP-IDF environment loaded, then GITHUB_ACTIONS=true idf.py -C test-ci build`

Test coverage:
- Configure JSON test now expects `1fffe000` instead of `ffffffff` for BM1370.
- ASIC clamp tests cover BM1370 masks: `ffffffff -> 1fffe000`, `00001fff -> 00000000`, `e0000000 -> 00000000`.
- BM1397 compatibility coverage keeps all-bits behavior unchanged.

Hardware smoke:
- OTA-tested on Bitaxe Gamma and BM1368-based Bitaxe.
- Both booted, API/UI stayed reachable, mining resumed, latest snapshots showed `0` rejected shares after warmup, then both were rolled back to v2.13.1.

PR #1741, opened 2026-05-31
Report found blocks only after pool acceptance

### LINKED ISSUE #1552: fw should only report a found block if share was submitted and accepted

vortexopenclaw:
Fixes #1552.

## Summary
- Stop marking `blockFound` when the ASIC only finds a locally valid nonce.
- Track block-candidate shares only after a share submit is successfully written.
- Promote a candidate to `blockFound` only when the pool accepts that exact share.
- Correlate SV1 candidates by JSON-RPC request id and SV2 candidates by SubmitShares sequence number.

## Root cause
`SYSTEM_notify_found_nonce()` updated the UI/block counters as soon as a local nonce met network difficulty. In setups where network difficulty is lower than the pool/share difficulty, a nonce can satisfy the local block-candidate check while never being submitted or while later being rejected by the pool. That can show a false found-block state.

## Fix
This separates best-diff tracking from accepted-block notification. `SYSTEM_notify_found_nonce()` keeps updating best/session diff only. `SYSTEM_is_block_candidate()` decides whether a successfully submitted share should be tracked as a block candidate. `SYSTEM_notify_accepted_block()` updates `blockFound` and the new-block banner only after pool acceptance.

For correlation, SV1 stores exact request ids instead of relying on the modulo slot alone. SV2 stores exact SubmitShares sequence numbers and keeps accepted/rejected share accounting pool-authoritative while using exact matches only for response timing, accepted-block notification, and slot cleanup.

## Verification
- `git diff --check` passed.
- `ESP-IDF environment loaded, then GITHUB_ACTIONS=true idf.py -C test-ci fullclean && GITHUB_ACTIONS=true idf.py -C test-ci build` passed.

## Hardware validation
- Smoke-tested the branch on Gamma `local test device`.
- Verified the device booted, AxeOS/API were reachable, mining resumed, and the post-test snapshot remained healthy.
- Rolled the device back to stock `v2.13.1` afterward; API remained healthy and mining resumed.

## Not tested
- I did not reproduce the original regtest-pool setup from #1552 locally. The fix is based on the code path that previously incremented `blockFound` before pool acceptance, plus firmware/test-ci builds and a live mining smoke test.

### DIFF
--- a/main/system.c
+++ b/main/system.c
@@ -310,6 +310,31 @@ void SYSTEM_notify_rejected_share(GlobalState * GLOBAL_STATE, char * error_msg)
     }    
 }
 
+bool SYSTEM_is_block_candidate(GlobalState * GLOBAL_STATE, double diff, uint8_t job_id)
+{
+    pthread_mutex_lock(&GLOBAL_STATE->valid_jobs_lock);
+    bool valid = (GLOBAL_STATE->valid_jobs[job_id] != 0);
+    bm_job *active_job = valid ? GLOBAL_STATE->ASIC_TASK_MODULE.active_jobs[job_id] : NULL;
+    uint32_t target = active_job != NULL ? active_job->target : 0;
+    pthread_mutex_unlock(&GLOBAL_STATE->valid_jobs_lock);
+
+    if (!valid || active_job == NULL) {
+        return false;
+    }
+
+    double network_diff = networkDifficulty(target);
+    return diff >= network_diff;
+}
+
+void SYSTEM_notify_accepted_block(GlobalState * GLOBAL_STATE, double diff)
+{
+    SystemModule * module = &GLOBAL_STATE->SYSTEM_MODULE;
+
+    module->block_found++;
+    module->show_new_block = true;
+    ESP_LOGI(TAG, "FOUND BLOCK accepted by pool: %f (count: %d)", diff, module->block_found);
+}
+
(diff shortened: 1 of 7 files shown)


vortexopenclaw:
## Validation proof

Commands run:
- `git diff --check`
- `ESP-IDF environment loaded, then GITHUB_ACTIONS=true idf.py -C test-ci fullclean && GITHUB_ACTIONS=true idf.py -C test-ci build`

Behavior check:
- Before: `SYSTEM_notify_found_nonce()` incremented `block_found` immediately when local nonce diff met network difficulty, before pool acceptance.
- After: local nonce discovery only updates best/session diff. `blockFound` increments only through `SYSTEM_notify_accepted_block()` after the matching submitted share is accepted by the pool.
- SV1 matching uses exact JSON-RPC request IDs. SV2 matching uses exact SubmitShares sequence numbers.
- Bitaxe Gamma mining smoke passed and the device was rolled back healthy.

Not covered locally: a full original regtest pool reproduction of a candidate network-difficulty block.

Issue #773, opened 2025-03-14
Webserver Crash When Using AxeOS via Mobile Browser


LisBerndt:
When operating the AxeOS via a mobile browser (in this case, Yandex), the web server crashes after a few actions (viewing logs, opening the dashboard, change settings etc.). The API gets unreachable and in some cases, the device even becomes unreachable via ping. As a test, I used the OS exclusively through a desktop browser (Chrome) for an entire day—without any issues.

Bitaxe Gamma BM1370
AxeOS 2.5.1

Uefi1:
Yeah, I also noticed that the web interface sometimes disconnects. I have to turn off the ASIC, wait for it to cool down, and then turn it back on.

NotAnotherHelloWorld:
I experience similar issues. Web frontend worked fine for a week without issues, but as soon as I started collecting stats using curl on API endpoint, the webserver becomes unresponsive after a few hours. Ping works fine even when web is down. WiFi deauth and letting it reconnect solves this, without having to reboot the device. I'll try to see if I can gather something useful from the logs when I have the time, maybe it's not closing sessions correctly.

Uefi1:
> I experience similar issues. Web frontend worked fine for a week without issues, but as soon as I started collecting stats using curl on API endpoint, the webserver becomes unresponsive after a few hours. Ping works fine even when web is down. WiFi deauth and letting it reconnect solves this, without having to reboot the device. I'll try to see if I can gather something useful from the logs when I have the time, maybe it's not closing sessions correctly.

I have some suspicions that the ESP32 chip is overheating because it's located near the heatsink, where the fan blows out hot air. Some of that hot air might be hitting the ESP32, which could be why the web interface stops responding.

LisBerndt:
> > I experience similar issues. Web frontend worked fine for a week without issues, but as soon as I started collecting stats using curl on API endpoint, the webserver becomes unresponsive after a few hours. Ping works fine even when web is down. WiFi deauth and letting it reconnect solves this, without having to reboot the device. I'll try to see if I can gather something useful from the logs when I have the time, maybe it's not closing sessions correctly.
> 
> I have some suspicions that the ESP32 chip is overheating because it's located near the heatsink, where the fan blows out hot air. Some of that hot air might be hitting the ESP32, which could be why the web interface stops responding.
> 
> BitAxe is probably poorly designed—it would have been better to solder the ESP32 chip on the back of the board rather than placing it next to the processor's cooling system.

[...]

PR #1742, opened 2026-05-31
Clean up stale WebSocket clients safely

vortexopenclaw:
Related to #773.

## Summary
- Keep WebSocket client registry updates under the existing mutex.
- Prune stale WebSocket file descriptors before counting or broadcasting.
- Snapshot broadcast recipients under the mutex, then send outside the mutex.
- Drop failed or no-longer-WebSocket clients from ESP-Miner's registry.
- Avoid logging while holding the WebSocket client mutex, because log handling can re-enter WebSocket client-count checks.

## Root cause
The HTTP server can keep running while ESP-Miner's WebSocket registry retains stale client entries. Before this patch, failed async sends only logged a warning and `websocket_broadcast()` iterated the shared client list without the mutex used by add/remove. Once stale clients consume the small WebSocket slot pool, the web UI/API can become difficult or impossible to use while the device still responds to ping.

During hardware testing, an earlier cleanup revision exposed another failure mode: logging while holding the WebSocket client mutex could re-enter the WebSocket log/count path and leave the firmware repeatedly failing to acquire that mutex before a reset. This patch keeps socket checks and sends outside the mutex, and moves add/update logs after the mutex is released.

## Fix
The patch adds shared helpers for client type validation, mutex acquisition, stale-client pruning, and locked registry removal. Count and broadcast paths prune inactive ESP-IDF WebSocket fds first, then use short mutex-protected sections only for registry reads/writes. Broadcast sends happen after the snapshot is copied, so a slow or dead client cannot hold the registry mutex.

## Verification
- `git diff --check` passed.

## Hardware validation
- OTA-tested on Gamma `local test device` with firmware built from this branch.
- Reproduced the previous raw `/api/ws/live` WebSocket handshake plus masked CLOSE-frame probe 5 times.
- Before the final mutex/logging correction, this probe caused `Failed to acquire WebSocket client mutex` spam followed by a software reset due to exception/panic.
- With this PR's final commit, the same 5-iteration probe completed successfully.
- Post-probe `/api/system/info` stayed HTTP 200, uptime continued increasing, reset reason stayed `Software reset via esp_restart`, mining resumed around normal Gamma hashrate, and rejected shares stayed at `0`.
- Post-probe logs for `v2.14.0b4-1-ge950ddd` showed the WebSocket connects without the mutex-acquire failure spam.

## Remaining risk
This is focused on stale WebSocket registry cleanup and the mutex re-entry issue seen during testing. It does not attempt to redesign Wi-Fi scan behavior or broader HTTP server session management.

### DIFF
--- a/main/http_server/websocket.c
+++ b/main/http_server/websocket.c
@@ -1,4 +1,7 @@
 #include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
 #include <unistd.h>
 #include "freertos/FreeRTOS.h"
 #include "freertos/task.h"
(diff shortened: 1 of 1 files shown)


vortexopenclaw:
## Validation proof

Repro used:
- Raw `/api/ws/live` WebSocket handshake plus masked CLOSE frame, repeated 5 times.

Before final fix:
- The same probe on the earlier cleanup revision produced repeated `Failed to acquire WebSocket client mutex` logs and then a software reset due to exception/panic.

Final fix:
- Commit tested: `e950ddd`.
- OTA-tested on Bitaxe Gamma as `v2.14.0b4-1-ge950ddd`.
- 5-iteration raw WebSocket CLOSE probe completed successfully.
- Post-probe `/api/system/info`: HTTP 200, uptime continued increasing, reset reason stayed `Software reset via esp_restart`, mining resumed around normal Gamma hashrate, `sharesRejected=0`.
- Post-probe logs showed WebSocket connects without the mutex-acquire failure spam.
- Bitaxe Gamma was rolled back to v2.13.1 afterward and confirmed healthy.

vortexopenclaw:
Hey, saw this was closed without comment. I’m trying to understand the reason so I can adjust future PRs appropriately.

Was this closed because the fix is unwanted, the approach is wrong, the issue isn’t considered valid, the validation isn’t sufficient, or because of contribution/process concerns? Happy to revise or split differently if useful.

nathang21:
@0xf0xx0 could clarify? #773 is very much still an issue. 

Nearly everytime I get curious and want to check my bitaxe, or if there is an update and I want to flash it, I notice the webserver is down, and I have to hard restart the device for it to come back,

0xf0xx0:
@nathang21 basically, this is an openclaw instance and we'd rather not have fully automated prs and issues being opened. leave a comment on 773 with your logs to help us figure out the real issue, it doesnt appear to be a common one.

vortexopenclaw:
True, this started off as me integrating my Openclaw into my Bitaxes and helping with development. I'll do a better job moving forward with focusing on discussion beforehand before telling OC to run off on its own. :) Makes sense that you don't want a bunch of agents firing off random disjointed stuff all over the place.

Small update from testing on current v2.14.0:

I moved the WebSocket cleanup work onto a fresh v2.14.0 baseline and tested it on one of my Gammas. I also kept another Gamma on stock v2.14.0 as a comparison.

So far:

- stock v2.14.0 did not reproduce the older raw `/api/ws/live` close-frame crash I had seen while testing earlier candidate builds
- patched v2.14.0 builds cleanly
- patched v2.14.0 is running on a Gamma and mining normally
- both stock v2.14.0 and patched v2.14.0 survived local WebSocket stress checks:
  - repeated `/api/ws/live` open/close cycles
  - repeated `/api/ws` open/close cycles
  - TCP-drop/replacement tests
  - idle WebSocket hold tests while the API stayed reachable

So I don’t want to overclaim this as a confirmed fix for #773. I’m not currently reproducing the original “ping works, AxeOS/API dead” state on stock v2.14.0, but I'll keep an eye out if any of the Bitaxes go down again in the future.

What I can say is that the cleanup branch is now based on v2.14.0 and has passed local no-regression WebSocket testing on hardware. If maintainers think stale WebSocket client cleanup is still useful, I’m happy to share the branch and validation logs.

PR #1743, opened 2026-06-01
Clean up SV2 NVS configuration

mutatrum:
This should bring it more in line with the rest of the NVS config.

**Note**: this is a breaking change from beta3 and beta4. There's an upgrade path, but the downgrade path is not there. So if you install this, you most likely can't go back to beta3 or beta4 without a factory reset.


main/nvs_config.c
```diff
@@ -112,15 +120,6 @@ static Settings settings[NVS_CONFIG_COUNT] = {
     [NVS_CONFIG_TPS546]                                = {.nvs_key_name = "TPS546",          .type = TYPE_BOOL},
     [NVS_CONFIG_TMP1075]                               = {.nvs_key_name = "TMP1075",         .type = TYPE_BOOL},
     [NVS_CONFIG_POWER_CONSUMPTION_TARGET]              = {.nvs_key_name = "power_cons_tgt",  .type = TYPE_U16},
-    // Protocol and channel-type are TYPE_U16 internally but exposed as strings via the API
-    // (see http_server.c GET/PATCH special-case handling). rest_name is omitted so the generic
```
mutatrum, line 116: No special handling

main/http_server/openapi.yaml
```diff
@@ -417,9 +417,6 @@ components:
           type: string
           enum: [SV1, SV2]
           description: Configured stratum protocol
-        activeProtocolLabel:
```
mutatrum, line 420: The rest of the active pool fields were calculated on the front-end, not on the back-end.

0xf0xx0 (approved):
code ack

PR #1750, opened 2026-06-03
Add AxeOS basic HTTP authentication

### LINKED ISSUE #1220: unlock OTA

mutatrum:
Proposal for simple basic HTTP authentication on the dashboard and API. It has 2 forms: only for modifications (including firmware updates), and complete lockdown. 

There's also support in swarm, but that's a bit tricky, as the password is kept in local storage. It's easiest to give all Bitaxe devices the same password, but it should work with different passwords. Haven't tested the swarm support yet.

Fixes #1220


philipbr:
I was looking for this feature 🤓 looking forward to having it released!

PR #1754, opened 2026-06-05
Buffer overflow in rejected-share stats.

Zayno:
Claude: if (module->rejected_reason_stats_count < sizeof(module->rejected_reason_stats)) { The array is RejectedReasonStat rejected_reason_stats[10] (global_state.h:56), and each element is 68 bytes, so sizeof(...) is 680, not the element count 10. The guard therefore lets rejected_reason_stats_count grow to 679, writing far past the 10-slot array into adjacent SystemModule fields. It triggers whenever a pool sends more than 10 distinct reject-reason strings over a session (entirely possible - the pool controls those strings). Fix: compare against ARRAY_SIZE (10).


WantClue (approved):

WantClue:
Good finding thank you

mutatrum:
There is an `ARRAY_SIZE` macro in use, but it's now defined in two separate places (`display.c` and `device_config.c`). Maybe extract that to a general place (maybe `utils.h`?) so it can also be used here.

WantClue:
> There is an `ARRAY_SIZE` macro in use, but it's now defined in two separate places (`display.c` and `device_config.c`). Maybe extract that to a general place (maybe `utils.h`?) so it can also be used here.

agreed, but this should be handled in a seperate PR

main/system.c
goldenpants85-prog, line 358: Congratulations!

Issue #1732, opened 2026-05-29
Just wanted to note a few things


mrv777:
Ran a little security check and a few things got mentioned I thought I would just pass along:

- /api/theme bypasses the normal network gate and has unsafe type handling. Unlike other API handlers, it does not call is_network_allowed(). POST also accepts unvalidated JSON types and can pass NULL into string handling.
- OTAWWW erases Axe-OS before validating www.bin. A bad or zero-length upload can wipe/corrupt the UI partition before failure is detected.
- Production frontend dependencies have known high CVEs. npm audit --omit=dev reports 9 high vulnerabilities in Angular/PrimeNG packages.

WantClue:
Thank you for letting us know, taking a look into that if any of these actually are exploitable.

mutatrum:
> Ran a little security check and a few things got mentioned I thought I would just pass along:
> 
> * /api/theme bypasses the normal network gate and has unsafe type handling. Unlike other API handlers, it does not call is_network_allowed(). POST also accepts unvalidated JSON types and can pass NULL into string handling.

See #1637, not sure if anything is already fixed there?

> * OTAWWW erases Axe-OS before validating [www.bin](http://www.bin). A bad or zero-length upload can wipe/corrupt the UI partition before failure is detected.

Ideally we should have a 2nd partition for www as well. No idea if this is feasible.

> * Production frontend dependencies have known high CVEs. npm audit --omit=dev reports 9 high vulnerabilities in Angular/PrimeNG packages.

First step is #1651 which fixes quite a few. Idea is to keep upgrading, as Angular 19 is also EOL.

mutatrum:
First point: #1759
Second point: #1763 aka the best part is no part.
Third point: #1651

PR #1759, opened 2026-06-08
HTTP Handler Audit

mutatrum:
As mentioned in #1732, first item:
> /api/theme bypasses the normal network gate and has unsafe type handling. Unlike other API handlers, it does not call is_network_allowed(). POST also accepts unvalidated JSON types and can pass NULL into string handling.

 * Add `is_network_allowed` to `/api/theme` and `/api/system/wifi/scan`
 * JSON type safety on several places
 * Several OOM protections


0xf0xx0 (approved):
ack

WantClue (approved):

Issue #1002, opened 2025-06-04
Unified firmware update


mutatrum:
Currently having a 2-stage firmware update is error prone and confusing. Secondary goal is to step away from the strict filenames. There are a few options to do this:

1. Merge `esp-miner.bin` and `www.bin` into a single binary file. This is a custom format, and needs some sort of simple header + size. One thing we could add into this header is a device marker, so we can do a validity check on that instead of the filename to prevent people of flashing the wrong firmware;

2. Have a zip file with both files. Same idea here, and it's trivial to add simple manifest file for device validity check;

3. Geet rid of `www.bin` and have the AxeOS files embedded into `esp-miner.bin`. I haven't looked at this in detail, but it's worth exploring this as well, as this makes the upgrade path way simpler. We can just forget about the spiffs partition when this is loaded onto the device.

The first two need a clear upgrade/downgrade path, and we need to supply both variants for a while, possibly adding to the confusion.

skot:
Having a unified firmware update image would be so helpful. I suspect a lot of update errors/confusion are caused by having to update two separate images.

My Search skills are failing me right now, But I know others have looked into this previously, like @tdb3 and @benjamin-wilson. IIRC the main issue is that a combined image is too big to fit into the Factory App partition (and corresponding OTA partition).

We should be able to change the default partition size though. Can we make it big enough for a combined image?

ghost:
#910 should cover it for now with ref to people not updating with both files that was included in v2.8.0.

mutatrum:
Previous issue with similar ideas and some technical exploration: #308

sz4bi:
I vote for Option 2, it is the easiest to implement and require no modifications in the system.

duckaxe:
Related https://github.com/bitaxeorg/ESP-Miner/discussions/1179

WantClue:
Community agreement atm is to keep it in for potential alternative WebUI like AxeWellUi

Issue #308, opened 2024-08-28
Unify the firmware images


skot:
right now we have two firmware images required for an update; `esp-miner.bin` and `www.bin`

It's nice because the `www.bin` image is much bigger and you don't always need it. 

But it's really complicated for end users. especially if one of the updates fails and they don't notice it.

I'd like to unify the images into one bigger one to simplify the update process.

tdb3:
I won't have time to implement this for at least a few weeks, but I looked into it a little bit.  Some thoughts/options below:

### Option 1: Embed Axe-OS web files into OTA partitions
- Embed Axe-OS files into `ota_0`/`ota_1`. These files are normally stored in `www` (spiffs partition)
- [This page](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/build-system.html?highlight=binary%20data#embedding-binary-data) describes a bit on how this might be done. [Here's an example](https://github.com/espressif/esp-idf/blob/c9763f62dd00c887a1a8fafe388db868a7e44069/examples/protocols/http_server/file_serving/main/file_server.c#L102-L108) of an application doing something similar.
- Might need to remove the `factory` partition (or make it smaller) to free up space, and operate solely from `ota_0`/`ota_1` (e.g. using `CONFIG_BOOTLOADER_APP_ROLLBACK_ENABLE`)
- OTA updates would then consist of a single file containing both the main app and axe-os web interface
- Rollback for a bad OTA update adds resiliency/redundancy for axe-os (in addition to the main app)

### Option 2: Package `esp-miner.bin` and `www.bin` together
- User sees and works with one firmware file (e.g. `esp-miner.zip`), but in reality, both firmware files are archived within it
- The archive is automatically extracted client-side in the web app, then the web app posts each file to the respective endpoints (`api/system/OTAWWW`, `api/system/OTA`) in sequence (e.g. `www.bin`, then `esp-miner.bin`)
- Could keep the single `www-data` partition
- Might be more prone to failure (more edge cases, less atomic)

tdb3:
In #649, the recovery page is refactored to use the approach described in the second bullet of option 1. Seems to work well. The next steps would include: updating http_server to serve the full web interface files this way (looks like the files might be < 1MB so could fit into the 4MB ota partition, adjusting partitioning to deprecate/remove the `www` partition, adjust building, etc.

tdb3:
Some thinking out loud:

[...]

PR #1763, opened 2026-06-15
Unified firmware

mutatrum:
This PR embeds Axe-OS into `esp-miner.bin` to create a unified firmware.

It is still possible to upload a custom front-end, but this is a separate configuration option, as it's no longer needed for default installs.

It also exposes the currently installed firmware versions, as there are several partitions (`factory`, `ota_0` and `ota_1`). This should make rolling back easier.

The `esp-miner.bin` is only 450kB bigger, instead of the full 3MB of the `www.bin` file. This was because the SPIFFS partition image of `www.bin` was not truncated, even though only 450kB of the 3MB was used. So the unified firmware is only a little bit larger, now clocking in at 2.4MB which leaves plenty of room on the 4MB partition.

image

Prior issue: #308 and #1002. This PR implements option 3 from 1002. The closing of these issues was done when Axewell-UI was released. Having the configuration option makes it possible to still use external developed front-ends.

Reason 3 was done to preserve the Angular filenames, which have a random hash at the end for cache-busting purposes. Other solutions would be trickier to get right, and disabling the cache-busting feature from Angular is not a viable option either.


main/http_server/axe-os/src/app/components/update/update.component.ts
```diff
@@ -187,4 +193,46 @@ export class UpdateComponent {
 
     this.localStorageService.setBool(IGNORE_RELEASE_CHECK_WARNING, true);
   }
+
+  public switchPartition(label: string): void {
+    if (confirm(`Set ${label} as the next boot partition? You will need to restart the device to apply this change.`)) {
```
0xf0xx0: nit: remove second sentence, device gets rebooted anyway

0xf0xx0 (changes requested):
couple nits: 

- had a write error get thrown mid-progress bar, but still had to wait for the bar to complete before the error was displayed
- maybe the restart modal on successful upload should clear itself when the page refreshes?
- can we have a button to delete partitions?
- choosing which partition to upload to would be cool too

main/http_server/axe-os/src/app/components/update/update.component.ts
mutatrum: Fixed popup, and device automatically reboots on confirm.

mutatrum:
> * had a write error get thrown mid-progress bar, but still had to wait for the bar to complete before the error was displayed

What caused the write error? It should handle it better now, but not sure how to test/reproduce this?

> * maybe the restart modal on successful upload should clear itself when the page refreshes?

Fixed.

> * can we have a button to delete partitions?

Why? I mean, yes, it's possible to add a delete button next to not-active partitions, not sure why you would want this. They would be marked as empty, not actually deleted. We cannot delete or change the partition table without forcing everyone to do a factory flash, it's a tiny miracle this is even possible without any partition table changes.

> * choosing which partition to upload to would be cool too

You can't upload to factory partition. And you always upload to the other OTA partition, you can't upload to the current active one. It's all managed by the ota library.

0xf0xx0:
> What caused the write error? It should handle it better now, but not sure how to test/reproduce this?

likely the same issue that triggers #764, sending a malformed request.


> Why? I mean, yes, it's possible to add a delete button next to not-active partitions, not sure why you would want this. They would be marked as empty, not actually deleted. We cannot delete or change the partition table without forcing everyone to do a factory flash, it's a tiny miracle this is even possible without any partition table changes.

well, not delete-delete, but clear em? marking as empty would be good enough i think

> You can't upload to factory partition. And you always upload to the other OTA partition, you can't upload to the current active one. It's all managed by the ota library.

ahhh i seee, ok nvm then

0xf0xx0:
ouh oop, never posted

heres whats displayed now when a write error occurs
image

0xf0xx0 (approved):
acked

Image

0xf0xx0:
i think the update page got overwritten in the merge x3 unless we're keeping it the same?

0xf0xx0:
oh i seeeee, it doesnt load the webui from the unified image when updating from the split update system, you have to do a usb flash

0xf0xx0:
wrong pr oops~

mutatrum:
> oh i seeeee, it doesnt load the webui from the unified image when updating from the split update system, you have to do a usb flash

That shouldn't, OTA should only need `esp-miner.bin`. I'll verify as well,

0xf0xx0:
time to trigger a million billion conflicts!

PR #1764, opened 2026-06-15
Configurable self test warmup temp

mutatrum:
Some devices with very efficient heatsinks never reach 55 °C so self-test doesn't continue. This PR makes it possible to comfigure the self-test temperature values through the config file.


WantClue (approved):
lgtm, we need to cherry pick this into the 2.14.x branch

PR #1766, opened 2026-06-15
Fix heap fragmentation crash

mutatrum:
This PR moves 2 sources of internal memory fragmentation into PSRAM: NVS configuration values string duplication and cJSON buffers. They both are used a lot in hot paths (REST API, stratum JSON decoding). By moving these to the much larger PSRAM fragmentation is prevented that presumably at some point kills the http server as it cannot allocate a single contiguous buffer for JSON encoding.

It also adds max contiguous heap information to system page.

Before / After:

image
image


WantClue (approved):
lgtm let's get this in for testing further.

mutatrum:
> lgtm let's get this in for testing further.

The fix is not in the PR yet, this is only exposing metrics for pre-fix verification.

mutatrum:
> lgtm let's get this in for testing further.

Re-review please :-)

0xf0xx0:
oops- my fault, github cli merged here and not in my local repo :\ testing now

main/main.c
```diff
@@ -31,8 +32,26 @@ static GlobalState GLOBAL_STATE;
 
 static const char * TAG = "bitaxe";
 
+static void *cjson_malloc_psram(size_t size)
+{
+    if (esp_psram_is_initialized()) {
+        return heap_caps_malloc(size, MALLOC_CAP_SPIRAM);
```
WantClue, line 38: might never happen but if PSRAM init happenes but the heap caps malloc fail this will return NULL. in utils.c we fall back to strdup when spiram alloc fails. this can cause fuck ups if the psram is exhausted but i have hard times believing that this might ever happen

PR #1771, opened 2026-06-17
Send SV2 frames in a single write

warioishere:
SV2 frames were sent as two writes — the encrypted header, then the encrypted payload — so every share submit left the device as two TCP segments. This builds both parts into one buffer and sends them in a single write, so a frame goes out as one segment.

On my hardware this drops the SV2 submit response time from ~8.1 ms to ~5.4 ms, on top of the TCP_NODELAY fix — so the two-write pattern still had measurable overhead even with Nagle disabled.

It also removes the root cause that #1722 (TCP_NODELAY) worked around: that latency came from the second segment being held back by Nagle until the first was ACKed. With the whole frame sent as a single segment there is no held-back second segment, so TCP_NODELAY is no longer needed as a fix here. It does not hurt to keep it in either — still sensible for a latency-sensitive socket — so this PR leaves it untouched.

The bytes on the wire are identical (header and payload use separate Noise nonces but are just consecutive bytes); the receiver reads the 22-byte header first and then the payload, so it is unaffected. No pool-side change needed.

Tested on hardware against an SV2 pool.

### DIFF
--- a/components/stratum_v2/sv2_noise.c
+++ b/components/stratum_v2/sv2_noise.c
@@ -501,36 +501,44 @@ int sv2_noise_send(sv2_noise_ctx_t *ctx, esp_transport_handle_t transport,
         return -1;
     }
 
-    // Encrypt header (6 bytes -> 22 bytes)
-    uint8_t enc_hdr[22];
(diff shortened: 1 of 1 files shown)


0xf0xx0 (approved):
acked

PR #1774, opened 2026-06-18
add preliminary bitaxeNaja support with the BM1340

skot:
This starts mining on the [bitaxeNaja-1200x](https://github.com/bitaxeorg/bitaxeNaja) hardware, but it's not the hashrate I would expect from the BM1373. It could be something to do with the BM1340 chips used. I will attempt to update the hardware with actual BM1373 chips and try again.


skot:
I syncronized the register writes in BM1340.c with what's happening in the quad BM1373 dump that we got. I'm still only seeing about half the expected hashrate I expect on the bitaxeNaja 1200x with the BM1340 ASICs.

skot:
This preliminary BM1340-based Naja work has been superseded by the current Naja Duo 1201 series:

- #1890 — BM1372/BM1373 driver support
- #1891 — factory self-test
- #1892 — Naja Duo mining support
- #1893 — LCD interface

Closing this draft in favor of that series.

PR #1778, opened 2026-06-20
Fix work received notification with SV2

mutatrum:
The work received notification on the dashboard (the green down arrow on the Block Header card) was triggered when the scriptSig changes. This never happens on SV2 Standard Channel, and only occasionally on SV2 Extended Channel. This changes the trigger to use the work received counter, similar to how the notification on the display works.

image

It also splits out the share accepted and rejected notification into a green up arrow and a red cross.


bonifacio123:
Tested this on a 601 against Blitzpool. Didn't do an extended vs standard comparison, just extended. But the down arrow only flashes when the output updates. Hope this is helpful.

0xf0xx0 (approved):
tested ack, pls rebase

PR #1779, opened 2026-06-21
fix: keep fractional SV2 pool difficulty to avoid difficulty-too-low rejects

warioishere:
In SV2 the pool derives the channel target from the miner's `nominal_hash_rate`, so fractional pool difficulties (e.g. `931.1`) are normal. `sv2_target_to_pdiff()` computed it as a `double` but returned `uint32_t`, truncating `931.1` → `931`. The submit threshold (`nonce_diff >= active_job->pool_diff`) then used `931`, so shares in `[931, 931.1)` were submitted and rejected by the pool as `difficulty-too-low`:

```
asic_result: ... Nonce 42B7858F diff 931.1 of 931.
stratum_v2_task: Share rejected: difficulty-too-low
```

Return the difficulty as a `double` to keep full precision through to the submit threshold. `GLOBAL_STATE->pool_difficulty` is already a `double`; only the two callers and the log format (`%lu` → `%g`) change. No valid shares are lost — the avoided ones were below the pool's real target and were being rejected anyway.

### DIFF
--- a/components/stratum_v2/sv2_protocol.c
+++ b/components/stratum_v2/sv2_protocol.c
@@ -545,12 +545,14 @@ void sv2_ext_job_free(sv2_ext_job_t *job)
 
 // --- Helpers ---
 
-uint32_t sv2_target_to_pdiff(const uint8_t target[32])
+double sv2_target_to_pdiff(const uint8_t target[32])
 {
     double target_d = le256todouble(target);
-    if (target_d == 0.0) return UINT32_MAX;
+    if (target_d == 0.0) return (double)UINT32_MAX;
     double pdiff = truediffone / target_d;
-    if (pdiff > (double)UINT32_MAX) return UINT32_MAX;
-    if (pdiff < 1.0) return 1;
-    return (uint32_t)pdiff;
+    // Keep full precision: pools may use fractional difficulty (e.g. 931.1).
+    // Truncating to an integer here makes the submit threshold too low, so
+    // shares in [floor(pdiff), pdiff) get sent and rejected as difficulty-too-low.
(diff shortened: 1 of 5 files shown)


mutatrum:
Related: #1594

components/stratum_v2/sv2_protocol.c
```diff
@@ -545,12 +545,14 @@ void sv2_ext_job_free(sv2_ext_job_t *job)
 
 // --- Helpers ---
 
-uint32_t sv2_target_to_pdiff(const uint8_t target[32])
+double sv2_target_to_pdiff(const uint8_t target[32])
```
mutatrum: Can this be moved to `mining.c` as the similar (SV1) code is inside `test_nonce_value`.

warioishere: Makes sense, will move it next to `test_nonce_value`. Keep the name `sv2_target_to_pdiff`, or rename to `target_to_pdiff` since the target→diff math is generic?

mutatrum: It's the same code for sv1 and sv2, so `hash_to_pdiff(uint8_t target[32])` or something like that, which can also be called from test_nonce_value.

warioishere: Done in 81336d8 — extracted `hash_to_pdiff(const uint8_t[32])` into `mining.c`, and `test_nonce_value` now calls it too. Also dropped the SV2-only `<1.0` clamp (a uint32 artifact; `test_nonce_value` never clamped).

mutatrum (approved):

0xf0xx0 (approved):

PR #1783, opened 2026-06-22
Increase SV2 frame size to 8kb

mutatrum:
Allocate in PSRAM instead of stack.

Should fix this issue:

image

### DIFF
--- a/main/tasks/stratum_v2_task.c
+++ b/main/tasks/stratum_v2_task.c
@@ -18,13 +18,14 @@
 #include "device_config.h"
 #include "coinbase_decoder.h"
 #include "esp_heap_caps.h"
+#include "esp_psram.h"
 
 #include <string.h>
 #include <stdlib.h>
 
 #define MAX_RETRY_ATTEMPTS 3
 #define TRANSPORT_TIMEOUT_MS 5000
-#define SV2_MAX_FRAME_SIZE 2048
+#define SV2_MAX_FRAME_SIZE 8192
 
 static const char *TAG = "stratum_v2_task";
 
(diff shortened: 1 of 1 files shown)


WantClue (approved):
code lgtm

main/tasks/stratum_v2_task.c
```diff
@@ -18,13 +18,14 @@
...
 #include "esp_heap_caps.h"
+#include "esp_psram.h"
 
 #include <string.h>
 #include <stdlib.h>
 
 #define MAX_RETRY_ATTEMPTS 3
 #define TRANSPORT_TIMEOUT_MS 5000
-#define SV2_MAX_FRAME_SIZE 2048
+#define SV2_MAX_FRAME_SIZE 8192
```
GitGab19, line 28: Where is this 8192 coming from?

WantClue: The payload you received is too large so this utilizes the PSRAM and increases the allocation

mutatrum: Nothing specific, just 2kb is obviously too small, and we have no counter-example of needing more than 8kb. Rough calculations say that should be enough for a few dozen to maybe 100 outputs, so unless someone starts an SV2 pool with 100s of payout addresses, or other crazy stuff in the coinbase transaction it should be fine.

But if you have a better suggestion, feel free to suggest.

GitGab19: I see.

I was asking the rationale behind that number because from the Sv2 specs the max size of a noise encrypted frame is 65535 bytes (see https://github.com/stratum-mining/sv2-spec/blob/main/04-Protocol-Security.md#46-encrypted-stratum-message-framing).

GitGab19:
I built the `esp-miner.bin`, flashed it on my Bitaxe Gamma 601 (BM1370), pointed it at `stratum2+tcp://public-pool.io:23331public-pool.io:23331/9c4zpyJ2ndm4e8sP2uNc1VNCGxYjqaxWS6wUCjk8zFj6njFquH6` (PPLNS mode), and it's now working.

main/tasks/stratum_v2_task.c
mutatrum: Good point, did some digging. The 65535 is the absolute maximum size a Noise protocol message, not a SV2 frame,  they're separate limits. Basically we're capping the SV2 `NewExtendedMiningJob` to around 100-something outputs. I don't even know if the ESP can handle that amount, so pushing this even further without good reason is opening it to more risks I guess?

GitGab19: Yeah that's correct, but if the `SV2_MAX_FRAME_SIZE` constant is used as the max size of a an ENCRYPTED Sv2 frame, the theoretical max limit is the max size of a noise frame, 65535 bytes.

But I get your point, I just wanted to point this out so that you're aware of that.

0xf0xx0 (approved):

PR #1790, opened 2026-06-27
fix: clamp overheat throttle to the regulator's minimum voltage (no more false PSU fault)

warioishere:
Ran into this the annoying way and figured it's worth fixing for everyone.

I always  undervolt my Gammas and run it at 1080mV. Even undervoltet on this hot day today around 40°c here it overheated a few times, and the overheat protection did its job, dropped voltage and frequency by 100 each and saved them. So my 1080mV quietly turned into 980mV in NVS. Actually I didn't check this explicitly because I ran arround changing PSUs and already thought that my little bee died because of high temp because the error message about the PSU didn't go away.

but the hing is, the TPS546's `VOUT_MIN` is 1.0V. Below that, `TPS546_set_vout()` rejects the value as out of range, the regulator never comes up clean, and the firmware flags a power fault → the UI shouts **"PSU fault, check your PSU"**. And since 980mV got written to NVS, it sticks across reboots. I swapped the PSU, power-cycled a bunch of times — nothing. The only thing that fixed it was going into settings and bumping the voltage back above 1000mV, at which point the fault vanished instantly.

So it's not the PSU at all. It's the throttle undervolting below what the regulator can actually deliver, and then persisting it.

I've already had several emails about this exact "PSU fault", people convinced their supply died, a couple of them bought a new one, and of course it still didn't help. Hence the fix.

**What it does:** clamps the throttle's reduced voltage to the regulator's minimum so it can never command something `set_vout()` will reject. Device-aware (`VOUT_MIN / voltage_domains` → 1000mV on Gamma, ~834mV on Hex), no magic number. Frequency reduction still happens exactly as before, so cooling behaviour is unchanged — it just won't push the voltage into the invalid range anymore.

### DIFF
--- a/main/power/vcore.c
+++ b/main/power/vcore.c
@@ -146,6 +146,20 @@ int16_t VCORE_get_voltage_mv(GlobalState * GLOBAL_STATE)
     return ADC_get_vcore();
 }
 
+// Lowest core voltage (mV) the regulator will accept. TPS546_set_vout() rejects
+// anything below VOUT_MIN as out of range, so callers must not command below this.
+int16_t VCORE_get_voltage_min_mv(GlobalState * GLOBAL_STATE)
+{
+    if (GLOBAL_STATE->DEVICE_CONFIG.TPS546) {
+        TPS546_CONFIG config = get_tps546_config(&GLOBAL_STATE->DEVICE_CONFIG.family);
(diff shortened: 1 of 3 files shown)


0xf0xx0 (approved):
code ack

WantClue (approved):

PR #1795, opened 2026-06-30
More pools

mutatrum:
With SV2 having an Authoroty Pubkey, it's way more difficult to switch pools, just remembering the stratum URL and maybe a non-default port is not sufficient anymore. Therefore: pool profiles:

image

I'm not super happy with the code, it puts each pool config in a JSON in NVS. That's fairly clean, but the migration code is quite a lot. Updating a pool config through the API is now with a complete JSON, instead of per field. For most use-cases I think this is not a big issue, as with a maximum of 8 possible pool configurations there should ample space to put them all in.

A long-term roadmap for possible future pool/stratum PRs:
 * Change pool settings without restarts. Most of the code is already there with the pool coordinator task;
 * Priority queue instead of primary/fallback. Same here, the pool coordinator helps a lot with this;
 * Multi-pool mining with a ratio per pool.


0xf0xx0:
minor nit, should have a `X/8` subtext next to the header imo 
image


also bigger nit, will the buttons be rounded ui-wide? i kinda like it but it should be consistent x3

0xf0xx0:
another nit, tho this might be better addressed in its own pr, the password field should be allowed to be empty

mutatrum:
> minor nit, should have a `X/8` subtext next to the header imo 

Can you explain, not sure what you mean here?

> also bigger nit, will the buttons be rounded ui-wide? i kinda like it but it should be consistent x3

No, I'll pull them in line.

0xf0xx0:
> Can you explain, not sure what you mean here?

see my screenshot, i added the pool count next to the "pool configurations" header. i think itd be nice for users to know how many can be stored uwu

mutatrum:
3x done:
 * No rounded buttons;
 * Empty password allowed;
 * Configuration counter added.

0xf0xx0 (approved):

0xf0xx0 (approved):
re-cked, will merge later

PR #1796, opened 2026-07-01
Add SV2 "require authentication" per-pool option

### LINKED ISSUE #1786: SV2 connects unauthenticated by default, and the cert validity window is parsed but never enforced
**Describe the bug**

SV2 server authentication is implemented but neither enforced nor complete. The Noise_NX handshake does a real Schnorr/BIP-340 check of the pool's certificate against `stratumV2AuthorityPubkey` and fails closed on a bad signature (#1553), but two gaps leave that protection unused in practice. Both are in `components/stratum_v2/sv2_noise.c`:

1. Validity window parsed but never enforced. Step 14 decodes and logs `valid_from` / `not_valid_after` but never compares them to the current time, so an expired or replayed cert passes as long as the signature verifies. Root cause: the firmware has no clock, the lone `esp_sntp.h` include (in `stratum_v1_task.c`) is unused and nothing reads wall-clock time. Not SV2-specific either: without a time source SV1 TLS almost certainly skips cert dates too. Only the signature/chain half of cert validation works today; the expiry half is skipped everywhere.

2. No way to require authentication. With no authority key set (the default), the handshake takes the "Skipping certificate verification (no authority pubkey)" branch and connects anyway, so SV2 is encrypted-but-unauthenticated out of the box. The Authority Pubkey tooltip already warns that without the key "a man-in-the-middle attack could redirect your hashrate," but nothing lets the user act on it.

**To Reproduce**

Unauthenticated by default:
1. Pool Settings > Advanced Options > Stratum Protocol = SV2.
2. Leave SV2 Authority Pubkey blank, save, restart.
3. Connect to any SV2 pool; serial log shows "Skipping certificate verification (no authority pubkey)" and it mines, with no way to refuse.

Validity not enforced:
1. Set SV2 with a valid Authority Pubkey against a pool whose cert `not_valid_after` is in the past (or replay an old, still-validly-signed cert).
2. Connect; log shows "Server certificate verified OK" and mining proceeds.
3. Expected the expired/stale cert to be rejected.

**Expected behavior**

- A "require authentication" option that refuses to connect when no key is set or verification can't complete, instead of silently connecting unauthenticated.
- The signed validity window enforced (`valid_from <= now <= not_valid_after`) once a clock exists.

**Additional context**

Fixes, in priority order:

- Enforcement toggle (small): per-pool "require authentication" NVS bool + a checkbox in the existing SV2 options block, next to the Authority Pubkey field. Verify path unchanged; it just refuses to fall through when verification can't happen. Lets the user act on the MITM warning the UI already shows.

- Validity check + time source: add a coarse SNTP sync at boot (network's already up, ESP-IDF's `esp_netif_sntp` is a few lines), then enforce the window. The accuracy bar is low, cert windows are wide, so one boot-time fetch is plenty, no RTC needed, with a "skip if time unknown" path for the pre-sync window. Honest limit: NTP is unauthenticated, so an on-path attacker who spoofs time can defeat expiry anyway, this is defense-in-depth (honest expiry, misconfig, passive replay), not airtight against a determined MITM. Bonus: the same clock lets SV1 TLS enforce cert dates too.

- TOFU pinning (optional): for pools without a published key, pin the server static key on first connect and warn if it changes. Lower priority.

Happy to PR the enforcement toggle. I run NerdQAxe++ against my own SV2 pool, so I can test the server side (valid/expired/wrong-key/no-key certs), but would want a maintainer with a Bitaxe to confirm the on-device flash.

cbyam:
## Summary

Addresses the second gap in #1786. By default SV2 connects encrypted but
unauthenticated: with no authority pubkey set, the Noise_NX handshake takes
the "no authority pubkey" branch and mines anyway, so the user can't act on
the man-in-the-middle warning the UI shows on the Authority Pubkey field.

This adds a per-pool "Require Authentication" toggle. When it's on and no
authority pubkey is set, the SV2 task refuses to connect instead of falling
through to an unverifiable server. The verify path is unchanged.

## Changes

- New per-pool NVS bool `stratumV2RequireAuth` / `fallbackStratumV2RequireAuth`,
  default false, so existing configs behave as before.
- `stratum_v2_task` refuses to connect when auth is required but no pubkey
  resolves (status shows `SV2: Auth required - no key`).
- REST config exposes the new booleans; `openapi.yaml` updated.
- Frontend checkbox next to the Authority Pubkey field, with a tooltip pointing
  back to the MITM warning.

## Scope

Just the enforcement toggle, the small fix offered in #1786. The validity-window
check and boot-time SNTP clock from that issue are not here.

## Testing

- AxeOS production build passes (`ng build --configuration=production`).
- Each changed C file passes a syntax and semantic compile under ESP-IDF v5.4,
  and `sv2_noise.c` compiles to an object in a full build. A complete link needs
  v5.5.3 (the CI toolchain) because of a pre-existing `esp_transport_get_socket`
  call elsewhere in the tree, so the final .bin is left to CI. Not flashed to
  hardware yet.


0xf0xx0:
code ack, will test soon

anything against having this be the default path for remote conns instead of a toggle? im still learning how sv2 security works, i was under the impression the pubkey/encryption is required for remote servers and optional for the lan. kinda averse to adding another button to the settings as well

cbyam:
you've basically got it right on the split, small correction: it's not encryption that's optional on LAN, that part's always on (Noise_NX establishes the session unconditionally). what's optional/mandatory is specifically the auth step - verifying the server cert against the authority pubkey. that's the "optional LAN, mandatory remote" language.

on defaulting instead of a toggle for remote - genuine limitation, not a preference: the firmware has no way to tell LAN from remote, it's operator config, not topology. so it can't switch on automatically by connection type, has to be a declared setting.

and the toggle isn't a duplicate of the pubkey field, even though it looks like it should be. empty key already skips verification and mines anyway - that's true today, no PR needed for that part. what #1786 flagged, and what this actually fixes, is the silent case: if someone thinks they're running verified but the key field is blank or got cleared, there's no signal - it just quietly falls back to unauthenticated. "no key" and "no key when I said I want this verified" need to be distinguishable, and the pubkey field alone can't do that.

so yeah, one more control, but it's closing a real footgun, not adding a redundant one.

main/http_server/axe-os/src/app/components/pool/pool.component.html
```diff
@@ -142,6 +141,21 @@ <h2 *ngIf="pool === 'fallbackStratum'" class="mt-5">Fallback Pool Configuration<
...
+                            <div class="field-checkbox grid">
+                                <div class="col-1 md:col-10 md:flex-order-2">
+                                    <p-checkbox [name]="pool === 'stratum' ? 'stratumV2RequireAuth' : 'fallbackStratumV2RequireAuth'"
+                                        [inputId]="pool === 'stratum' ? 'stratumV2RequireAuth' : 'fallbackStratumV2RequireAuth'"
+                                        [formControlName]="pool === 'stratum' ? 'stratumV2RequireAuth' : 'fallbackStratumV2RequireAuth'"
+                                        [binary]="true"></p-checkbox>
+                                </div>
+                                <label [htmlFor]="pool === 'stratum' ? 'stratumV2RequireAuth' : 'fallbackStratumV2RequireAuth'" class="col-11 m-0 pb-0 pl-3 md:col-2 md:flex-order-1 md:pl-2">
+                                    <tooltip-text-icon
+                                        text="Require Authentication"
```
0xf0xx0: mayb add a forced linebreak here, "authentication" overlaps the button slightly on my screen

0xf0xx0 (changes requested):
one nit, otherwise acked

main/http_server/axe-os/src/app/components/pool/pool.component.html
cbyam: fixed in 20de1ef. the overlap came from tooltip-text-icon keeping the last word + icon in a nowrap span, so "Authentication" could not wrap in the narrow desktop label column. now it breaks after "Require" on desktop widths only - mobile keeps the single line since the label is full-width there.

0xf0xx0 (approved):
i forgot to reack

0xf0xx0:
pls rebase uwu ill merge this soon

cbyam:
rebased onto master. heads up that the More Pools refactor (#1795) replaced the flat stratum NVS keys with per-pool JSON, so this ended up re-implemented on the pool model rather than textually rebased: sv2_require_auth lives on PoolConfig now, parsed from stratumV2RequireAuth in the pool JSON and persisted through the same REST path as the other pool fields. The old standalone NVS keys are gone, and since they never shipped there's nothing to migrate. The checkbox sits in each pool card under the Authority Pubkey field. Enforcement in the SV2 task is unchanged: require + no usable key = refuse to connect. Built firmware on IDF 6.0.2 and the web UI locally, checked the layout at desktop and phone widths.

mutatrum (approved):

PR #1799, opened 2026-07-03
fix: reduce sv2 minimum extranonce size from 6->2 bytes

### LINKED ISSUE #1780: reduce sv2 minimum extranonce size to match sv1
The bitaxe on sv2 requests a minimum extranonce size of 6 bytes, while the sv1 side has no minimum. should we set the minimum at 0 and rely on the sv2 pool to provide enough nonce space for its job interval?

0xf0xx0:
closes: #1780

should this be extracted into a constant, and where? we could also use this value to create the `version-rolling.min-bit-count` value for sv1 version rolling. i run a 2-byte extranonce myself, havent had any issues so im hoping this doesnt introduce any.

E: also, this hard minimum should be documented somewhere...


mutatrum (approved):
Makes sense

PR #1802, opened 2026-07-04
Fix websocket stale client cleanup

frickl:
## Summary

This change improves WebSocket connection handling by removing stale clients
when `httpd_ws_send_frame_async()` fails.

Additionally, `websocket_broadcast()` now copies the client list while holding
the mutex and performs the actual sends after releasing the mutex.

## Problem

Under repeated WebSocket reconnects (built-in UI and external monitoring
clients), stale client entries could accumulate after failed async sends.

Over time this could exhaust `MAX_WEBSOCKET_CLIENTS`, resulting in:

- `429 Too Many Requests`
- `httpd_accept_conn: error in accept (113)`
- lost WebSocket connections
- unavailable log/graph pages

## Solution

- remove stale WebSocket clients after async send failures
- trigger HTTP session cleanup
- avoid broadcasting while holding the client mutex

## Testing

Tested on a Bitaxe Gamma (BM1370, Board 601) running v2.14.1.

The miner was operated for several hours with both the built-in web UI and
additional external WebSocket monitoring clients.

After the change no further exhaustion of `MAX_WEBSOCKET_CLIENTS` was observed.


mutatrum:
Please clean up the PR, there are some other commits merged into this as well.

frickl:
This PR was accidentally created from the `v2.14.x` branch instead of `master`, which caused unrelated commits to appear in the diff.

I've opened a clean PR based on `master` here:

#1803

Closing this PR to avoid confusion.

mutatrum:
Thanks. For the next time you don't have to close and open a new PR, you can just force push a complete new branch state, GitHub will pick that up.

PR #1798, opened 2026-07-02
Add 601a device config with TPS546 and nominal_voltage=12

awxhw:
This PR adds complete firmware support for StarMiner bitaxeGamma601a.

1. Updated board config, power and thermal settings for V2 hardware revision.
2. Code follows the existing project structure without breaking other devices.
3. Full stability mining test finished, all core functions operate normally.
4. I will maintain this hardware port long-term and handle subsequent compatibility updates.
Hardware repo: [https://github.com/awxhw/StarMiner/tree/master/sch](https://github.com/awxhw/StarMiner/tree/master/sch).

PR #1804, opened 2026-07-05
Add GammaV2/601a board support (12V input variant)

tsoares37:
Adds support for the 601a board variant, which is electrically identical to the standard Gamma (BM1370 + TPS546D24A) except for a 12V input rail instead of 5V. This requires different TPS546 input voltage lockout thresholds (VIN_ON/VIN_OFF/0V_FAULT).

This mirrors the GammaV2 implementation already published by StarMiner under GPL-3.0 at github.com/awxhw/StarMiner, ported here without their additional telemetry/BLE code, so this hardware variant can run mainline, privacy-respecting firmware.

Note: the 601a board is a third-party hardware variant (not sold by bitaxeorg) that swaps the input rail from 5V to 12V. This PR only adds firmware support for that voltage difference.

Tested on physical hardware.


WantClue:
thats a duplicate of #1798

PR #1805, opened 2026-07-08
fix: adjust freq to also fallback to default on self test

WantClue:


main/tasks/power_management_task.c
```diff
@@ -230,7 +230,9 @@ void POWER_MANAGEMENT_task(void * pvParameters)
         uint16_t core_voltage = GLOBAL_STATE->SELF_TEST_MODULE.is_active
                                  ? GLOBAL_STATE->DEVICE_CONFIG.family.asic.default_voltage_mv
                                  : nvs_config_get_u16(NVS_CONFIG_ASIC_VOLTAGE);
-        float asic_frequency = nvs_config_get_float(NVS_CONFIG_ASIC_FREQUENCY);
+        float asic_frequency = GLOBAL_STATE->SELF_TEST_MODULE.is_active
+                                 ? GLOBAL_STATE-> DEVICE_CONFIG.family.asic.default_frequency_mhz
+                                 : nvs_config_get_u16(NVS_CONFIG_ASIC_FREQUENCY);
```
0xf0xx0: float or u16? x3

WantClue: ah shit lol yes

0xf0xx0 (approved):

mutatrum:
I don't think this is correct. A manufacturer is free to change the initial frequency, f.e. when delivered with a better heatsink + powersupply. A cleaner solution would be to put `default_freq` and `default_voltage` in NVS as a config only setting and read that before the hardcoded value. That way the self-test will run with the frequency and voltage the device is shipped.

PR #1807, opened 2026-07-10
Add GammaHex board support

benjamin-wilson:
## Summary
- add GammaHex board/config support, large ST7789 display assets, and TPS546/VCORE handling
- add family-level default frequency/voltage so self-test uses board CSV defaults, including GammaHex 690 MHz / 1200 mV
- fix settings save path for Axe-OS fan controls and align manual/min fan setting names
- remove accidental local factory binary from the branch

## Testing
- git diff --check
- npx ng build --configuration=production
- idf.py -B /tmp/esp-idf-codex-build build


components/asic/bm1370.c
```diff
@@ -231,28 +258,33 @@ uint8_t BM1370_init(void * pvParameters)
...
-    //Core Register Control
-    //TX: 55 AA 51 09 [00 3C 80 00 80 0C] 11  //command all chips, write chip address 00, register 3C, data 80 00 80 0C - Core Register Control
-    _send_BM1370((TYPE_CMD | GROUP_ALL | CMD_WRITE), (uint8_t[]){0x00, 0x3C, 0x80, 0x00, 0x80, 0x0C}, 6, BM1370_SERIALTX_DEBUG); //from S21Pro dump
-    //_send_BM1370((TYPE_CMD | GROUP_ALL | CMD_WRITE), (uint8_t[]){0x00, 0x3C, 0x80, 0x00, 0x80, 0x18}, 6, BM1370_SERIALTX_DEBUG); //from S21 dump
+    // Core Register Control, internal ClockDelayCtrlV2.
+    BM1370_write_clock_delay_ctrl(0x00, (TYPE_CMD | GROUP_ALL | CMD_WRITE));
+    BM1370_init_core_register_delay();
+
+    // Present in the S21 XP init between core register control and analog mux.
+    _send_BM1370((TYPE_CMD | GROUP_ALL | CMD_WRITE), (uint8_t[]){0x00, 0x14, 0x00, 0x00, 0x00, 0xFF}, 6, BM1370_SERIALTX_DEBUG);
```
mutatrum, line 268: Why not do the delay inside `_send_BM1370`?

main/bap/bap.c
```diff
@@ -27,6 +28,11 @@ esp_err_t BAP_init(GlobalState *state) {
         ESP_LOGE(TAG, "Invalid global state pointer");
         return ESP_ERR_INVALID_ARG;
     }
+
+    if (state->DEVICE_CONFIG.family.id == GAMMA_610 || state->DEVICE_CONFIG.family.id == GAMMA_HEX) {
```
mutatrum: It would be cleaner if there is a `disable_bap` bool flag in device_config. This would be `false` by default so it only needs to be set to true on this model.

mutatrum:
Not sure what to make of this, it's quite a code-dump. Some questions:
 * Why not create a larger Bitaxe logo for this resolution? It's a one-off thing, unless we're planning more displays
 * LVGL can read `.png` directly, so that should be easier for the background. Also, why are there 2 background images?
 * As this is a completely different display device class, we might need to separate out the code a bit more. There's quite a bit of spaghetti going on here, code wise :-)  
 * As this is a different display connection, I don't think it should be in the dropdown as one can't just switch from OLED to ST7789. I assume the 610 doesn't have the OLED at all?
 * Does the LVGL config changes also impact the OLED? Not sure if moving to 16-bit color depth has a performance impact there. Also not sure how one would make this configurable, if needed?

components/asic/bm1370.c
benjamin-wilson: _send_BM1370() is also used for normal mining jobs and register reads, so adding an unconditional 10 ms delay there would throttle runtime traffic. The delay is only needed between specific initialization register writes.

benjamin-wilson:
The C representation avoids enabling a PNG decoder and doing runtime decompression.
The OLED path still explicitly uses LV_COLOR_FORMAT_I1, monochrome = true

benjamin-wilson:
I think I fixed most of your concerns @mutatrum, ready for another review.

components/asic/bm1370.c
WantClue: this delay needs to be tested also on single asic chips, will do that

config-801.cvs
```diff
@@ -27,5 +27,5 @@ boardversion,data,string,801
 rotation,data,u16,0
 autofanspeed,data,u16,1
 fanspeed,data,u16,100
-selftest,data,u16,1
+selftest,data,u16,0
```
WantClue: why

WantClue:
@benjamin-wilson please resolve the conflicts

benjamin-wilson:
Closing this superseded cumulative branch. I am rebuilding Gamma Hex as a clean, conflict-resolved change set against current master and the new Naja/shared hardware PR series.

benjamin-wilson:
Replacement: #1894. The new branch is rebuilt on current master and the Naja/shared hardware PR stack, with the duplicated LCD assets and board-plumbing changes removed from its Gamma-specific commit.

PR #1704, opened 2026-05-25
Add BM1368/BM1370 predictive efficiency telemetry and autonomous tuning

0xjc65eth:
## Summary

This PR proposes a BM1368/BM1370-focused predictive efficiency layer for ESP-Miner/AxeOS.

It adds:

- Passive predictive-efficiency telemetry exposed under `/api/system/info` as `predictiveEfficiency`.
- BM1368 and BM1370 profile detection for Supra, SupraHex, Gamma, GammaDuo, and GammaTurbo-style devices.
- Composite efficiency scoring based on hash-per-watt, accepted share ratio, thermal margin, expected hashrate ratio, ASIC error pressure, and pool latency pressure.
- A conservative autonomous frequency autotune loop behind the `predictiveAutotune` NVS/API setting.
- A standalone optional Ollama companion tool that runs outside the ESP32 and can observe AxeOS telemetry, keep local memory, perform research, and suggest/apply settings through the existing AxeOS API.

## Important boundary

This PR does **not** attempt to predict winning Bitcoin nonces and does not claim to break or shortcut SHA-256 proof-of-work.

The intent is operational optimization only: better visibility into miner efficiency, safer frequency exploration, thermal rollback behavior, and external tooling for supervised experimentation.

## Firmware behavior

The firmware-side autotune is intentionally narrow:

- It only acts on supported BM1368/BM1370 ASIC profiles.
- It only changes ASIC frequency.
- It uses small 25 MHz exploration steps.
- It includes warmup, monitor, trial, rollback, and throttle states.
- It rolls back when score drops, ASIC errors rise, rejected shares rise, or thermal margin is exhausted.
- It does not change voltage, pool settings, fan behavior, or job scheduling.

## API additions

`/api/system/info` now includes a `predictiveEfficiency` object with fields such as:

- `enabled`
- `autotuneEnabled`
- `bm1368Profile`
- `bm1370Profile`
- `profileName`
- `score`
- `hashPerWatt`
- `usefulShareRatio`
- `thermalMarginC`
- `hashrateRatio`
- `errorPenalty`
- `latencyPenalty`
- `recommendedAction`
- `agentState`
- `agentReason`
- `tunedFrequency`
- `bestFrequency`
- `trialFrequency`
- `baselineScore`
- `lastTuneMs`

## External companion tool

The Ollama research companion is deliberately external to the ESP32. It is intended for advanced users and supervised test setups. It uses the existing AxeOS API rather than embedding internet access or LLM behavior into the firmware runtime.

[...]

PR #1810, opened 2026-07-11
Auto-Tune — automatic frequency/voltage tuning with Eco/Balanced/Aggressive profiles

iitssjstn:
## What this adds
A new background task (`autotune_task.c`) that automatically finds the
highest stable frequency for the installed ASIC, then shaves voltage
down at that frequency for efficiency — all without manual tuning.

Three selectable profiles let the person choose how aggressively it
tunes:

| | Eco | Balanced | Aggressive |
|---|---|---|---|
| Thermal margin below throttle cutoff | 10°C | 5°C | 2°C |
| Reject-rate tolerance | 2% | 3% | 3% |
| Voltage rescue on a failing step | no | yes | yes |
| Beyond-spec climbing (opt-in, see below) | no | no | yes |

## How it works
- Starts from the bottom of the vendor-tested frequency/voltage table
  every time it's (re-)enabled, and climbs step by step.
- If a frequency step is unstable, it first tries a bit more voltage to
  rescue it (Balanced/Aggressive only) before giving up on that step.
- Reacts to two independent instability signals: pool-level rejected
  shares (with "Duplicate share" rejections excluded, since those are a
  pool/protocol artifact unrelated to stability) and the ASIC's own
  on-chip error-rate register, which is unaffected by network/pool noise.
- Once the top frequency is reached, it keeps shaving voltage down at
  that frequency for as long as it stays stable — same hashrate, less
  heat, less power.
- Runs continuously for as long as it's enabled, so it re-adapts to
  changing ambient temperature automatically.
- If the person changes frequency/voltage manually while it's running,
  it detects that and adopts the new value as its starting point rather
  than overwriting it.

## Optional beyond-spec extension (Aggressive profile only)
With "Custom Settings" unlocked (`?oc`) and the Aggressive profile
selected, it can continue past the vendor-tested table in small 5MHz
steps, up to a hard, non-configurable ceiling — either a
personally-verified value for a specific chip (currently BM1370 →
800MHz, based on my own testing) or a conservative 15% fallback for
other chips. This is clearly opt-in and separately explained in the UI
as an overclocking feature outside vendor-tested limits.

## Safety
- The existing reactive overheat-protection in `power_management_task`
  is untouched and remains the final backstop regardless of profile.
- Voltage is never extended past the vendor table under any
  circumstances, even in beyond-spec mode.
- All safety thresholds are hard-coded, not user-configurable, aside
  from the three profile presets.

## Note
This was developed with AI assistance (Claude), with all logic
iteratively tested and debugged on real hardware based on live device logs.


0xf0xx0:
may i suggest reducing this pr to the relevant commits?

main/http_server/axe-os/src/app/services/system.service.ts
```diff
@@ -332,25 +340,21 @@ export class SystemApiService {
   }
 
   public updateSystem(uri: string = '', update: any): Observable<any | ISystemUpdateResponse> {
-    if (!environment.mock && this.api && !uri) {
```
mutatrum, line 335: This reverts a bugfix (#1809)

main/http_server/axe-os/src/app/services/system.service.ts
```diff
@@ -332,25 +340,21 @@ export class SystemApiService {
...
+    if (environment.mock && this.api && !uri) {
       return from(this.api.invoke(functions.updateSystemSettings, { body: update as Settings }));
     }
 
-    if (!environment.mock && uri) {
```
mutatrum, line 339: Here too

mutatrum:
Some remarks:

I'm not sure what 'Thermal margin below throttle cutoff' means, if that's the target temperature then we might as well just use target temperature here. Then the parameters are the target temp, max error percentage and if it can go above spec or not, which should be locked behind the oc enable switch, IMO.

I'm not sure if stratum errors are any indication of overlocking. What's missing is a threshold on Input Voltage. If that drops too much, the power brick is crapping out and can't sustain the load. This might precede any hashrate error register readings.

I would have expected `Eco` to optimise for efficiency, not highest hash rate at a lower temperature. So, then there's only 2 modes: optimise for efficiency or optimise for hashrate.

Also: take a look (or your agent) at #1152 and #1704. There is quite some prior work on auto-tuning, it might be good to have a proper analysis on the different attempts.

And before continuing, please remove unrelated changes, or separate them out into their own PR.

0xf0xx0:
you can reopen this once its been cleaned up, stacking more commits without discussion isn't very helpful x3

iitssjstn:
> you can reopen this once its been cleaned up, stacking more commits without discussion isn't very helpful x3

My bad for that, when im done with it ill reopen it

PR #1818, opened 2026-07-17
Fix: one invalid field in a settings save blocked all other fields

iitssjstn:
## Problem
`check_settings_and_update()` in `http_server.c` used a single shared
`bool result` flag across all fields in a settings PATCH request. If
any one field failed validation (out of range, stale, unexpected type),
the entire request was rejected and *nothing* was saved -- including
fields the user never touched, since the frontend always submits the
whole settings form on every save.

## Fix
Replaced the single shared flag with a `bool field_ok[NVS_CONFIG_COUNT]`
array tracked per field. Invalid fields are now skipped individually
(with a warning logged) while every other field that validated
correctly is still applied.

## Testing
Verified on real hardware that settings saves now succeed even when an
unrelated field in the same request would previously have blocked
everything.

## Note
This was developed with AI assistance (Claude), with all logic
iteratively tested and debugged on real hardware based on live device
logs

### DIFF
--- a/main/http_server/http_server.c
+++ b/main/http_server/http_server.c
@@ -619,6 +619,17 @@ static esp_err_t handle_options_request(httpd_req_t * req)
 
 bool check_settings_and_update(const cJSON * const root, char **redirect_url)
 {
+    // Track validity PER FIELD rather than one global pass/fail flag. With a
+    // single shared flag, one invalid or unexpected field anywhere in the
+    // request silently blocked saving *everything* else in the same
+    // request -- including fields the person never touched, since the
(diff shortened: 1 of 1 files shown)


WantClue:
remove all the unnecessary comments in the code thats not needed

iitssjstn:
> remove all the unnecessary comments in the code thats not needed

trimmed the comments down to just the essentials in the latest commit.

mutatrum:
This behaviour is nu design. Either an update request succeeds totally, or not at all. Half updates are unpredictable behaviour.

PR #1821, opened 2026-07-19
Add optional password protection (HTTP Basic auth) for the AxeOS web interface

romainwg:
Right now anyone on the same Wi-Fi as a Bitaxe can open AxeOS and read the pool/Wi-Fi details, change any setting, or even flash new firmware — there's no authentication at all. This PR adds an optional password so you can lock that down on a shared or untrusted network.

It's completely opt-in. If you don't set a password nothing changes and the device behaves exactly as it does today, so updating to this is safe even for people who don't want it. Once you do set a password (under Settings → Security), the browser asks you to sign in before you can read data, change settings, or run an OTA update.

How it works

I tried to keep the change small and contained. There's a new `components/http_auth/` component that stores and checks the credential, and a single gate function (`http_auth_gate()`) that hooks into the existing `is_network_allowed()` check in `http_server.c`, plus the static-file handler so the login prompt appears on first load. That one choke point covers every `/api/*` route, both WebSockets, and OTA when auth is on. The password is only ever stored as a salted SHA-256 hash in its own NVS namespace (`authcfg`), it's never returned by any endpoint, and the comparison is constant-time. If reading the config ever fails, the gate fails closed. There are two new endpoints, `GET` and `POST /api/system/auth`, to report status and to set/change/clear the password (sending an empty password turns auth back off). It's all in `openapi.yaml`, and there's a small Settings → Security panel on the frontend.

Recovery — you can't get locked out or brick the device

This is the part I was most careful about. AP setup mode (hold the BOOT button) already bypasses auth, and I surface the Security panel right on that setup screen, so if you forget the password you just drop into AP mode and clear it from the browser. Erasing NVS (a factory image or `idf.py erase-flash`) also clears it. There are no changes to the partition table, the bootloader, or the flash layout, so a USB reflash always works as a last resort. The setup AP stays unauthenticated on purpose — that's what makes recovery possible.

Testing

I added C unit tests for the auth logic (hashing against known vectors, right/wrong/edge-case passwords, parsing the Basic header including odd and malformed input, the constant-time compare) that run under QEMU, plus Angular specs for the auth service, the Security panel, and the AP-mode reset. The full esp32s3 build and the production web build both pass, and I've been running it on a real Bitaxe 601 Gamma.

Why you might want to merge it

- It closes a real gap: today there's zero protection against anyone on your LAN changing settings or flashing firmware.
- No risk for people who don't want it — it does nothing at all until a password is set.
- It's small and isolated: one new component, one gate function, a couple of endpoints and a settings panel. No flash-layout, partition, or bootloader changes.
- Reasonable crypto choices: salted SHA-256, constant-time compare, the hash never leaves the device, and the gate fails closed.
- Recovery is safe, so it shouldn't generate "I locked myself out" support reports.

Things to weigh (the honest trade-offs)

- It's Basic auth over plain HTTP. It keeps out casual LAN access but doesn't encrypt traffic, so it won't stop someone who can sniff the network or sit on-path — the device has no TLS.
- The setup AP is intentionally left open so recovery works, which means anyone within Wi-Fi range of the device can still get in. Physical/RF access is out of scope here.
- It's a single shared admin credential — no multiple users or roles.
- There's no rate limiting or lockout on failed attempts. Since it's LAN-scoped I judged that acceptable, but it's worth noting.
- The UX is the browser's native Basic-auth dialog rather than a custom login page, and "logging out" means clearing the browser's saved credentials.
- It does add a bit of ongoing surface to maintain (the new component, endpoints and frontend panel).

### DIFF
--- a/components/http_auth/CMakeLists.txt
+++ b/components/http_auth/CMakeLists.txt
@@ -1,0 +1,13 @@
+idf_component_register(
+SRCS
+    "http_auth.c"
+
+INCLUDE_DIRS
+    "include"
+
+REQUIRES
+    "esp_http_server"
+    "mbedtls"
+    "nvs_flash"
+    "esp_hw_support"
+)
--- a/components/http_auth/http_auth.c
+++ b/components/http_auth/http_auth.c
@@ -1,0 +1,416 @@
+#include "http_auth.h"
+
+#include <string.h>
+#include <strings.h>
+#include <stdlib.h>
+
+#include "esp_log.h"
+#include "esp_random.h"
+#include "nvs.h"
+#include "nvs_flash.h"
+#include "mbedtls/sha256.h"
+#include "mbedtls/base64.h"
+#include "freertos/FreeRTOS.h"
+#include "freertos/semphr.h"
+
+static const char *TAG = "http_auth";
+
+// Credentials live in their own NVS namespace so they can never collide with,
+// or be dumped alongside, the regular device settings ("main" namespace).
+#define AUTH_NVS_NAMESPACE "authcfg"
+#define AUTH_NVS_KEY_USER  "user"
+#define AUTH_NVS_KEY_HASH  "hash"
+
+#define AUTH_SALT_LEN      16
+// "<32 hex salt>:<64 hex sha256>" = 97 chars + NUL. Round up for headroom.
+#define AUTH_HASH_STR_MAX  128
+#define AUTH_DEFAULT_USER  "admin"
+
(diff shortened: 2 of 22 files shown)


mutatrum:
Duplicate of #1750, I didn't review the code so not sure if there are differences between the two implementations.

There are some nuanced issues:
 * What to do if the user locks himself out, how can they recover from that
 * Websockets should follow same auth
 * Is read-only access ok, or does it need to be locked down completely
 * Swarm access

PR #1829, opened 2026-07-25
Upgrade to ESP-IDF 6.0.2

eandersson:
This is a rough conversion as they completely migrated to MbedTLS 4.x and adopts the PSA Crypto API.
https://github.com/espressif/esp-idf/releases#release-v6.0

- Updated sha256 / TLS compatibility with 6.x.
- Updated Web Socket handshake.
- Fixed GCC 15.x compiler issues.


WantClue:
SV2 testing is required, if @STSMiner1 could do that would be appreciated, otherwise i will see if i can do a sv2 setup for testing

mutatrum:
~I'm surprised all the SHA256 constants and code is now needed.~

They are needed. A follow-up could be a clean-up of the midstates, as only the BM1397 actually needs and uses them. For all other chips, midstate calculations are done but not used.

bonifacio123:
Saw on Discord the need for testing SV2. Using a DUO I updated the firmware using the unified update method. Have been running against a local Stratum SV2 UI instance for the past three hours. No issues observed so far. Hope this is helpful in some way.

Screenshot From 2026-07-17 09-46-53
Screenshot From 2026-07-28 11-12-28

eandersson:
> I'm surprised all the SHA256 constants and code is now needed.

Same.

> Saw on Discord the need for testing SV2. Using a DUO I updated the firmware using the unified update method. Have been running against a local Stratum SV2 UI instance for the past three hours. No issues observed so far. Hope this is helpful in some way.
> 
> Screenshot From 2026-07-17 09-46-53 Screenshot From 2026-07-28 11-12-28

Appreciate it. Thanks.

WantClue (approved):
code lgtm

WantClue:
@eandersson 
 PSA key imported and destroyed on every Noise frame.
 
noise_encrypt/noise_decrypt do psa_import_key + psa_destroy_key per call, but post-handshake ctx->send_key/ctx->recv_key are fixed for the session (sv2_noise.c:555,572,579,606,635). ESP-IDF 6.0 enables MBEDTLS_THREADING_C by default, so each import/destroy takes the global key-store mutex. Import once after the handshake, cache two psa_key_id_t in sv2_noise_ctx_t, destroy on teardown. hmac_sha256 has the same pattern but is handshake-only

main/system.c
```diff
@@ -477,12 +477,9 @@ void SYSTEM_init_partitions(GlobalState * GLOBAL_STATE) {
 
             esp_app_desc_t app_desc;
             if (esp_ota_get_partition_description(p, &app_desc) == ESP_OK) {
-                strncpy(cp->version, app_desc.version, sizeof(cp->version) - 1);
-                cp->version[sizeof(cp->version) - 1] = '\0';
-                strncpy(cp->compileDate, app_desc.date, sizeof(cp->compileDate) - 1);
-                cp->compileDate[sizeof(cp->compileDate) - 1] = '\0';
-                strncpy(cp->compileTime, app_desc.time, sizeof(cp->compileTime) - 1);
-                cp->compileTime[sizeof(cp->compileTime) - 1] = '\0';
+                snprintf(cp->version, sizeof(cp->version), "%s", app_desc.version);
```
WantClue, line 480: trades a bounded read for an unbounded one.

I'd suggest to use "%.*s" with an explicit expression, that should silence -Wstringop-truncation

PR #1831, opened 2026-07-25
feat: prompt before restart in topbar

### LINKED ISSUE #1633: Prevent accidental reset through web interface - Feature impovement

0xf0xx0:
closes: #1633 

copies the reset prompt from the partition reboot button, i wanted to do a panel but angular is eughhhhh  
future things

### DIFF
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
@@ -371,6 +371,9 @@ private isIpAddress(value: string): boolean {
   }
 
   public postAction(device: any, action: string) {
+    if (action === 'restart' && !confirm('Are you sure you want to restart the device?')) {
+      return;
+    }
     this.httpClient.post(`http://${device.connectionAddress}/api/system/${action}`, {}, { responseType: 'text' }).pipe(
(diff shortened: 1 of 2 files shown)


WantClue:
Does this make the UX exp weird on mobile ?

0xf0xx0:
non, it uses the browser alert box uwu should be fine on all devices

mutatrum:
Works like it says on the tin. Maybe also add the same confirm to the restart button on swarm?

mutatrum (approved):

0xf0xx0:
do we have a central utils module to throw the restart func in? thatd make it easier to use in the swarm page as well

0xf0xx0:
prompt added to swarm, its different enough that it needs its own impl  
needs testing

mutatrum (approved):

0xf0xx0:
will merge soon

PR #1832, opened 2026-07-25
remove luck from hashrate card

0xf0xx0:
image


WantClue (approved):

MisterFixx:
Maybe make this a toggleable option? I'd still like to see luck on my Bitaxe...

0xf0xx0:
it was just your chance to get a block, which is a fraction of a fraction of a percent for us x3 what usecase did it provide? maybe we can find something

PR #1815, opened 2026-07-12
Purge PrimeNG and migrate to Tailwind CSS

mutatrum:
This PR purges PrimeNG and migrates to Tailwind CSS. PrimeNG stopped being open source and requires a license from version 22. We could get a license but are not allowed to ship the firmware with a license embedded, so this makes it impossible to move forward with PrimeNG.

To resolve this predicament, the following changes were made:
 * Remove PrimeNG
 * Add custom components for checkbox, dropdown, slider and tooltips
 * Migrate css variable names to Tailwind CSS standards
 * Make system page and drop downs only update on changing fields

As Tailwind CSS is much more efficient in what to pack, the size of the packages has reduced by ~25%:

Before:

| Initial chunk files   | Names         |  Raw size | Estimated transfer size |
|  --- | --- | ---: | ---: |
| main.js      | main          |   1.55 MB |               343.55 kB |
| styles.css   | styles        | 378.67 kB |                24.91 kB |
| polyfills.js | polyfills     |  34.76 kB |                11.41 kB |
|                       | Initial total |   1.96 MB |               379.87 kB |

After:

| Initial chunk files   | Names         | Raw size | Estimated transfer size |
|  --- | --- | ---: | ---: |
| main.js      | main          |  1.04 MB |               263.22 kB | 
| styles.css   | styles        | 60.53 kB |                11.88 kB | 
| polyfills.js | polyfills     | 34.76 kB |                11.41 kB | 
|                       | Initial total |  1.13 MB |               286.51 kB | 

Follow-up of #1651.
Supersedes #1800 and https://github.com/mutatrum/ESP-Miner/pull/10

To keep the scope slightly smaller *ahem*, this PR does _not_ upgrade Angular,, does _not_ replace toastr with a native component nor replaces moment.js with `Intl`. As always with these changes, `package-lock.json` causes a big chunk of the diff with `+657/-313`, so the net diff of this PR is `+1703/-1289`. Biggest impact is _all_ `.html` files are changed a lot.

PR #1835, opened 2026-07-29
Angular Control Flow Migration

mutatrum:
Slowly moving forward with the frontend modernization, refactoring all `*ngIf` and `*ngFor` to Angular 17+ control flow constructs. This unfortunately again touches everything, but this is a lower priority change compared to #1815.


0xf0xx0 (approved):
guess who forgot to ack :D

PR #1836, opened 2026-07-29
BM1370 ASIC Driver Timing & Initialization Fixes

mutatrum:
Adds ASIC init delays and per-chip register read loops. Also refactors VLA to a fixed buffer to eliminate stack overflow risks. 

Extracted from #1807.


mutatrum:
Merged in #1916

PR #1837, opened 2026-07-31
TPS546 PMBus Telemetry & Active-High ASIC Enable

mutatrum:
* TPS546 regulator & PMBus telemetry;
 * Add support for asic_enable_active_high and setup delays;
 * Add generic multi-phase support to power consumption self test.

Extracted from #1807


mutatrum:
Merged in #1916

PR #1838, opened 2026-07-31
Add TPS546 config to NVS

mutatrum:
This PR adds configuration options for TPS546 fields. This allows for device variants where the power setup is different, but otherwise follow the standard device model. For example #1798, the 601a 12V model would be supported by adding the following lines to the config:

```
boardversion,data,string,601a
devicemodel,data,string,gamma
asicmodel,data,string,BM1370

# TPS546 Custom 12V Overrides
tps546_phase,data,u16,0
tps546_vin_on,data,string,11.0
tps546_vin_off,data,string,10.5
tps546_vin_uvw,data,string,11.0
tps546_vin_ovf,data,string,14.0
tps546_scale,data,string,0.25
tps546_vout_min,data,string,1.0
tps546_vout_max,data,string,2.0
tps546_vout_cmd,data,string,1.2
tps546_oc_warn,data,string,25.0
tps546_oc_fault,data,string,30.0
tps546_stack,data,u16,0
tps546_sync,data,u16,16
nominal_voltage,data,u16,12
```

Not all TPS546 fields are needed, but this would fix them all for this variant. Otherwise, only changed values:
```
tps546_vin_on,data,string,11.0
tps546_vin_off,data,string,10.5
tps546_vin_uvw,data,string,11.0
tps546_vin_ovf,data,string,14.0
nominal_voltage,data,u16,12
```


WantClue (approved):
ACK

PR #1840, opened 2026-08-02
GPIO pin abstraction

mutatrum:
Extracts pin configuration for I2C, BAP and I80 connections.

Extracted from #1807


mutatrum:
Merged in #1916

PR #1842, opened 2026-08-03
Fix crash due to missing GLOBAL_STATE setter

mutatrum:
Bamboozled by #1826 being only a compile time change. It was not.


benjamin-wilson (approved):
lgtm

PR #1845, opened 2026-08-05
Harden API and WebSocket receive paths

skot:
## Summary

Harden inbound HTTP API and WebSocket handling against oversized, fragmented, and malformed client input.

- Add a bounded HTTP body receive helper alongside the existing HTTP server code.
- Keep request lengths as `size_t`, require space for the terminating NUL, and reject zero or oversized JSON bodies.
- Use the bounded receive path for system settings, pool settings, boot selection, and theme updates.
- Replace peer-sized WebSocket heap allocation with a fixed 1 KiB stack buffer.
- Reject oversized WebSocket frames so ESP-IDF closes and cleans up the session.

## Why

`PATCH /api/system` previously narrowed the network-controlled `req->content_len` from `size_t` to `int`. On ESP32, a value such as `4294967294` becomes `-2`, bypasses the upper-bound check, and reaches `buf[total_len] = '\0'`, producing an out-of-bounds write before a request body is required.

Several JSON handlers also called `httpd_req_recv()` only once, so ordinary TCP fragmentation could produce partial JSON parsing and leave request data unread. The WebSocket handler allocated `frame_length + 1` bytes without an application-level maximum, allowing a client to consume heap or leave the stream unsynchronized after allocation failure.

### Build

- ESP-IDF v5.5.1 firmware build: `idf.py build`
- `git diff --check`

### Hardware testing

Flashed the resulting firmware over USB to a physical bitaxeGamma 601 with one BM1370 and monitored the serial console throughout the tests.

The miner booted normally, joined Wi-Fi, reached its configured 525 MHz frequency, connected to CKPool, received new work, and continued submitting accepted shares during and after the malformed API traffic. At the end of the test it was hashing at approximately 1.09 TH/s with 29 accepted shares, zero rejected shares, approximately 7.65 MB free heap, and no panic, watchdog reset, allocation failure, or unexpected reboot in the serial log.

Live API cases exercised against the device included:

- Baseline system-info and theme requests returned HTTP 200.
- A valid theme body split across several TCP writes returned HTTP 200.
- A 10,239-byte settings body split into 137-byte writes was fully received and returned HTTP 200.
- Settings lengths of 10,240 and `4294967294` were safely rejected; the device remained healthy and the next system-info request returned HTTP 200.
- A normal WebSocket connection completed the HTTP 101 upgrade and remained active.
- A masked 1,025-byte WebSocket frame was rejected and its client session was removed.
[...]

PR #1846, opened 2026-08-05
Harden WebSocket handshake and frame handling

Distortions81:
## What this fixes

WebSocket requests were upgraded before LAN/type/capacity rejection, and ignored inbound frames allocated memory from a peer-controlled length. That allowed untrusted clients to consume heap or occupy upgraded connections before the server decided they were ineligible.

## Changes and reasoning

- Drain accepted inbound frames through a fixed 1 KiB stack buffer and reject larger frames instead of allocating from the advertised frame length.
- Validate the peer, endpoint type, browser `Origin`, and available capacity before the handshake; register the client only after a successful upgrade.
- Require a supplied browser `Origin` to match `Host`, while retaining origin-less command-line clients.
- Calculate capacity from `type_counts[]` while holding the client mutex, so the decision uses the same synchronized state as registration.
- Bound copied `Origin` and `Host` values to 256 bytes. The inbound server header limit is `CONFIG_HTTPD_MAX_REQ_HDR_LEN=1024`; `ESP_HTTP_CLIENT_MAX_RESPONSE_HEADER_SIZE` applies to outbound client response headers and does not control this path.

## Review follow-up

The one-use helper was inlined, constant-comparison-only tests were removed, and the behavior-focused origin/capacity tests were moved into a real registered test component. This directly addresses all current review comments.

## Scope and related work

This preserves the existing trusted-LAN design; it does not add authentication. Open PR #1845 overlaps in WebSocket/HTTP hardening, while this PR additionally covers pre-handshake same-origin enforcement, race-safe capacity accounting, and peer-sized frame allocation.

## Validation

- The added tests were compiled and linked into the test image, but were not executed locally because no ESP32-S3 QEMU or device is available.


components/api_rx/api_rx.c
```diff
@@ -0,0 +1,36 @@
...
+{
+    if (origin == NULL || host == NULL || host[0] == '\0') {
+        return false;
+    }
+
+    const char *authority = NULL;
+    static const char http_prefix[] = "http://";
+    static const char https_prefix[] = "https://";
+
+    if (strncasecmp(origin, http_prefix, sizeof(http_prefix) - 1) == 0) {
```
johnny9: Coverage: the new origin matcher tests common matches and mismatches, but not several parser boundaries used by the pre-handshake decision: empty authority/Host, case-insensitive scheme, bracketed IPv6 authority, query/fragment suffixes, absent scheme, and malformed ports. The proposal reply adds those positive and negative cases so later parser changes cannot silently widen browser handshake authorization.

johnny9 (commented):
Reviewed exact head 4103bfda3cceb39776dcfbb238918b60efc7e93d. I did not find a correctness defect in the changed handshake/frame logic. The remaining issue is incomplete boundary coverage for the new Origin/Host policy. The proposed test-only patch applies cleanly and the full ESP32-S3 QEMU suite remains green: 76 tests, 0 failures.

components/api_rx/api_rx.c
johnny9: Proposed test-only patch. It applies to the exact reviewed head and passes the full ESP32-S3 QEMU suite (76/76):

```diff
diff --git a/components/api_rx/test/test_api_rx.c b/components/api_rx/test/test_api_rx.c
index 8f12053..9525834 100644
--- a/components/api_rx/test/test_api_rx.c
+++ b/components/api_rx/test/test_api_rx.c
@@ -21,6 +21,10 @@ TEST_CASE("WebSocket origin must match request host", "[api_rx]")
         "https://BITAXE.local", "bitaxe.LOCAL"));
     TEST_ASSERT_TRUE(api_rx_websocket_origin_matches_host(
         "http://bitaxe.local:8080", "bitaxe.local:8080"));
+    TEST_ASSERT_TRUE(api_rx_websocket_origin_matches_host(
+        "HTTP://BITAXE.LOCAL", "bitaxe.local"));
+    TEST_ASSERT_TRUE(api_rx_websocket_origin_matches_host(
+        "http://[fd00::1]:8080", "[fd00::1]:8080"));
 
     TEST_ASSERT_FALSE(api_rx_websocket_origin_matches_host(
         "http://evil.local", "bitaxe.local"));
@@ -32,8 +36,19 @@ TEST_CASE("WebSocket origin must match request host", "[api_rx]")
         "file://bitaxe.local", "bitaxe.local"));
     TEST_ASSERT_FALSE(api_rx_websocket_origin_matches_host(
         "http://bitaxe.local/", "bitaxe.local"));
+    TEST_ASSERT_FALSE(api_rx_websocket_origin_matches_host(
+        "http://bitaxe.local?query", "bitaxe.local"));
+    TEST_ASSERT_FALSE(api_rx_websocket_origin_matches_host(
+        "http://bitaxe.local#fragment", "bitaxe.local"));
     TEST_ASSERT_FALSE(api_rx_websocket_origin_matches_host(
         "http://bitaxe.local@example.com", "example.com"));
+    TEST_ASSERT_FALSE(api_rx_websocket_origin_matches_host(
+        "http://bitaxe.local:invalid", "bitaxe.local"));
+    TEST_ASSERT_FALSE(api_rx_websocket_origin_matches_host(
+        "http://", "bitaxe.local"));
+    TEST_ASSERT_FALSE(api_rx_websocket_origin_matches_host(
+        "bitaxe.local", "bitaxe.local"));
     TEST_ASSERT_FALSE(api_rx_websocket_origin_matches_host(NULL, "bitaxe.local"));
     TEST_ASSERT_FALSE(api_rx_websocket_origin_matches_host("http://bitaxe.local", NULL));
+    TEST_ASSERT_FALSE(api_rx_websocket_origin_matches_host("http://bitaxe.local", ""));
 }

```

johnny9:
Physical Bitaxe 602 smoke test passed on head `40161f9a56b3f3d66c332908b6aa64b07a03e90d`.

The application-only OTA reported the expected firmware version, the device returned to healthy mining with three stable API samples, zero fault indication, and the configured pool intact. An independent authorized Stratum V1 probe received a fresh `mining.notify`. The 602 has no separate bridge firmware, so no bridge image was applied.

Full test result and artifacts: https://mining-qa-status.vercel.app/results/c9de3189-270a-428f-95c2-2e254f1dda70

Test harness: `johnny9/miner-testcode@b3694f6b01617baa8b71ed84c28300baf4cdab61`.

main/http_server/websocket.c
```diff
@@ -26,6 +30,90 @@ static SemaphoreHandle_t clients_mutex = NULL;
 static httpd_handle_t server_handle = NULL;
 static TaskHandle_t s_websocket_log_task_handle = NULL;
 
+WEBSOCKET_STATIC bool websocket_payload_fits(size_t payload_len)
```
mutatrum: This check is only used once, might as well inline it. I had tests added, but those are really low value tests, they just compare the constant.

main/http_server/websocket.c
```diff
@@ -26,6 +30,90 @@ static SemaphoreHandle_t clients_mutex = NULL;
...
+    if (authority_len == 0 || authority[authority_len] != 0) {
+        return false;
+    }
+
+    size_t host_len = strlen(host);
+    return authority_len == host_len && strncasecmp(authority, host, host_len) == 0;
+}
+
+
+WEBSOCKET_STATIC bool websocket_has_free_slot(void)
```
mutatrum, line 58: why not use the number in `type_counts[]`?

main/http_server/websocket_internal.h
```diff
@@ -0,0 +1,24 @@
...
+#define WEBSOCKET_INTERNAL_H_
+
+#include <stdbool.h>
+#include <stddef.h>
+#include "esp_err.h"
+#include "esp_http_server.h"
+#include "websocket.h"
+
+#define WS_MAX_WEBSOCKET_PAYLOAD_SIZE 1024U
+#define WS_HANDSHAKE_HEADER_SIZE 256
```
mutatrum, line 13: In `KConfig` of `esp_http_client`:
```
config ESP_HTTP_CLIENT_MAX_RESPONSE_HEADER_SIZE
    depends on ESP_HTTP_CLIENT_SAVE_RESPONSE_HEADERS
    int "Maximum size of response header"
    default 128
```

main/http_server/websocket.c
Distortions81: Addressed in 3397d98: the one-use check is now inlined. I removed tests that only compared the constant and kept the behavior-focused Origin/capacity coverage in a registered test component.

main/http_server/websocket.c
Distortions81: Addressed in 3397d98: the pre-handshake capacity check now sums type_counts[] while holding the client mutex, so it uses the same synchronized accounting as registration.

main/http_server/websocket_internal.h
Distortions81: This option is for outbound esp_http_client response headers, not inbound esp_http_server request headers. The server path uses CONFIG_HTTPD_MAX_REQ_HDR_LEN=1024; copied Origin and Host values are each capped at 256 bytes and oversized values are rejected. The published head documents that distinction.

PR #1853, opened 2026-08-07
Bound Wi-Fi scan results to fixed storage

Distortions81:
## What this fixes

The ESP-IDF scan result count was reused as the destination capacity for a fixed 20-entry global array, then copied through another fixed 20-entry HTTP buffer. More than 20 visible access points could therefore write past either array.

## Changes and reasoning

- Clamp `ap_number` directly to the existing 20-record limit before calling `esp_wifi_scan_get_ap_records()`.
- Pass the caller's destination capacity through `wifi_scan()` and clamp the copy as a second boundary defense.
- Clear stale scan state on failure and explicitly terminate copied SSIDs.

The endpoint still returns at most 20 networks; excess scan results are intentionally ignored. The direct driver-boundary clamp is the primary fix, while the explicit caller capacity keeps future callers from reintroducing the same assumption.

## Review follow-up

The implementation was simplified as requested: `ap_number` is now clamped at the API call instead of routing the limit through extra counting logic.

## Scope and related work

This does not change Wi-Fi selection policy or the trusted-network model. PR #690 previously corrected scan behavior, and #1759 audited HTTP handlers; neither enforces both fixed-capacity boundaries addressed here.

## Validation

- The ESP-IDF scan API's input/output capacity contract was checked against the production call path. A hardware scan with more than 20 visible APs was not run locally.

### DIFF
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -22,9 +22,6 @@
 #include "nvs_config.h"
 #include "esp_app_desc.h"
 
-// Maximum number of access points to scan
-#define MAX_AP_COUNT 20
-
 #if CONFIG_ESP_WPA3_SAE_PWE_HUNT_AND_PECK
 #define ESP_WIFI_SAE_MODE WPA3_SAE_PWE_HUNT_AND_PECK
 #define EXAMPLE_H2E_IDENTIFIER ""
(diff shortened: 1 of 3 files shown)


components/connect/connect.c
```diff
@@ -391,10 +405,26 @@ static void event_handler(void * arg, esp_event_base_t event_base, int32_t event
     if (event_base == WIFI_EVENT)
     {
         if (event_id == WIFI_EVENT_SCAN_DONE) {
-            esp_wifi_scan_get_ap_num(&ap_number);
```
mutatrum, line 394: Who not just clamp `ap_number` to MAX_AP_COUNT here? It seems this code is overly complex, just to limit the number.

Distortions81: Addressed in b900ddb: ap_number is now clamped directly to MAX_AP_COUNT before esp_wifi_scan_get_ap_records(). The explicit caller capacity remains only as a second boundary defense for future callers.

mutatrum (approved):

PR #1553, opened 2026-02-14
Add Stratum V2 (SV2) protocol support

warioishere:
## Summary

Adds Stratum V2 binary protocol support alongside the existing V1 JSON-RPC implementation. Tested on a BM1370 bitaxe against a local SRI server and the SRI reference pool — full 1.3 TH/s hashrate with shares accepted.

**What's included:**
- SV2 binary protocol: frame encoding/decoding, mining channel messages (SetupConnection, OpenStandardMiningChannel, NewMiningJob, SetNewPrevHash, SetTarget, SubmitSharesStandard)
- SV2 Noise encryption: Noise_NX handshake with ChaCha20-Poly1305 transport via libsecp256k1 (git submodule v0.6.0)
- Protocol coordinator with separate V1/V2 task files and fallback/recovery logic
- Protocol selectable via NVS config and AxeOS Pool Settings UI (both primary and fallback pool)

**What works:**
- Full hashrate (1.3 TH/s on BM1370)
- Encrypted connection via Noise protocol
- Share submission and acceptance
- Pool difficulty tracking via SetTarget
- V1/V2 protocol switching and fallback

## Open decisions

none

## Test plan

- [x] SV2 connection + Noise handshake against SRI server
- [x] Full hashrate (1.3 TH/s) with BM1370
- [x] Shares accepted, no duplicates
- [x] Dashboard loads in SV2 mode
- [x] V1 mode unaffected
- [x] Tested against SRI reference pool
- [x] Test on other ASIC variants (BM1366, BM1368, BM1397)

## Test servers

Our own server (confirmed working):
- **Host:** `blitzpool.yourdevice.ch`
- **Port:** `3333`
- **Authority Public Key:** `9bCoFxTszKCuffyywH5uS5o6WcU4vsjTH2axxc7wE86y2HhvULU`
- **running on mainnet**

SRI reference pool (confirmed working):
- **Host:** `75.119.150.111`
- **Port:** `3333`
- **Authority Public Key:** `9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72`

**For transparency: most of this implementation was done with the help of Claude (Opus 4.6). I hope that doesn't detract from the goal of bringing SV2 support to bitaxe.**

PR #1854, opened 2026-08-07
Bound SV2 extended-channel extranonce negotiation

Distortions81:
## What this fixes

An SV2 pool supplies extended-channel `extranonce_size` as 16 bits. The value was accepted without a bound, narrowed to eight bits, and later used with fixed 32-byte buffers. Values above 32 could cause out-of-bounds access, and values such as 256 changed meaning during narrowing.

## Changes and reasoning

- Validate the negotiated size when the success message is parsed and assigned, before any job can observe it.
- Accept only the existing safe 2-32 byte range and require exact success-message framing with no truncation or trailing bytes.
- Use one fixed protocol minimum instead of passing a redundant `min_extranonce_size` argument.
- Exit share handling immediately when the connection is absent, and retain defensive length/conversion checks at the use sites.

Valid 2-32 byte extended channels behave as before. Invalid negotiation now fails at the protocol boundary instead of surfacing later as a generic connection or job error.

## Review follow-up

This removes the questioned unsigned suffixes, redundant minimum parameter, and late job-time validation. It also makes the null-connection path explicit in share submission, directly addressing all current review comments.

## Scope and related work

PR #1553 introduced the extended-channel path and #1799 established the two-byte minimum. This PR preserves those semantics while enforcing the existing 32-byte storage limit.

## Validation

- Parser and builder boundary tests were compiled and linked, but were not executed locally because no ESP32-S3 QEMU or encrypted SV2 test peer is available.

### DIFF
--- a/components/stratum_v2/include/sv2_protocol.h
+++ b/components/stratum_v2/include/sv2_protocol.h
@@ -29,6 +29,8 @@
 #define SV2_MSG_SET_TARGET                              0x21
 
 #define SV2_MAX_MERKLE_BRANCHES 20
+#define SV2_MIN_EXTRANONCE_SIZE 2U
+#define SV2_MAX_EXTRANONCE_SIZE 32U
 
 // Extension type flag for channel messages
 #define SV2_CHANNEL_MSG_FLAG 0x8000
@@ -109,7 +111,7 @@ typedef struct sv2_conn {
     sv2_channel_type_t channel_type;
     uint8_t  extranonce_prefix[32];
     uint8_t  extranonce_prefix_len;
-    uint8_t  extranonce_size;              // total extranonce bytes assigned by pool
+    uint8_t  extranonce_size;              // locally rollable extranonce bytes
     sv2_ext_job_t *ext_pending_jobs[SV2_PENDING_JOBS_SIZE];
 } sv2_conn_t;
 
--- a/components/stratum_v2/include/sv2_protocol.h
+++ b/components/stratum_v2/include/sv2_protocol.h
@@ -29,8 +29,8 @@
 #define SV2_MSG_SET_TARGET                              0x21
 
 #define SV2_MAX_MERKLE_BRANCHES 20
-#define SV2_MIN_EXTRANONCE_SIZE 2U
-#define SV2_MAX_EXTRANONCE_SIZE 32U
+#define SV2_MIN_EXTRANONCE_SIZE 2
+#define SV2_MAX_EXTRANONCE_SIZE 32
 
 // Extension type flag for channel messages
 #define SV2_CHANNEL_MSG_FLAG 0x8000
@@ -178,7 +178,7 @@ int sv2_parse_submit_shares_error(const uint8_t *payload, uint32_t len,
 
 int sv2_build_open_extended_mining_channel(uint8_t *buf, size_t buf_len,
                                            uint32_t request_id, const char *user_identity,
-                                           float nominal_hash_rate, uint16_t min_extranonce_size);
+                                           float nominal_hash_rate);
 
 int sv2_build_submit_shares_extended(uint8_t *buf, size_t buf_len,
                                      uint32_t channel_id, uint32_t sequence_number,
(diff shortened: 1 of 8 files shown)


0xf0xx0 (approved):
code ack

components/stratum_v2/include/sv2_protocol.h
```diff
@@ -29,6 +29,8 @@
 #define SV2_MSG_SET_TARGET                              0x21
 
 #define SV2_MAX_MERKLE_BRANCHES 20
+#define SV2_MIN_EXTRANONCE_SIZE 2U
```
mutatrum: What's with the `U` suffixes?

components/stratum_v2/sv2_protocol.c
```diff
@@ -343,6 +343,11 @@ int sv2_build_open_extended_mining_channel(uint8_t *buf, size_t buf_len,
                                            uint32_t request_id, const char *user_identity,
                                            float nominal_hash_rate, uint16_t min_extranonce_size)
 {
+    if (min_extranonce_size < SV2_MIN_EXTRANONCE_SIZE ||
+        min_extranonce_size > SV2_MAX_EXTRANONCE_SIZE) {
+        return -1;
```
mutatrum: Returning `-1` here will show `SV2: Connection lost` to the user, which is not very informative.

Edit: this is always called with `min_extranonce_size` of 2. So not sure why this is tested here? Why not just supply the constant into the function call, as on the other side it's a magic number.

main/tasks/asic_result_task.c
```diff
@@ -87,14 +87,22 @@ void ASIC_result_task(void *pvParameters)
...
                     // The pool prepends its extranonce_prefix separately.
-                    uint8_t en2_len = conn->extranonce_size;
-                    uint8_t extranonce_2[32];
-                    hex2bin(active_job->extranonce2, extranonce_2, en2_len);
-                    ret = stratum_v2_submit_share_extended(GLOBAL_STATE, sv2_job_id,
-                                                           asic_result->nonce,
-                                                           active_job->ntime,
-                                                           asic_result->rolled_version,
-                                                           extranonce_2, en2_len);
+                    uint8_t en2_len = conn != NULL ? conn->extranonce_size : 0;
```
mutatrum: If `conn == NULL`, don't we just have to abort everything here? Now this check is rolled into the en2_len assignment.

main/tasks/create_jobs_task.c
```diff
@@ -312,6 +312,13 @@ static void generate_work_sv2_ext(GlobalState *GLOBAL_STATE, sv2_ext_job_t *ext_
     sv2_conn_t *conn = GLOBAL_STATE->sv2_conn;
     if (!conn) return;
 
+    uint8_t extranonce_2_len = conn->extranonce_size;
+    if (extranonce_2_len < SV2_MIN_EXTRANONCE_SIZE ||
```
mutatrum: Should this be tested here, or where `conn->extranonce_size` is set? IMO it's too late here.

components/stratum_v2/include/sv2_protocol.h
Distortions81: Addressed in e94eac5: the unnecessary U suffixes were removed from the production constants.

components/stratum_v2/sv2_protocol.c
Distortions81: Addressed in e94eac5: the redundant min_extranonce_size parameter was removed. The fixed protocol minimum is written by the builder, and the negotiated 2-32 byte range is validated in the success parser before assignment.

main/tasks/asic_result_task.c
Distortions81: Addressed in e94eac5: a null connection now logs and exits the share path immediately, before extranonce length or data is accessed.

main/tasks/create_jobs_task.c
Distortions81: Addressed in e94eac5: the primary range check moved to sv2_parse_open_extended_channel_success(), before conn->extranonce_size is assigned. The use-site check remains defensive only.

johnny9:

PR #1855, opened 2026-08-07
Bind Stratum work queue items to source epochs

Distortions81:
## What this fixes

The shared work queue stored untyped pointers and selected their interpretation/destructor from mutable global protocol state. V1/V2 failover or SV2 standard/extended transitions could therefore cast or free an item as the wrong type. A V1 producer also continued decoding through a notification after transferring that notification to the queue, allowing a consumer or source switch to free it first.

## Changes and reasoning

- Represent each queued item as an owned tagged union with an immutable kind and matching destructor.
- Bind items to a source epoch: an epoch is one generation of work for a protocol source, and clean jobs, reconnects, or source switches invalidate older generations.
- Generate work from the item's owned kind and copied extended-channel extranonce state, not mutable global mode.
- Replace the oldest item under the queue lock at capacity and make timeout/error handling exact (`ETIMEDOUT` is the only timeout result).
- Allow epoch zero and natural unsigned wrap; equality, rather than ordering, determines whether work is current.
- Finish V1 notification decoding before queue ownership transfer, eliminating the use-after-free window.

## Review follow-up

The source-epoch meaning is documented, timeout handling now follows `pthread_cond_timedwait()`, and wraparound no longer treats zero specially. Focused regressions cover epoch zero and lifecycle behavior.

## Scope and related work

This complements upstream's producer-side duplicate-job tracking; it does not replace or claim that behavior. PR #1854 independently bounds SV2 extranonce negotiation, while this PR owns queue lifetime, type, and generation safety.

## Validation

- Typed destruction, source mismatch, stale/current epoch, epoch-zero, capacity, clear, and timeout tests were compiled and linked, but not executed locally because no ESP32-S3 QEMU or device is available.
- Live V1/V2 failover and standard/extended transitions were not exercised locally.

### DIFF
--- a/main/work_queue.c
+++ b/main/work_queue.c
@@ -1,5 +1,6 @@
 #include "work_queue.h"
 #include "esp_log.h"
+#include <errno.h>
 #include <stdlib.h>
 #include <time.h>
 
@@ -15,13 +16,14 @@ void queue_init(work_queue *queue)
     pthread_cond_init(&queue->not_full, NULL);
 }
 
-work_queue_item_t work_queue_item_create(work_queue *queue, void *data,
+work_queue_item_t work_queue_item_create(work_queue *queue,
                                          work_item_kind_t kind,
+                                         work_item_data_t job,
                                          void (*free_fn)(void *))
 {
     work_queue_item_t item = {
-        .data = data,
         .kind = kind,
+        .job = job,
         .free_fn = free_fn,
     };
 
@@ -37,11 +39,11 @@ void work_queue_item_free(work_queue_item_t *item)
         return;
     }
 
-    if (item->data != NULL) {
+    if (item->job.data != NULL) {
         if (item->free_fn != NULL) {
-            item->free_fn(item->data);
+            item->free_fn(item->job.data);
         } else {
-            free(item->data);
+            free(item->job.data);
         }
     }
 
@@ -64,9 +66,6 @@ void queue_set_source(work_queue *queue, work_item_kind_t kind)
     queue_clear_locked(queue);
     queue->source_kind = kind;
     queue->source_epoch++;
-    if (queue->source_epoch == 0) {
-        queue->source_epoch = 1;
-    }
     pthread_cond_broadcast(&queue->not_empty);
     pthread_cond_broadcast(&queue->not_full);
     pthread_mutex_unlock(&queue->lock);
@@ -151,7 +150,11 @@ work_queue_item_t queue_dequeue_timeout(work_queue *queue, int timeout_ms)
 
     while (queue->count == 0) {
         int result = pthread_cond_timedwait(&queue->not_empty, &queue->lock, &timeout_time);
-        if (queue->source_epoch != starting_epoch || result != 0) {
+        if (queue->source_epoch != starting_epoch || result == ETIMEDOUT) {
+            pthread_mutex_unlock(&queue->lock);
+            return (work_queue_item_t) {0};
+        }
+        if (result != 0) {
             pthread_mutex_unlock(&queue->lock);
             return (work_queue_item_t) {0};
         }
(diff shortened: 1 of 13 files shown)


main/work_queue.h
```diff
@@ -2,25 +2,56 @@
...
+    WORK_ITEM_STRATUM_V1,
+    WORK_ITEM_STRATUM_V2_STANDARD,
+    WORK_ITEM_STRATUM_V2_EXTENDED,
+} work_item_kind_t;
+
+typedef struct
+{
+    void *data;
+    work_item_kind_t kind;
+    uint32_t source_epoch;
```
mutatrum, line 36: What is `source_epoch`?

main/work_queue.c
```diff
@@ -51,34 +130,34 @@ void *queue_dequeue(work_queue *queue)
...
+    if (timeout_time.tv_nsec >= 1000000000L) {
+        timeout_time.tv_sec += 1;
+        timeout_time.tv_nsec -= 1000000000L;
+    }
 
+    while (queue->count == 0) {
         int result = pthread_cond_timedwait(&queue->not_empty, &queue->lock, &timeout_time);
-        if (result == ETIMEDOUT) {
-            // Timeout occurred, return NULL
+        if (queue->source_epoch != starting_epoch || result != 0) {
```
mutatrum: result == `ETIMEDOUT` according to the spec of `pthread_cond_timedwait`

main/work_queue.c
```diff
@@ -2,26 +2,104 @@
...
+    }
+}
+
+void queue_set_source(work_queue *queue, work_item_kind_t kind)
+{
+    pthread_mutex_lock(&queue->lock);
+    queue_clear_locked(queue);
+    queue->source_kind = kind;
+    queue->source_epoch++;
+    if (queue->source_epoch == 0) {
```
mutatrum: Is this for when it overflows? And why can't it be 0?

mutatrum:
Ok, the system is clear now. In another PR there was a types struct used, maybe we can use something like this here as well, so simplify the code and eliminate unsafe castings. Something like:
```
typedef struct {
    work_item_kind_t kind;
    uint32_t source_epoch;
    union {
        mining_notify v1;
        sv2_job_t sv2_std;
        sv2_ext_job_t sv2_ext;
    } job;
} mining_work_item_t;
```

main/work_queue.h
Distortions81: Documented in 423bbc5: source_epoch is the generation of the active protocol source. A clean job, reconnect, or source switch advances it, so an older item is stale even when its protocol kind is unchanged.

main/work_queue.c
Distortions81: Addressed in 423bbc5: ETIMEDOUT is now handled explicitly as the timeout result; any other nonzero pthread_cond_timedwait() result also exits as an error.

main/work_queue.c
Distortions81: Addressed in 423bbc5: the epoch now uses natural uint32_t wrap and zero is a valid generation. Freshness is equality-based, so wrap does not need a reserved value; an epoch-zero regression was added.

mutatrum:
This needs more though, using the ntime directly easily fails the SV2 certificate check. The strict `(uint32_t)now < valid_from` causes handshakes to fail when connecting to SV2 pools with freshly issued certificates.

The Bitcoin block header `ntime` can easily drifts by seconds or even minutes relative to wall-clock time. Without a clock-skew tolerance, even a minor drift aborts the handshake with `Server certificate not yet valid`.

As `ntime` is also pool controlled, this check can only cause false positives. Or we have to add in a skew tolerance of 2 hours, as that's the drift a single ntime could have by itself.

johnny9:

PR #1856, opened 2026-08-07
Reject oversized BAP UART frames

Distortions81:
## What this fixes

`snprintf()` truncates safely but returns the length that would have been written. The immediate BAP UART path passed that larger length to `uart_write_bytes()`, causing a bounded stack over-read; the queued path could emit a malformed truncated frame.

## Changes and reasoning

- Centralize formatting so it either returns one complete BAP frame with checksum or fails without exposing a usable length.
- Reject invalid arguments, negative formatter results, and exact-fit/overflow truncation before either UART path sends or queues data.
- Queue only validated complete frames and handle UART return values as signed lengths.

Valid frames are byte-for-byte unchanged; oversized frames are dropped rather than truncated or read beyond their buffer.

## Review follow-up

The empty success block was removed, and the duplicated `message_len == NULL` guard was moved ahead of dereference so the remaining formatter checks each stand alone.

## Scope and related work

PR #1525 improved the BAP implementation; this PR only hardens the existing UART formatting and queue boundaries and does not change the protocol.

## Validation

- Exact-fit, overflow, invalid-argument, and undersized-buffer tests were compiled and linked, but not executed locally because no ESP32-S3 QEMU, device, or BAP UART peer is available.


main/bap/bap_uart.c
```diff
@@ -189,9 +175,9 @@ static void uart_send_task(void *pvParameters) {
                     xSemaphoreGive(bap_uart_send_mutex);
                 } else {
                     int bytes_sent = uart_write_bytes(BAP_UART_NUM, msg.message, msg.length);
-                    if (bytes_sent == msg.length) {
+                    if (bytes_sent >= 0 && (size_t)bytes_sent == msg.length) {
```
mutatrum: Empty if block (was already there?)

main/bap/bap_protocol.c
```diff
@@ -85,3 +85,40 @@ uint8_t BAP_calculate_checksum(const char *sentence_body) {
...
+
+bool BAP_format_message(char *message, size_t message_size, bap_command_t cmd,
+                        const char *parameter, const char *value, size_t *message_len) {
+    char sentence_body[BAP_MAX_MESSAGE_LEN];
+    int formatted_len;
+
+    if (message_len != NULL) {
+        *message_len = 0;
+    }
+    if (message == NULL || message_size == 0 || parameter == NULL || message_len == NULL) {
```
mutatrum: Double guard on `message_len == NULL`. Move this up so line 95 can stand on it's own.

main/bap/bap_uart.c
Distortions81: Addressed in 9127a86: the empty success block was removed and the signed UART result is checked directly.

main/bap/bap_protocol.c
Distortions81: Addressed in 9127a86: message_len is checked once before it can be dereferenced, leaving the remaining formatter guards independent.

PR #1857, opened 2026-08-08
Dashboard performance

mutatrum:
I think after #1815 there were some performance regressions. This tries to fix some:

 * Angular zone isolation for Chart.js
 * Add HeatmapLightnessPipe for memoization
 * Flash indicator timeouts runs outside Angular
 * Optimized date calculations
 * Bulk array pruning of data points
 * Shorten timeout to reconnect faster on restart


WantClue:
@mutatrum there are now conflicts :-D

0xf0xx0 (approved):
code ack, needs testin on a low power device? mines are beefy enough that i see no difference

mutatrum:
On my machine :tm: it dropped CPU usage of a dashboard tab from +50% to low 20%.

PR #1861, opened 2026-08-08
Ram first OTA progress

mutatrum:
Implements RAM-First chunked OTA streaming into PSRAM.
Displays real-time progress toaster notifications in Axe-OS Web UI via WebSocket updates.

Question: do we still need support for devices without PSRAM? As that would eliminate the fallback codepath.


WantClue:
We should remove the support for devices without PSRAM, as they are not in the specification and any manufacturer using modules without PSRAM did not follow the recommended BOM. If they want to support these devices they shall come up with their own path, we need to move on

PR #1865, opened 2026-08-10
hostname: validate charset as RFC 1123 label; keep mDNS conflict renames runtime-only

Schnitzel:
Two hostname-handling fixes:

**1. No charset validation.** The settings API length-checked `hostname` (1–32) but never validated its contents, so arbitrary strings flowed into the mDNS responder name, DHCP option 12 and the display, none of which sanitize. Values are now validated as RFC 1123 labels (letters, digits, hyphens; no leading/trailing hyphen; ≤ 63 chars) before being accepted.

**2. mDNS conflict resolution persisted spoofable renames.** `check_and_resolve_hostname_conflict` trusts any multicast A-record answer, and both call sites persisted the resulting rename to NVS — so a spoofed conflict answer could rewrite the device's configured identity and cause repeated flash writes. Conflict renames are now applied for the runtime session only; the configured hostname remains authoritative across reboots.

### DIFF
--- a/components/connect/connect.c
+++ b/components/connect/connect.c
@@ -171,10 +171,14 @@ static void initialize_mdns_if_needed(GlobalState *GLOBAL_STATE) {
             return;
         }
 
-        /* If conflict resolution changed the hostname, persist to NVS */
+        /* A conflict-resolved rename is applied at runtime only and
+         * is deliberately NOT persisted to NVS. mDNS conflict answers are
+         * unauthenticated multicast: any LAN host can spoof one, and persisting
+         * the rename let an attacker rewrite device identity and amplify flash
+         * writes. The configured hostname stays authoritative across reboots. */
         if (strcmp(final_hostname, hostname) != 0) {
-            nvs_config_set_string(NVS_CONFIG_HOSTNAME, final_hostname);
-            ESP_LOGI(TAG, "Hostname conflict resolved, updated NVS to: %s", final_hostname);
+            ESP_LOGW(TAG, "mDNS conflict: using '%s' for this session only, configured hostname '%s' unchanged",
+                     final_hostname, hostname);
         }
 
         /* Set mDNS hostname */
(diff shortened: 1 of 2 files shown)


components/connect/connect.c
```diff
@@ -171,10 +171,14 @@ static void initialize_mdns_if_needed(GlobalState *GLOBAL_STATE) {
...
-        /* If conflict resolution changed the hostname, persist to NVS */
+        /* A conflict-resolved rename is applied at runtime only and
+         * is deliberately NOT persisted to NVS. mDNS conflict answers are
+         * unauthenticated multicast: any LAN host can spoof one, and persisting
+         * the rename let an attacker rewrite device identity and amplify flash
+         * writes. The configured hostname stays authoritative across reboots. */
         if (strcmp(final_hostname, hostname) != 0) {
-            nvs_config_set_string(NVS_CONFIG_HOSTNAME, final_hostname);
-            ESP_LOGI(TAG, "Hostname conflict resolved, updated NVS to: %s", final_hostname);
+            ESP_LOGW(TAG, "mDNS conflict: using '%s' for this session only, configured hostname '%s' unchanged",
```
mutatrum, line 180: This is a weirdly worded warning. This is inside an init method, but the context of the warning implies a configuration change.

The change is correct, but now we no longer update NVS here so the 2nd part is irrelevant.

main/http_server/http_server.c
```diff
@@ -307,6 +307,30 @@ static void normalize_hostname(char *hostname, size_t max_len) {
     }
 }
 
+// Validate hostname charset (RFC 1123 label: letters, digits,
+// hyphens; no leading/trailing hyphen; <= 63 chars). The value flows into the
+// mDNS responder name, DHCP option 12 and the display, none of which sanitize.
```
mutatrum, line 312: kill comment, not relevant

main/http_server/http_server.c
```diff
@@ -307,6 +307,30 @@ static void normalize_hostname(char *hostname, size_t max_len) {
...
+    size_t len = strlen(hostname);
+    if (len == 0 || len > 63) {
+        return false;
+    }
+    if (hostname[0] == '-' || hostname[len - 1] == '-') {
+        return false;
+    }
+    for (size_t i = 0; i < len; i++) {
+        char c = hostname[i];
+        if (!((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') ||
```
mutatrum, line 326: `isalnum(c) || c == '-'`

PR #1866, opened 2026-08-10
time trust: plausibility-check pool ntime; enforce SV2 certificate validity window

Schnitzel:
Two commits that should ship together — the second's correctness depends on the first.

**Commit 1 — `system.c`: plausibility band for pool-supplied `ntime`.** Every V1 `mining.notify` and every V2 job fed pool-controlled `ntime` straight into `settimeofday()` with no bounds, and the hourly rate limiter compared against the supplied value itself, so a far-future set was sticky. The device has no RTC/NTP, so the pool effectively owned the system clock — the same clock TLS certificate expiry and the SV2 certificate window are judged against. `ntime` is now rejected if it predates the firmware build time, moves the clock backward (5 min tolerance), or jumps forward more than 4 h per accepted update; the first set after boot is capped at build time + 400 days; the rate limiter measures against the current system clock instead of the supplied value.

**Commit 2 — `sv2_noise.c`: enforce the SV2 server certificate validity window.** The Noise handshake parsed `valid_from`/`not_valid_after` from the pool's certificate message and verified the Schnorr signature, but never compared the window against current time. The window is SV2's only revocation mechanism, so a leaked or retired pool endpoint key remained usable indefinitely. Certificates outside their validity window now fail the handshake.

### DIFF
--- a/main/system.c
+++ b/main/system.c
@@ -387,12 +387,84 @@ void SYSTEM_notify_rejected_share(GlobalState * GLOBAL_STATE, char * error_msg)
     }    
 }
 
+// Pool-supplied ntime used to drive settimeofday() with no
+// plausibility band and a rate limiter that compared against the
+// supplied value itself. The device has no RTC/NTP, so the pool fully owned
+// the system clock — which is also the clock TLS certificate expiry (and the
+// SV2 certificate validity window) is judged against.
+//
+// Policy implemented here:
+//  - floor: ntime must not predate the firmware build time;
+//  - backward moves are rejected (small 5 min tolerance for pool jitter);
+//  - forward jumps are capped at 4 hours per accepted update;
+//  - the hourly rate limiter measures against the *current* system clock,
+//    not the supplied value, so a far-future set is no longer sticky.
+#define NTIME_BACKWARD_TOLERANCE_SEC (5 * 60)
+#define NTIME_MAX_FORWARD_JUMP_SEC (4 * 60 * 60)
+#define NTIME_MIN_SYNC_INTERVAL_SEC (60 * 60)
+// Cap for the very first clock set after boot (no trusted reference yet):
+// firmware build time + 400 days. Devices older than that simply skip clock
+// sync (mining is unaffected) rather than accepting an arbitrary epoch.
+#define NTIME_MAX_INITIAL_AGE_SEC (400LL * 24 * 60 * 60)
+
+// Epoch of firmware build, from the compiler's __DATE__/__TIME__ ("Mmm dd yyyy").
+// The device default timezone is UTC, so mktime() is UTC-based here.
+static time_t system_build_epoch(void)
+{
+    static const char month_names[] = "JanFebMarAprMayJunJulAugSepOctNovDec";
+    char mon[4] = {0};
+    int day = 0, year = 0, hour = 0, min = 0, sec = 0;
+    if (sscanf(__DATE__, "%3s %d %d", mon, &day, &year) != 3 ||
+        sscanf(__TIME__, "%d:%d:%d", &hour, &min, &sec) != 3) {
+        return 0;
+    }
+    const char *hit = strstr(month_names, mon);
+    if (hit == NULL) {
+        return 0;
+    }
+    struct tm build = {0};
+    build.tm_mon = (int) ((hit - month_names) / 3);
+    build.tm_mday = day;
+    build.tm_year = year - 1900;
+    build.tm_hour = hour;
+    build.tm_min = min;
+    build.tm_sec = sec;
+    return mktime(&build);
+}
+
 void SYSTEM_notify_new_ntime(GlobalState * GLOBAL_STATE, uint32_t ntime)
 {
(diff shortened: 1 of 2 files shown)


main/system.c
```diff
@@ -393,12 +387,84 @@ void SYSTEM_notify_rejected_share(GlobalState * GLOBAL_STATE, char * error_msg)
     }    
 }
 
+// Pool-supplied ntime used to drive settimeofday() with no
```
mutatrum, line 390: This is not true, `settimeofday` was not used. The miner never knows what time it is.

main/system.c
```diff
@@ -393,12 +387,84 @@ void SYSTEM_notify_rejected_share(GlobalState * GLOBAL_STATE, char * error_msg)
...
+}
+
 void SYSTEM_notify_new_ntime(GlobalState * GLOBAL_STATE, uint32_t ntime)
 {
     SystemModule * module = &GLOBAL_STATE->SYSTEM_MODULE;
 
-    // Hourly clock sync
-    if (module->lastClockSync + (60 * 60) > ntime) {
+    time_t now = time(NULL);
+    time_t floor = system_build_epoch();
```
mutatrum, line 440: This is done on every notify_new_ntime.

main/system.c
```diff
@@ -393,12 +387,84 @@ void SYSTEM_notify_rejected_share(GlobalState * GLOBAL_STATE, char * error_msg)
...
+// supplied value itself. The device has no RTC/NTP, so the pool fully owned
+// the system clock — which is also the clock TLS certificate expiry (and the
+// SV2 certificate validity window) is judged against.
+//
+// Policy implemented here:
+//  - floor: ntime must not predate the firmware build time;
+//  - backward moves are rejected (small 5 min tolerance for pool jitter);
+//  - forward jumps are capped at 4 hours per accepted update;
+//  - the hourly rate limiter measures against the *current* system clock,
+//    not the supplied value, so a far-future set is no longer sticky.
```
mutatrum, line 401: What are the official specs of ntime? Shouldn't we follow those?

main/system.c
```diff
@@ -393,12 +387,84 @@ void SYSTEM_notify_rejected_share(GlobalState * GLOBAL_STATE, char * error_msg)
...
+// Cap for the very first clock set after boot (no trusted reference yet):
+// firmware build time + 400 days. Devices older than that simply skip clock
+// sync (mining is unaffected) rather than accepting an arbitrary epoch.
+#define NTIME_MAX_INITIAL_AGE_SEC (400LL * 24 * 60 * 60)
+
+// Epoch of firmware build, from the compiler's __DATE__/__TIME__ ("Mmm dd yyyy").
+// The device default timezone is UTC, so mktime() is UTC-based here.
+static time_t system_build_epoch(void)
+{
+    static const char month_names[] = "JanFebMarAprMayJunJulAugSepOctNovDec";
```
mutatrum, line 414: It's probably simpler to inject a unix timestamp build time with some CMake stuff:
```
string(TIMESTAMP FIRMWARE_BUILD_EPOCH "%s")
target_compile_definitions(${COMPONENT_LIB} PRIVATE FIRMWARE_BUILD_EPOCH=${FIRMWARE_BUILD_EPOCH})
```

main/system.c
```diff
@@ -393,12 +387,84 @@ void SYSTEM_notify_rejected_share(GlobalState * GLOBAL_STATE, char * error_msg)
...
+        return 0;
+    }
+    struct tm build = {0};
+    build.tm_mon = (int) ((hit - month_names) / 3);
+    build.tm_mday = day;
+    build.tm_year = year - 1900;
+    build.tm_hour = hour;
+    build.tm_min = min;
+    build.tm_sec = sec;
+    return mktime(&build);
```
mutatrum, line 432: `timegm` for UTC, mktime uses local time which we don't have here, AFAIK.

But with a unix timestamp this also falls away. But beware of timezones!

0xf0xx0:
nack in its current form, we should be usin (s)ntp to set our time. theres no guarantee the pool-supplied ntime is current, and for the cert side we dont set the clock currently so theres nothing to validate against as you dont get a New(Extended)MiningJob until after the handshake. see #1786

mutatrum:
This needs more though, using the ntime directly easily fails the SV2 certificate check. The strict (uint32_t)now < valid_from causes handshakes to fail when connecting to SV2 pools with freshly issued certificates.

The Bitcoin block header ntime can easily drifts by seconds or even minutes relative to wall-clock time. Without a clock-skew tolerance, even a minor drift aborts the handshake with Server certificate not yet valid.

As ntime is also pool controlled, this check can only cause false positives. Or we have to add in a skew tolerance of 2 hours, as that's the drift a single ntime could have by itself.

0xf0xx0:
would you mind redoing this pr on top of #1900? we can skip the pool ntime check, as the miner could be rehashing old blocks (see: https://github.com/skot/ArtifactPool), but sv2/sv1 validity would be much easier now.

johnny9:

PR #1868, opened 2026-08-10
axe-os UI hardening: release-notes escaping, swarm action target, quicklink encoding

Schnitzel:
Three small, independent UI fixes (one commit each):

**1. Escape release notes before markdown rendering (`update.component.ts`).** `simpleMarkdownParser` ran its regex chain on the raw, remotely authored GitHub release body and bound the result to `[innerHTML]` with no HTML-escaping step, relying entirely on Angular's sanitizer. The input is now entity-escaped before any markup is generated, and generated `target="_blank"` anchors carry `rel="noopener noreferrer"`.

**2. Swarm actions target the connected address (`swarm.component.ts`).** The swarm scanner accepted any LAN HTTP responder as a device and took `connectionAddress` from the response-supplied `ipv4` field, so a rogue responder could redirect restart/identify POSTs to an arbitrary IP. The address actually connected to is now used. Remote `swarmColor` strings are also clamped to the known palette before being rendered as CSS values.

**3. URL-encode stratum user in pool quicklinks (`quicklink.service.ts`).** `getQuickLink` spliced `stratumUser` into pool stats URL templates with no encoding, so a value containing `/`, `?` or `#` mutated the destination. The token is now `encodeURIComponent`'d.

### DIFF
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
@@ -345,7 +345,7 @@ private isIpAddress(value: string): boolean {
         return;
       }
 
-      if (this.swarm.some(item => item.connectionAddress === info['ipv4'])) {
+      if (this.swarm.some(item => item.connectionAddress === address)) {
         this.toastr.warning('Device already added to the swarm.', `Device at ${address}`);
         return;
       }
@@ -353,7 +353,10 @@ private isIpAddress(value: string): boolean {
       const device = {
         address: info['fullHostname'] || info['hostname'] || address,
         displayName: info['hostname'] ? info['hostname'].replace(/\.local$/i, '') : address,
-        connectionAddress: info['ipv4'] || address,
+        // The action target must be the address we actually
+        // connected to, never a response-supplied field — a rogue LAN
+        // responder could otherwise redirect restart/identify POSTs.
+        connectionAddress: address,
         ...asic,
         ...info,
         ...this.numerizeDeviceBestDiffs(info)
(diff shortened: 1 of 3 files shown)


main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -519,14 +522,25 @@ private isIpAddress(value: string): boolean {
...
+    const swarmColor = safeColor || this.deriveSwarmColor(deviceModel);
     const poolDifficulty = data.poolDifficulty || data.stratumDiff;
     const hashRate = data.hashRate || data.hashRate_10m;
     return { ...data, deviceModel, swarmColor, poolDifficulty, hashRate };
   }
 
+  // swarmColor is a remote-supplied string rendered as a CSS
+  // value; clamp it to the known palette instead of trusting the peer.
+  private sanitizeSwarmColor(color: any): string | null {
+    const palette = ['red', 'purple', 'blue', 'orange', 'green', 'cyan', 'gray'];
```
mutatrum, line 540: CSS has other color options as well (`#RRGGBB` f.e.). Apparently there's such a thing as `CSS.supports('color', value);`.

This is also important for NerdAxes or new devices which this firmware might not know and use different colors.

main/http_server/axe-os/src/app/services/quicklink.service.ts
```diff
@@ -18,7 +18,9 @@ export class QuicklinkService {
    * @returns A URL to the pool's user stats page, or undefined if no matching pool is found
    */
   public getQuickLink(stratumURL: string, stratumUser: string): string | undefined {
-    const user = stratumUser.split('.')[0];
+    // stratumUser is user-controlled via the settings API;
+    // encode it before splicing it into third-party URL paths/queries.
+    const user = encodeURIComponent(stratumUser.split('.')[0]);
```
0xf0xx0, line 23: this would be the mining address, thered be no spaces or special chars to encode

PR #1873, opened 2026-08-10
Only init filesystem if custom frontend is enabled

mutatrum:
This also fixed self-test on a factory flash


mutatrum:
Replaced by #1839

PR #1878, opened 2026-08-13
sv2: handle SetExtranoncePrefix (0x19)

warioishere:
## What

Handle `SetExtranoncePrefix` (message type `0x19`). It was missing from the SV2
frame dispatch, so it hit the `default` arm and was logged as an unknown
message type.

This is what lets a pool or a proxy re-point a channel's extranonce while the
miner keeps mining — the SV2 equivalent of `mining.set_extranonce`, which the
SV1 path already opts into and handles. Useful mainly for proxies, which need
to hand each downstream miner its own extranonce.

## Implementation note

Spec 5.3.10: the new prefix applies to jobs sent *after* the message.
`generate_work_sv2_ext()` reads the prefix when it builds work, which happens
long after a job arrived and again for every `extranonce_2` — so a job now pins
the prefix that was in force when it arrived instead of reading the
connection's current one.

`extranonce_size` stays on the connection: 5.3.12 fixes it at channel opening
and this message does not change it.

## Testing

Builds clean against IDF 6.0.2, no new warnings. The parser is covered by a
native test over the B0_32 encoding (valid frame, empty prefix, length 32,
length > 32, truncated payload, short payload).

Not yet exercised on hardware against a pool that changes the prefix; the same
change on another ESP-Miner-derived firmware was, with shares accepted across
the switch and none rejected.

### DIFF
--- a/components/stratum_v2/include/sv2_protocol.h
+++ b/components/stratum_v2/include/sv2_protocol.h
@@ -25,6 +25,7 @@
 #define SV2_MSG_SUBMIT_SHARES_EXTENDED                  0x1b
 #define SV2_MSG_SUBMIT_SHARES_SUCCESS                   0x1c
 #define SV2_MSG_SUBMIT_SHARES_ERROR                     0x1d
+#define SV2_MSG_SET_EXTRANONCE_PREFIX                   0x19
 #define SV2_MSG_SET_NEW_PREV_HASH                       0x20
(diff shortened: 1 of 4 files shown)


warioishere:
its not a bug, jst an SV2 feature I missed to implement and I am experimenting on blitzpool that allows users to control their EN thatswhy I run into this

0xf0xx0 (commented):
code ack, will need to test

PR #1879, opened 2026-08-14
define ARRAY_SIZE in one place

### LINKED ISSUE #1776: array sizing for
https://github.com/bitaxeorg/ESP-Miner/pull/1754#issuecomment-4667245517

realPJL:
Fixes #1776

### DIFF
--- a/main/array.h
+++ b/main/array.h
@@ -1,5 +1,5 @@
-#ifndef UTILS_H
-#define UTILS_H
+#ifndef ARRAY_H
+#define ARRAY_H
 
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
 
--- a/main/device_config.c
+++ b/main/device_config.c
@@ -3,7 +3,7 @@
 #include "nvs_config.h"
 #include "global_state.h"
 #include "esp_log.h"
-#include "utils.h"
+#include "array.h"
 
 static const char * TAG = "device_config";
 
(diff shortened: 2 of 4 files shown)


realPJL:
I tried the "Create PR" from inside vs code but ig that leaves out the description.

I added `utils.h` which now includes the #define `ARRAY_SIZE` and is being included by `display.c` and `device_config.c`. I also checked for other implementations of `ARRAY_SIZE` but haven't found any.

mutatrum:
There's a build failure, not sure why. Maybe because the filename shadows `utils.h` in `components/asic`?

realPJL:
The simplest solution that comes to mind would be to just rename `utils.h `to something different (?)

realPJL:
The other utils.h is inside `components/stratum/include/utils.h`.

mutatrum:
Yeah, or just put it in the existing `utils.h`, that's being included everywhere already.

realPJL:
`device_config.c` and `display.c` do not (yet) include the existing utils.h.
I'd suggest renaming my `utils.h` to `array.h` and keeping it in the root directory. That way we can simply add to it if we need more array functions

realPJL:
> `device_config.c` and `display.c` do not (yet) include the existing utils.h. I'd suggest renaming my `utils.h` to `array.h` and keeping it in the root directory. That way we can simply add to it if we need more array functions

Should I implement it that way @mutatrum ?

mutatrum:
> Should I implement it that way @mutatrum ?

Go ahead, please.

realPJL:
Done. Also confirmed the build with `idf.py build`.

0xf0xx0 (approved):
ack

PR #1849, opened 2026-08-05
Harden Stratum V1 input parsing and work generation

Distortions81:
## What this fixes

Pool-controlled Stratum V1 data reached parsing, allocation, coinbase decoding, and work generation with incomplete structural and length validation. Malformed or fragmented messages could cause invalid reads, unsafe conversions, excessive allocation, partial state updates, or work queued from invalid input.

## Changes and reasoning

- Preserve fragmented and consecutive JSON-RPC lines while enforcing a 16 KiB maximum message size.
- Validate notification/result shapes, numeric and hex widths, extranonce bounds, and Merkle counts before mutating state.
- Make coinbase/varint decoding length-aware and require exactly one complete transaction, including exactly four lock-time bytes.
- Decode hex incrementally into a 128-byte stack chunk and stream it through PSA SHA-256, avoiding a coinbase-sized allocation and a second full input scan.
- Store Merkle branches in binary form and check all work-size arithmetic and metadata allocations.
- Decode and validate before counters, duplicate-ID tracking, or queue side effects; clamp the pool difficulty floor to the configured ASIC difficulty.

Upstream's duplicate-job ID tracking is retained, but duplicate suppression is not claimed as part of this PR.

## Review follow-up

This incorporates the requested suffix cleanup, single-pass decode/hash path, corrected constant/test naming, ASIC difficulty floor, removal of the unreachable no-PSRAM branch, exact lock-time framing, and malformed/boundary regressions. The final pass also handles `extranonce2`/job-ID allocation failure without leaking or dereferencing partial state.

## Scope and related work

Open PR #1844 substantially overlaps in receive/parser hardening. This PR still adds unique coinbase/work validation, exact transaction framing, streaming hashing, binary Merkle handling, allocation cleanup, and difficulty-floor enforcement. If #1844 merges first, this PR should be rebased and narrowed to those remaining pieces rather than merged unchanged.

## Validation

- The current PR diff passes `git diff --check` against upstream `master` at `d3dbcc5`.
- In a fresh combined integration worktree, the ESP32-S3 unit-test image and full firmware built successfully with the locally available ESP-IDF 5.5.3. Temporary component-version compatibility gates were used; the full build also omitted only the ESP-IDF 6-only `ws_post_handshake_cb` initializers.
- The added tests were compiled and linked into the test image, but were not executed locally because no ESP32-S3 QEMU or device is available.
- GitHub CI on the project's required ESP-IDF 6.0.2 is the authoritative current-head result and has been restarted by this update.

PR #1880, opened 2026-08-14
stratum: fix two pool-controlled crash vectors (notify type confusion, negative extranonce2_size)

Schnitzel:
Reported via coordinated disclosure by the 256 Foundation Red Team (findings RT-2026-050 / RT-2026-051). Two independent pool-controlled crash vectors in the stratum v1 parser, each reboot-looping the miner. Both are availability-only (DoS); no code execution observed. Maintainer team asked for a public PR.

## Issue 1 — mining.notify type confusion (crash → reboot loop)

`parse_mining_notify()` in `components/stratum/stratum_api.c` dereferenced `->valuestring` on cJSON items without verifying they are strings. A `mining.notify` with non-string/null `prevhash`, `coinbase1/2`, merkle entries, `version`, `nbits`, or `ntime` reached `strdup(NULL)` / `strtoul(NULL)` / `hex2bin(NULL)` and crashed the stratum task → panic → reboot → reconnect → re-crash: a permanent reboot loop.

Trigger example:
```
{"id":null,"method":"mining.notify","params":["aa",12345,"00","00",[],"20000000","1a02b290","66000000",true]}
```

**Reproduced on v2.14.2 hardware** (panic loop, 9–16 s uptime cycles, 8 cycles observed) and on a host build of master @ dec2b46 (ASan: `wrap_strdup → STRATUM_V1_parse`). Since stratum v1 is plaintext by default, the trigger is reachable by the pool operator or any LAN MITM.

**Fix (commit 1):** validate `cJSON_IsString` for params[1..3], each merkle branch entry, and params[5..7] before use; reject the message (connection survives) like the existing job_id/param-count/merkle-count checks.

## Issue 2 — negative extranonce2_size accepted (crash on next job)

`parse_set_extranonce()` / `parse_subscribe_result()` validated EN2 size only on the high side (>32 clamp). A negative size (e.g. −1) was stored as signed int, then passed to `extranonce_2_generate()` whose length parameter is `uint32_t`: −1 wraps to 4294967295 and the function `memset()`s a VLA of that size on the task stack → crash when the next job is generated.

**Master-only regression:** v2.14.2 parses the size as unsigned (−1 → 4294967295) and the existing >32 clamp fires (verified on hardware); the signed acceptance path exists on master @ dec2b46.

**Fix (commit 2):** reject `extranonce_2_len < 1` in `parse_set_extranonce` and `parse_subscribe_result`, reject it in the task handler, and guard the `generate_work()` call site (defense in depth at the VLA).

## Verification

Host fuzz harness (ESP-IDF `linux` target, ASan+UBSan, production `stratum_v1_task` + `components/stratum` built unmodified except these patches):

- Before: 7 distinct crash sites from notify type confusion; EN2 −1 accepted.
- After: **zero crashes across the full 23-case battery (120 s loop)**; all 19 isolated prober cases survive; the 7 former crash sites are clean rejects (`Invalid prev_block_hash/coinbase in mining.notify`); EN2 −1/0 rejected.
- No regressions: EN2 255/1 000 000 still clamp to 32; valid subscribe (EN2 4), valid notify parse, difficulty flows unchanged; 1 MB line tolerated; `client.reconnect` still not followed to advertised addresses.

## See also

#1881 — the difficulty-validation fix (`parse_set_difficulty` range check, our ref RT-2026-052), split out per maintainer preference.

### DIFF
--- a/components/stratum/stratum_api.c
+++ b/components/stratum/stratum_api.c
@@ -292,9 +292,19 @@ static bool parse_mining_notify(cJSON *json, StratumApiV1Message *message)
     }
 
     new_work->job_id = strdup(job_id_item->valuestring);
-    new_work->prev_block_hash = strdup(cJSON_GetArrayItem(params, 1)->valuestring);
-    new_work->coinbase_1 = strdup(cJSON_GetArrayItem(params, 2)->valuestring);
-    new_work->coinbase_2 = strdup(cJSON_GetArrayItem(params, 3)->valuestring);
+
+    cJSON *prev_block_hash_item = cJSON_GetArrayItem(params, 1);
+    cJSON *coinbase_1_item = cJSON_GetArrayItem(params, 2);
+    cJSON *coinbase_2_item = cJSON_GetArrayItem(params, 3);
+    if (!cJSON_IsString(prev_block_hash_item) || !cJSON_IsString(coinbase_1_item) || !cJSON_IsString(coinbase_2_item)) {
+        ESP_LOGE(TAG, "Invalid prev_block_hash/coinbase in mining.notify");
+        free(new_work->job_id);
+        free(new_work);
+        return false;
+    }
+    new_work->prev_block_hash = strdup(prev_block_hash_item->valuestring);
+    new_work->coinbase_1 = strdup(coinbase_1_item->valuestring);
+    new_work->coinbase_2 = strdup(coinbase_2_item->valuestring);
 
     cJSON *merkle_branch = cJSON_GetArrayItem(params, 4);
     if (!merkle_branch || !cJSON_IsArray(merkle_branch)) {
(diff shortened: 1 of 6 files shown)


Schnitzel:
**Fix verified on real hardware (2026-08-15).** Flashed a build of master + these two commits (`v2.14.0-50-ga18520a-dirty`) via OTA onto a Bitaxe (board 401 / BM1368) that previously panic-looped on v2.14.2, pointed it at a hostile pool, and re-sent the original payloads:

- `mining.notify` with numeric `prevhash` (the pre-fix crash trigger): now rejected — device log shows `Invalid prev_block_hash/coinbase in mining.notify`, mining continues uninterrupted.
- `set_extranonce` with size `−1`: now rejected — `Invalid extranonce_2_len -1 in set_extranonce`; the unit then mined 14+ hours on the hostile pool with zero resets (~23.6k shares).

(Unit reverted to stock v2.14.2 after the test.)

cbyam:
Heads up on overlap: #1849 rewrites this same parse_mining_notify path and already covers both vectors here. It type-checks every param (prevhash, coinbase1/2, each merkle entry, version, nbits, ntime) before allocating, and validates extranonce2_size, so the notify type confusion and the negative-size crash are both closed there. It builds clean and is currently mergeable.

Two things worth deciding before either lands:

1. #1849 validates up front, which lets it drop the per-field cleanup entirely. The approach here adds a free cascade to each new failure path instead. If #1849 goes in, these stratum_api.c changes become redundant and the two will conflict.
2. This PR also carries three axe-os files (update.component.ts, swarm.component.ts, quicklink.service.ts) that look like they belong to #1868, not to this stratum fix. Might want to drop them so the crash-vector fix stays isolated for the disclosure record.

If the red-team provenance (RT-2026-050/051) is the reason to keep this one separate, one option is to land the narrow stratum_api.c fix here first for fast closure, then rebase #1849 on top so the broader hardening doesn't re-touch the same lines.

johnny9:

PR #1881, opened 2026-08-14
stratum: validate set_difficulty range (reject NaN/inf/<1/>uint32)

Schnitzel:
Reported via coordinated disclosure by the 256 Foundation Red Team (finding RT-2026-052; follow-up to #1880, split out per maintainer preference).

## set_difficulty accepts −1 / 0 / inf / NaN → UB float→int cast + wrong-work modes

`parse_set_difficulty()` in `components/stratum/stratum_api.c` checked only `cJSON_IsNumber`, then stored the value verbatim into the pool difficulty. The pool (or a LAN MITM — stratum v1 is plaintext by default) could set difficulty to **−1**, **0**, **1e999** (→ +inf), or **NaN**.

Downstream, `get_difficulty_mask()` in `components/asic/asic_common.c` does `(uint32_t)ceil(difficulty)`: **undefined behavior** for negative/inf/NaN, and difficulty 0 yields mask 0 — the ASIC hardware filter then accepts every nonce, silently distorting work accounting versus the pool's share target ( floods of "easy" shares the pool rejects; local stats skew).

No crash demonstrated — this is a state-corruption / wrong-work hazard plus a UB hot spot.

**Fix:** reject non-finite, `< 1`, and `> UINT32_MAX` at parse time (message ignored, connection survives). The uint32 bound keeps the value inside the domain `get_difficulty_mask()`'s cast requires.

## Verification

Host fuzz harness (ESP-IDF `linux` target, ASan+UBSan, production `stratum_v1_task` + `components/stratum`):

- `-1`, `0`, `1e999`, `NaN` → all rejected (`Difficulty value out of range` / type check; NaN additionally rejected by cJSON's parser)
- valid `100.5` → accepted normally
- Combined with #1880's two patches: full 23-case hostile-pool battery + 19-case isolated prober pass with zero crashes / zero sanitizer reports and no regressions (EN2 clamps intact, valid subscribe/notify flows unchanged, `client.reconnect` still not followed).

### DIFF
--- a/components/stratum/stratum_api.c
+++ b/components/stratum/stratum_api.c
@@ -16,6 +16,7 @@
 #include "esp_timer.h"
 #include "esp_heap_caps.h"
 #include <inttypes.h>
+#include <math.h>
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


Schnitzel:
**Fix verified on real hardware (2026-08-15).** Same test setup (master + this commit on a Bitaxe 401, hostile pool): pool-sent `set_difficulty` with `0` and `−1` are now rejected in the device log (`Difficulty value out of range in set_difficulty: 0.000000` / `-1.000000`); valid values pass and mining continues normally.

johnny9:

PR #1883, opened 2026-08-15
Add BM1720 (Blake2b/Sia) command-layer driver + host selftest

Luckst4r:
## ⚠️ Status: Beta — not silicon-verified

> **This driver is experimental.** It was reconstructed from the disassembly of two shipped
> Bitmain firmwares (AntRouter R3-Sia and Antminer A3) and cross-checked against both — but it
> has **never been run against real BM1720 silicon**. Frame layouts and opcodes are marked
> `CONFIRMED` (read out of the disassembly) or `INFERRED` (educated guesses) throughout the
> sources. It compiles into the firmware but is deliberately **not wired into the ASIC
> dispatcher** — it cannot be selected, and it cannot mine. Treat every register write as a
> starting point for bench probing, not gospel.

## Why

The BIP-110 minority chain selected **BLAKE2b** as its replacement proof-of-work algorithm on
11 August 2026, with a hard fork proposed for **1 September 2026**. The Bitmain **BM1720** —
the ASIC inside the AntRouter R3-Sia and Antminer A3 — is proven, existing BLAKE2b silicon.

This PR lands the low-level command driver for that chip, giving ESP-Miner-class hardware a
path toward mining the new algorithm. It is the first step, not the destination: ESP-Miner's
work pipeline (stratum, job construction, share validation) is SHA256d end-to-end today, and
adapting it for BLAKE2b is tracked as follow-up work below.

## What's in this PR

| File | Purpose |
|---|---|
| `components/asic/bm1720.c` | Driver core: CRC5/CRC16, command framing, chain enumeration, PLL programming, ticket mask, register reads, nonce/response parsing |
| `components/asic/include/bm1720.h` | Public interface + protocol documentation (opcode map, frame layouts, provenance notes) |
| `components/asic/include/bm1720_pll_table.h` | 99-entry PLL table, byte-exact from the shipped firmware |
| `components/asic/test/host/bm1720_selftest.c` | Host-side selftest pinning the driver to golden wire frames from both firmware disassemblies |
| `components/asic/CMakeLists.txt` | Adds `bm1720.c` to the component build (compile coverage only — nothing dispatches to it) |

## What is verified

- ✅ Selftest passes: golden TX frames (set-address, chain-inactive, set-frequency, ticket-mask),
  response/nonce parsing incl. CRC rejection, full PLL-table consistency, CRC16 check vector
- ✅ Handles both known transports: bare frames (A3 FPGA bridge) and `0x55 0xAA`-preambled UART (R3)
- ✅ Pure C99, no dependencies — builds warning-free with `-Wall -Wextra`

Run it yourself:

```bash
cd components/asic/test/host
cc -Wall -Wextra -I../../include -o bm1720_selftest bm1720_selftest.c ../../bm1720.c
./bm1720_selftest
```

## What is *not* here yet

- [ ] **Work-send / job format** — never reverse-engineered; needs captures from live A3 hardware
- [ ] Dispatcher wiring (`asic.c` switches, `device_config.h` tables) — intentionally deferred so
      a non-functional chip can't be selected at runtime
- [ ] BLAKE2b work pipeline — job construction, share validation, 64-bit nonce plumbing
- [ ] Board definition (power/thermal/provisioning) for whatever carries the chips
- [ ] Silicon validation of every `INFERRED` register

## Notes for reviewers

The driver is self-contained (`bm1720_` prefix, no symbol overlap with `crc.c`/`pll.c`) and
kernel-style formatted. The selftest lives in `test/host/` where ESP-IDF's non-recursive
`SRC_DIRS` can't see its `main()` — same convention as `components/stratum/test/verifiers/`.


benjamin-wilson:
Absolutely not.

PR #1848, opened 2026-08-05
Harden HTTP request body and upload handling

Distortions81:
## What this fixes

Several HTTP handlers performed a single read, narrowed peer-controlled lengths before validating them, or allowed upload loops to continue without total/stall deadlines. Fragmented requests could become partial JSON, unsafe size arithmetic, or indefinitely occupied handlers.

## Changes and reasoning

- Read bounded request bodies completely across fragmentation, using `size_t` lengths and a total deadline.
- Require the entire body to be one valid JSON document and bound statistics query input.
- Reject empty and partition-oversized firmware/AxeOS uploads before the first write.
- Enforce total and no-progress upload deadlines, abort OTA cleanly on failure, and keep progress arithmetic size-safe.
- Treat impossible accounting such as `received_total > content_len` as an error rather than allowing an underflow.

The implementation uses small production helpers for size, read, and deadline decisions so the tested boundary logic is the logic used by the handlers.

## Scope and related work

This is an independent PR and preserves the existing trusted-LAN model. Open PR #1845 overlaps in bounded HTTP body handling; this PR additionally covers statistics bounds, exact JSON consumption, OTA/AxeOS partition limits, size-safe progress accounting, stall deadlines, and OTA abort behavior. It should be re-evaluated if #1845 merges first, but it is not currently obsolete.

## Validation

- The current PR diff passes `git diff --check` against upstream `master` at `d3dbcc5`.
- In a fresh combined integration worktree, the ESP32-S3 unit-test image and full firmware built successfully with the locally available ESP-IDF 5.5.3. Temporary component-version compatibility gates were used; the full build also omitted only the ESP-IDF 6-only `ws_post_handshake_cb` initializers.
- The added tests were compiled and linked into the test image, but were not executed locally because no ESP32-S3 QEMU or device is available.
- GitHub CI on the project's required ESP-IDF 6.0.2 is the authoritative current-head result and has been restarted by this update.
- The AxeOS production build also completed successfully with Node.js 22; npm reported existing dependency audit notices.

PR #1884, opened 2026-08-15
build: cap cJSON nesting depth at 32 (httpd stack overflow DoS)

Schnitzel:
## Summary

One-line config fix for an unauthenticated LAN DoS: the AxeOS API parses every request body with stock cJSON at the stock nesting limit (`CONFIG_CJSON_NESTING_LIMIT=1000`), but the httpd task runs on an **8 KiB stack** (`http_server.c:1845`) with compiler stack probing disabled. cJSON parses nested arrays/objects by mutual recursion (~70+ bytes of stack per level), so a **~220-byte** deeply-nested body overflows the stack long before the cap is reached.

## Measured on hardware (Bitaxe 401, BM1368, v2.14.2)

| Nesting depth | Body size | Result |
|---|---|---|
| 16–104 | 32–208 B | HTTP 200, mining uninterrupted |
| **112** | **224 B** | **panic reboot** (`resetReason="Software reset due to exception/panic"`) |
| **128** | **256 B** | crash reproduced |

Every JSON-accepting endpoint is reachable (`PATCH /api/system`, pool PUT/DELETE, `/api/system/boot`, `POST /api/theme`), the API has no authentication, and a sustained 1 pkt/s stream pins the unit in a reboot loop = mining halt. Body-size bounds (e.g. #1848) don't help — depth is orthogonal to size.

## Fix

`CONFIG_CJSON_NESTING_LIMIT=32` in `sdkconfig.defaults`. AxeOS bodies are flat key/value maps (deepest legitimate payload ~3 levels: the pools array), so 32 leaves ample headroom while bounding worst-case parse stack to ~2.5 KiB. The managed cJSON component already plumbs this Kconfig to `-DCJSON_NESTING_LIMIT`.

## Verification (same device, this commit)

- depth 31/32 → HTTP 200 (legit nesting unaffected)
- depth **128** and **1000** → **HTTP 400 "Invalid JSON"**, no crash, mining uninterrupted throughout
- flat `{}` PATCH → HTTP 200

(The payload used for testing is a pure JSON array — it touches nothing in NVS even when it parses, so the prober is non-mutating by construction.)

### DIFF
--- a/sdkconfig.defaults
+++ b/sdkconfig.defaults
@@ -11,6 +11,12 @@ CONFIG_HTTPD_MAX_URI_LEN=2048
 CONFIG_HTTPD_WS_SUPPORT=y
 CONFIG_HTTPD_WS_PRE_HANDSHAKE_CB_SUPPORT=y
 CONFIG_HTTPD_WS_POST_HANDSHAKE_CB_SUPPORT=y
+# Bound cJSON parse recursion: the httpd task stack is 8 KiB and the stock
+# nesting limit (1000) lets a ~220-byte deeply-nested body overflow it
+# (crash threshold measured at ~105-112 levels on hardware). AxeOS API bodies
+# are flat key/value maps (deepest legitimate payload ~3 levels), so 32 leaves
+# ample headroom while capping worst-case parse stack at ~2.5 KiB.
+CONFIG_CJSON_NESTING_LIMIT=32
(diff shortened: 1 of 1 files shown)


mutatrum (approved):
Comment can be removed, or greatly condensed. git blame points to the PR.

PR #1887, opened 2026-08-16
fix: preserve offline devices during manual refresh

r3mko:
Fixes the Swarm page becoming partially empty after manually refreshing a list containing an offline device.

## Root cause

Errors from `/api/system/info` were converted to `null` before reaching the per-device refresh error handler. The refresh result could therefore contain `null` entries, which were assigned to the swarm list and caused template rendering to fail.

Swarm

## Changes

- Route system-info failures through the existing refresh error handler.
- Preserve offline devices in the list with their live statistics cleared.
- Correct the error-handler type from an array result to a single-device result.
- Clear stale secondary metrics:
  - Rejected shares
  - Session best difficulty
  - Secondary ASIC temperature
  - Voltage-regulator temperature

## Result

When a device is unreachable, the Swarm page remains fully rendered. The device stays visible with zeroed runtime statistics while its hostname and IP are preserved.

Swarm-fix

## Testing

- Scanned a network containing multiple devices.
- Made a previously discovered device unavailable.
- Manually refreshed the Swarm page.
- Verified that the page remained rendered and responsive.
- Verified that both primary and secondary runtime values were cleared for the offline device.

### DIFF
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
@@ -293,16 +293,13 @@ private isIpAddress(value: string): boolean {
     });
   }
 
-  private getAllDeviceInfo(addresses: string[], errorHandler: (error: any, address: string) => Observable<SwarmDevice[] | null>, fetchAsic: boolean = true) {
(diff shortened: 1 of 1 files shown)


mutatrum:
Merged into #1737, thank you!

PR #1889, opened 2026-08-17
Non-blocking UART logging

mutatrum:
UART logging is really expensive because it's blocking and only running at the serial speed. This PR leverages the already in place `log_buffer.c` to defer UART logging to a thread.

Before, when a new `mining.notify` arrived:
```
I (46174) create_jobs_task: Job 6a4aed5b0001ee87 processing: 121.79 ms (stratum: 120.43 ms, gen: 1.36 ms [cb_hash: 245 us, merkle: 576 us, midstate: 178 us])
```

After:
```
I (44071) create_jobs_task: Job 6a4aed5b0001f325 processing: 2.15 ms (stratum: 0.88 ms, gen: 1.27 ms [cb_hash: 239 us, merkle: 490 us, midstate: 176 us])
```

### DIFF
--- a/main/log_buffer.c
+++ b/main/log_buffer.c
@@ -10,6 +10,7 @@
 #include "freertos/FreeRTOS.h"
 #include "freertos/queue.h"
 #include "freertos/semphr.h"
+#include "freertos/task.h"
 
 #include "log_buffer.h"
 #include "websocket.h"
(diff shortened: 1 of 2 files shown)


mutatrum:
Another measurement is from the log timing of `mining.notify` with `cleanJobs:true`. This means that a new block was found and the ASIC needs to switch jobs as quickly as possible:

Before, 84ms:
```
I (74606) stratum_api: rx: {"params":["6a4aed5b0001f53d",...],"20000000","1702353d","6a82fc67",true],"id":null,"method":"mining.notify"}
I (74690) create_jobs_task: New Work Dequeued 6a4aed5b0001f53d
```

After, 2ms:
```
I (7569025) stratum_api: rx: {"params":["6a4aed5b0001f535",...],"20000000","1702353d","6a82fbb7",true],"id":null,"method":"mining.notify"}
I (7569027) create_jobs_task: New Work Dequeued 6a4aed5b0001f535
```

WantClue (approved):

WantClue:
main/log_buffer.c:111,119–121 ignores the return values from fwrite() and fflush(). If the UART stream short-writes or fails, s_stdout_pos has already advanced, so those bytes are lost without being included in s_stdout_dropped_bytes. This is worth improving later but does not block the PR’s main goal of keeping UART latency away from mining tasks.

PR #1890, opened 2026-08-17
Add BM1372/BM1373 ASIC driver support

benjamin-wilson:
## Summary

- add BM1372 and BM1373 ASIC model support
- add the BM1373 ASIC driver and register initialization
- add device and API configuration support for the new ASIC models

This is the foundation PR for the Naja Duo 1201 series. It contains no board-specific display or self-test code.

## Validation

- full ESP-IDF 6.0.2 firmware builds pass on both downstream integration branches
- `git diff --check` passes
- regular hashrate-monitor calculations are unchanged

PR #1891, opened 2026-08-17
Add Naja Duo factory self-test

benjamin-wilson:
## Summary

- add Naja Duo factory self-test defaults and limits
- obtain per-domain measurements through the ASIC driver interface so each driver owns its register conversion
- apply BM1373 domain normalization only in its driver, without changing regular hashrate monitoring
- widen the Naja power acceptance range to 50 W +/- 10 W and add self-test cleanup improvements

## Series

Depends on #1890 and #1892. This is a sibling of #1893 and does not depend on it. After its predecessors merge, the diff reduces to the single self-test commit.

## Validation

- full ESP-IDF 6.0.2 firmware build passes
- regular `hashrate_monitor_task.c` is unchanged from the mining-support branch
- `git diff --check` passes


skot:
Hardware validation on a Naja Duo 1201 using the integrated `1201-naja-duo` branch at `3d84e44`, built with ESP-IDF 5.5.3. The factory NVS was generated from `config-1201.cvs`.

Self-test passed with 4,579.04 GH/s versus 3,738.43 GH/s expected; 2,022 valid nonces and 0 rejected. Per-domain averages ranged from 543.42 to 589.00 GH/s. Core voltage measured 994 mV against the 1,000 ±100 mV target, power was 47.84 W against the 50 W target/60 W maximum, and fan speed was 991 RPM.

The self-test flag cleared correctly and the board rebooted into normal operation. This validates the integrated series rather than the exact standalone PR head.

mutatrum:
Merged in #1916

PR #1892, opened 2026-08-17
Add Naja Duo 1201 mining support

benjamin-wilson:
## Summary

- add Naja Duo family and board configuration
- add the 1201 factory NVS configuration, including automatic fan control and a 50 C target temperature
- initialize the dual TPS546 power stages, active-high ASIC enable, reset/retry flow, and Naja thermal calibration
- disable BAP for this board family

## Series

Depends on #1890. This draft is intentionally separate from the LCD and factory self-test work. Once #1890 merges, its diff reduces to this single board-support commit.

## Validation

- full ESP-IDF 6.0.2 firmware builds pass on both downstream integration branches
- `git diff --check` passes


skot:
Tested on physical Naja Duo 1201 hardware using integrated branch `3d84e44`, ESP-IDF 5.5.3. The firmware recognized `NajaDuo`, board `1201`, initialized the TPS546 and EMC2103, detected both BM1372/BM1373 ASICs, ramped successfully to 327 MHz, reached `ASIC Ready`, and subsequently submitted accepted Stratum shares.

One provisioning observation: flashing only the normal firmware images onto this board initially left NVS reporting board version `000`, causing degraded mode. Generating `config.bin` from `config-1201.cvs` and flashing it at `0x9000` fixed the board identity and hardware initialization. The PR addition of 1201 to the factory-image matrix is therefore important; it may also be worth documenting that unprovisioned boards need the 1201 factory image.

This validates the integrated series rather than the exact standalone PR head.

main/main.c
```diff
@@ -132,6 +131,16 @@ void app_main(void)
...
     }
 
+    int i2c_sda = GLOBAL_STATE.DEVICE_CONFIG.i2c_sda != 0
+        ? GLOBAL_STATE.DEVICE_CONFIG.i2c_sda
+        : DEFAULT_GPIO_I2C_SDA;
+    int i2c_scl = GLOBAL_STATE.DEVICE_CONFIG.i2c_scl != 0
+        ? GLOBAL_STATE.DEVICE_CONFIG.i2c_scl
+        : DEFAULT_GPIO_I2C_SCL;
+
+    ESP_ERROR_CHECK(i2c_bitaxe_init(i2c_sda, i2c_scl));
```
mutatrum, line 141: Why did the i2c init move so far down? For the other devices, this means any init failure does not show up on the display.

Maybe the device_config_init can be moved up, or is something blocking that?

mutatrum:
Merged in #1916

PR #1893, opened 2026-08-17
Add Naja Duo LCD interface

benjamin-wilson:
## Summary

- add the Naja Duo ST7789 320x170 parallel LCD interface
- add the large-display statistics layout and associated fonts/images
- add reproducible LVGL image-generation tooling

## Series

Depends on #1890 and #1892. This draft contains no factory self-test changes. After its predecessors merge, the diff reduces to the single LCD commit.

## Validation

- full ESP-IDF 6.0.2 firmware build passes
- web UI production build passes under Node 22
- `git diff --check` passes

PR #1894, opened 2026-08-17
Add Gamma Hex 1300 board support

benjamin-wilson:
## Summary

- add Gamma Hex board 1300 and factory NVS configuration at 690 MHz / 1200 mV
- initialize six BM1370 ASICs using the S21-derived register sequence and read counters per chip
- configure the four-phase TPS546 stack and verify phase participation during factory self-test
- configure Gamma Hex direct fan PWM without changing other EMC2103 boards
- reuse the shared ST7789 implementation and assets from #1893
- include board 1300 in factory release builds

## Series

Supersedes #1807. Depends on #1890, #1892, #1893, and #1891. The branch contains those predecessor commits so it is conflict-free today; after they merge, this PR reduces to one Gamma Hex commit.

The Gamma-specific commit does not duplicate the Naja LCD assets, I2C pin plumbing, active-high ASIC enable support, BAP-disable support, or driver-owned self-test domain interface.

## Validation

- full ESP-IDF 6.0.2 firmware build passes
- `config-1300.cvs` generates a valid 0x6000 NVS partition
- all 56 Axe-OS tests pass under Node 22
- OpenAPI client generation passes
- `git diff --check` passes


mutatrum:
Merged in #1916

PR #1896, opened 2026-08-18
NVS, i2c and task priority stabilisation

mutatrum:
This combines 3 improvements:
 * Increase fan_controller task priority from 5 to 10, so a TLS handshake (also on prio 5 in http_server task) that takes 2 seconds won't starve the fan_controller
  * Protects i2c bus with a mutex
  * Updates NVS cache immediately, so immediate read after writes don't return stale values

### DIFF
--- a/main/nvs_config.c
+++ b/main/nvs_config.c
@@ -49,6 +49,7 @@ static const char * TAG = "nvs_config";
 static QueueHandle_t nvs_save_queue = NULL;
 static nvs_handle_t handle;
 static SemaphoreHandle_t nvs_cache_mutex = NULL;
+static SemaphoreHandle_t nvs_update_mutex = NULL;
 
 static Settings settings[NVS_CONFIG_COUNT] = {
     [NVS_CONFIG_WIFI_SSID]                             = {.nvs_key_name = "wifissid",        .type = TYPE_STR,   .default_value = {.str = (char *)CONFIG_ESP_WIFI_SSID},                .rest_name = "ssid",                               .min = 1,  .max = 32},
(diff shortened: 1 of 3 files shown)


main/nvs_config.c
```diff
@@ -641,7 +636,16 @@ uint16_t nvs_config_get_u16(NvsConfigKey key)
...
+    if (setting->value[0].u16 == value && setting->is_set) {
+        xSemaphoreGive(nvs_cache_mutex);
+        return;
+    }
+    setting->value[0].u16 = value;
+    setting->is_set = true;
+    xSemaphoreGive(nvs_cache_mutex);
 
     ConfigUpdate update = { .key = key, .type = TYPE_U16, .value.u16 = value };
     xQueueSend(nvs_save_queue, &update, portMAX_DELAY);
```
johnny9, line 707: Think the queue needs to be protected with a mutex as well to maintain cache/nvs ordering. Probably add a new nvs_update_mutex to protect this order so that you don't block cache readers using the nvs_cache_mutex.

main/nvs_config.c
```diff
@@ -602,7 +576,18 @@ char *nvs_config_get_string_indexed(NvsConfigKey key, int index)
...
-    if (!setting || setting->type != TYPE_STR || (setting->value[0].str && strcmp(setting->value[0].str, value) == 0)) return;
+    if (!setting || setting->type != TYPE_STR) return;
+
+    xSemaphoreTake(nvs_cache_mutex, portMAX_DELAY);
+    if (setting->value[0].str && strcmp(setting->value[0].str, value) == 0) {
+        xSemaphoreGive(nvs_cache_mutex);
+        return;
+    }
+    char *old_str = setting->value[0].str;
+    setting->value[0].str = strdup(value);
```
johnny9: strdup(value) should be checked to make sure the allocation didnt fail

main/nvs_config.c
```diff
@@ -602,7 +576,18 @@ char *nvs_config_get_string_indexed(NvsConfigKey key, int index)
...
+    setting->is_set = true;
+    xSemaphoreGive(nvs_cache_mutex);
+    if (old_str) free(old_str);
 
     ConfigUpdate update = { .key = key, .type = TYPE_STR, .value.str = strdup(value) };
```
johnny9: same here

main/nvs_config.c
johnny9: this function should be using the queued value and not the cached setting value now that they are on separate tasks. pass in &update

main/nvs_config.c
mutatrum: Added. There are now 2 semaphores, one for writer/writer and the other for reader/writer. Good find!

main/nvs_config.c
mutatrum: We can probably delete the fallbacks by now, but fixed this anyways.

johnny9 (approved):
ACK 9b31c1a35518a23f42fac8f2449498d6a1b325e5

tested on a gamma 602

0xf0xx0 (approved):

PR #1900, opened 2026-08-21
feat: add sntp

0xf0xx0:
puts sntp behind a nvs bool gate, defaulting to disabled  
ntp uses pool.ntp.org to find the nearest timeserver if dhcp doesn't provide server(s)

in follow-up prs we can implement cert expiry checking and tls by default, closing a major gap in security


WantClue (approved):
ACK

PR #1905, opened 2026-08-22
Fix swarm ips

### LINKED ISSUE #1769: Nerdos ip address missing on swarm
There's a slight mismatch on the API between Nerdos and AxeOS, as the IP address is missing.

Screenshot is from #1240, but I suspect this is also for `master`.

Image

WantClue:
fixes the issue that swarm doesn't show ip addresses for Nerd* and BitForge devices.

Closes #1769

### DIFF
--- a/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
+++ b/main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
@@ -156,7 +156,7 @@ export class SwarmComponent implements OnInit, OnDestroy {
 
     this.httpClient.get(`http://${window.location.hostname}/api/system/info`).subscribe({
       next: (response: any) => {
-        this.currentDeviceIp = response.ipv4;
+        this.currentDeviceIp = this.deviceIpv4(response) ?? null;
         this.currentDeviceVersion = response.version;
         this.initSwarm(response.version);
       },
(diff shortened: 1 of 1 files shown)


main/http_server/axe-os/src/app/components/swarm/swarm.component.ts
```diff
@@ -337,26 +371,35 @@ private isIpAddress(value: string): boolean {
...
       const device = {
         address: info['fullHostname'] || info['hostname'] || address,
         displayName: info['hostname'] ? info['hostname'].replace(/\.local$/i, '') : address,
-        connectionAddress: info['ipv4'] || address,
+        connectionAddress: ipv4 || address,
         ...asic,
         ...info,
-        ...this.numerizeDeviceBestDiffs(info)
+        ...this.numerizeDeviceBestDiffs(info),
+        ipv4
```
mutatrum, line 402: isn't `ipv4` already added on line 398?

WantClue: No duplicate. connectionAddress is a separate key that falls back to the hostname, and ...info only carries an ipv4 field on AxeOS (NerdOS → hostip, Bitforge → staIp from /api/ap/info), so without line 402 those devices get no ipv4 property and getDeviceLink()/isThisDevice() break again

connectionAddress on 398 is a different key

mutatrum (approved):

PR #1907, opened 2026-08-23
Adapt download assets message in update page

tassoneroberto:
# Summary

- Remove obsolete release files download UI message and download link for www.bin on assets download.
- Add a warning text in case the esp-miner.bin update asset is still missing in the latest release.

# Why

The new version of `esp-miner.bin` includes the UI update file as well, so the user does not have to download an additional `www.bin` file anymore. After checking for a new version by clicking the `Check` button in the Update section of the UI, the message "Download both files for update!" is still printed regardless if the user is running an old version requiring both files to be downloaded or the newer one.

image

# Impact

The UI behaves accordingly to the current version of the software.

# Validation

 - git diff --check

### DIFF
--- a/main/http_server/axe-os/src/app/components/update/update.component.html
+++ b/main/http_server/axe-os/src/app/components/update/update.component.html
@@ -30,23 +30,16 @@
               <a style="text-decoration: underline;" target="_blank"
               [href]="latestRelease.html_url">{{latestRelease.name}}</a>
             </h3>
-            <div class="text-orange-500 mb-3">Download both files for update!</div>
-            @for (asset of latestRelease.assets; track asset) {
-              <div>
-                @if (asset.name == 'esp-miner.bin') {
-                  <div>
-                    <a style="text-decoration: underline;" target="_blank"
-                    [href]="asset.browser_download_url">esp-miner.bin</a>
-                  </div>
-                }
-                @if (asset.name == 'www.bin') {
-                  <div>
-                    <a style="text-decoration: underline;" target="_blank"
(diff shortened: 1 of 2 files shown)


mutatrum:
Good find, but seems a solution is simpler. Any release newer than the one that contains this code only has `esp-miner.bin`, which means the warning - and the `www.bin` download link - can just be deleted.

Also, the assets also contain all factory firmwares, so it's always be > 2.

It would make sense to have a warning if `esp-miner.bin` is not part of the assets. This could mean something in the release process went wrong, or someone did a version check while the release build is still running.

tassoneroberto:
Good point. I reverted my changes and instead I removed the message completely. I also removed the www.bin if block with the downlink link.

mutatrum:
> Good point. I reverted my changes and instead I removed the message completely. I also removed the [www.bin](http://www.bin) if block with the downlink link.

Thanks. Please consider the other point as well on showing a warning if `esp-miner.bin` is not in the assets (yet). Last release took about 9 hours after the release until the assets were uploaded. A notification like `Firmware binaries are not yet available for this release. Please check back shortly.` instead of nothing makes sense IMO.

And nit: as we're only looking for a single entry, the for loop can be replaced by a find.

tassoneroberto:
Updated.

Happy path (esp-miner.bin asset exists in latest relase):

image

If the asset is missing:

image

mutatrum (approved):
LGTM, nice find!

WantClue (approved):
ACK

PR #1897, opened 2026-08-18
Instant pool switching

mutatrum:
This PR separates Stratum V1 and V2 protocols from the mining path. By generalising the whole job preparation, coinbase decoding, asic job creation and nonce return handling it greatly simplifies the code. SV1/2 protocol differences are handled by separate clients, instead of having it in separate task implementations.

Current jobs are stored in a ringbuffer, so ownership transfer and task handover is also no longer a problem, making it much easier to prevent memory leaks or use-after-free bugs. All parsing from SV1 and SV2 go directly into this ring buffer, so no intermediate allocations and/or memory ownership handovers.

It also moves a lot of pool specific fields from `GlobalState` into a connection object, so the whole mining state machine is way more modular now. It also codified ASIC capabilities and adds software version rolling into the generic mining path. This enables BM1397 devices to do header only mining (Stratum V2 Standard Channel).

Needs #1896. The live pool switching uncovered some stability bugs where the TLS handshake starved the fan_controller, killing the i2c bus.

This PR also incorporates the findings and unit tests of the following PRs:
* closes #1880 by Schnitzel
* closes #1881 by Schnitzel
* closes #1849 by Distortions81
* closes #1854 by Distortions81
* closes #1866 by Schnitzel
* closes #1855 by Distortions81
* closes #1909 by cbyam
* closes #1918 by seby1302

Furthermore, adds the following stratum protocol safety improvements as well:
 * Zero-Target & `hash_to_pdiff()` Division-by-Zero Safety
 * SV1 BIP-320 Version Rolling Mask Validation
 * SV1 Extranonce1 Parity & Length Validation
 * SV2 Channel ID Verification
 * `coinbase_decode_varint` Bounds-Checking
 * SV2 accepted_count Bounded by Pending Shares
 * Bitmask Safety & Assertion
 * `calculate_coinbase_tx_hash_bin` Allocation Safety
 * Stratum V1 `version`, `nbits`, and `ntime` Length Validation
 * `CLIENT_RECONNECT` Minimum Delay

PR #1909, opened 2026-08-23
Reject SV2 job messages that do not match the negotiated channel

cbyam:
The SV2 receive loop dispatches job messages purely on `msg_type`, while every consumer picks the job struct type from the negotiated channel type. The stratum queue holds `void *` items, so a pool that negotiated an extended channel but sends `NewMiningJob` gets an 84-byte `sv2_job_t` enqueued that `create_jobs_task` then casts to a ~700-byte `sv2_ext_job_t`. The out-of-bounds reads are bad enough, but `sv2_ext_job_free()` also frees pointers read from that out-of-bounds memory. The reverse direction (`NewExtendedMiningJob` on a standard channel) misreads the same way.

Fix: check the incoming job message against `conn->channel_type` at the dispatch switch and drop mismatched job messages with a warning, before any allocation happens. A conforming pool never sends these, so there is no behavior change for normal operation.

This is the SV2 analog of the SV1 `mining.notify` type confusion fixed in #1880, and sits alongside the other SV2 hardening in #1854 (extranonce bound) and #1855 (work-queue epochs) on this same path. It is not present in v2.15.0rc1.

Built against ESP-IDF 6.0.2.

### DIFF
--- a/main/tasks/stratum_v2_task.c
+++ b/main/tasks/stratum_v2_task.c
@@ -1003,11 +1003,22 @@ void stratum_v2_task(void *pvParameters)
             sv2_parse_frame_header(hdr_buf, &hdr);
 
             switch (hdr.msg_type) {
+                // Job messages must match the channel we negotiated. Consumers pick
+                // the job struct type from the channel type, so accepting the wrong
+                // message here would have them read and free the wrong type.
                 case SV2_MSG_NEW_MINING_JOB:
+                    if (conn->channel_type != SV2_CHANNEL_STANDARD) {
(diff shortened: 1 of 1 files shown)


mutatrum:
~Please verify against #1897, that PR did a massive refactoring of all stratum code and incorporated all the pending stratum hardening PRs. The current stratum code structure is very difficult to harden, due to all the separate code paths and memory ownership transfers. That PR tries to simplify a lot of code and eliminate many painpoints.~

I added your commit on top of #1897. As this PR simplified a lot of memory management, it probably won't crash but would either result in invalid hash efforts or invalid nonces. Still good to add the checks for protocol validity, though.

johnny9:

PR #1910, opened 2026-08-24
Add BTC price 

arcsin3x:


mutatrum:
NACK. The firmware will not connect to servers other than the pool.

PR #1911, opened 2026-08-24
Store scoreboard unordered to limit NVS writes

mutatrum:
Review comments on #1896 showed a contention point where a new high-score shifts down all lower entries, creating a bulk NVS update, up to 20 writes. This could exhaust the NVS dispatch queue. This PR switches the scoreboard to an unordered list with in-memory sorting, so every new high score table entry results in a single NVS write.

### DIFF
--- a/main/http_server/http_server.c
+++ b/main/http_server/http_server.c
@@ -1623,8 +1623,11 @@ static esp_err_t GET_scoreboard(httpd_req_t * req)
     cJSON * root = cJSON_CreateArray();
 
     if (xSemaphoreTake(scoreboard->mutex, portMAX_DELAY) == pdTRUE) {
-        for (int i = 0; i < scoreboard->count; i++) {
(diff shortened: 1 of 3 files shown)


0xf0xx0:
im struglin to read this, how does it avoid overwriting the wrong entry?

mutatrum:
> im struglin to read this, how does it avoid overwriting the wrong entry?

The NVS is stored unsorted, and on load the NVS slot is added to the record. Then the array is sorted on score. If a new score comes in that's higher than the last slot, it just uses that one and sorts it again, putting it in the correct spot.

If the scoreboard is empty (new device), all scores are 0, so the last entry (after sorting) will always be an unused slot. If the scoreboard is full, the last entry is the lowest score.

The addition of `slot_entry` to the ScoreboardEntry struct is what makes it possible. It decouples ordering from storage.

Issue #1793, opened 2026-06-30
AxeOS after two days not reachable


mcdermit667:
Hi everyone. AxeOS becomes unreachable after about two days—often with the message "Unable to reach the device for X seconds." It continues running normally in the backend. I’ve tried different browsers. Access to the dashboard is only restored after a reboot or a Wi-Fi channel change.

Can anyone confirm this issue?

WantClue:
Can you occasionally look into the logs and the system page? a couple hours apart and post that here? also are you leaving the browser tab open all the time ?

mcdermit667:
The browser tab remained open the entire time. Here is the first log file; I’ll upload the second one later. I noticed an `httpd_sock_err` on line 3,581,243.

The dashboard remained accessible the whole time.

[bitaxe-logs (1).txt](https://github.com/user-attachments/files/29546768/bitaxe-logs.1.txt)
[bitaxe-logs (2).txt](https://github.com/user-attachments/files/29553649/bitaxe-logs.2.txt)

mcdermit667:
My Bitaxe is currently inaccessible via Hashlink or Hashwatcher.
I probably should have mentioned that it drops out there first. It’s still accessible in the browser for now, but it likely won’t be long before it goes down there, too.

[bitaxe-logs (3).txt](https://github.com/user-attachments/files/29640721/bitaxe-logs.3.txt)
[bitaxe-logs (4).txt](https://github.com/user-attachments/files/29640722/bitaxe-logs.4.txt)

mcdermit667:
Unfortunately, the problem with v2.15.0 still persists.

[bitaxe-logs(7).txt](https://github.com/user-attachments/files/31364242/bitaxe-logs.7.txt)

awi81:
This may be what you're hitting. I can reproduce this failure mode on a Gamma 601.

`sdkconfig.defaults` leaves `CONFIG_LWIP_MAX_ACTIVE_TCP` at the IDF default of 16, while `http_server.c` allows `max_open_sockets = 20` (both `master` and `v2.15.x`). httpd only purges its least-recently-used session once it reaches 20, but lwIP runs out of TCP PCBs at 16, and stratum holds one of them. After that, the listen PCB can't get a PCB for a new SYN and drops it, and `tcp_kill_prio()` won't evict connections of equal priority. So every new connection times out while the miner keeps hashing, and a stratum reconnect in that state would find no PCB either.

[...]

PR #1913, opened 2026-08-25
fix: prevent reconnect storms from slow clients

r3mko:
### Problem

When the live WebSocket became unavailable, AxeOS used `/api/system/info` as a fallback.

Visible fallback polling runs every three seconds and previously used `switchMap`. If a request took longer than the polling interval, the next tick cancelled it and immediately created another request. Slow, suspended, or overloaded clients could therefore generate repeated aborted TCP connections.

The WebSocket also timed out after five seconds and retried after two seconds, creating additional connection pressure during temporary browser or network interruptions.

This could contribute to repeated HTTPD messages such as:

```
httpd_accept_conn: error in accept (113)
httpd_server: error accepting new connection
```

### Solution

- Use `exhaustMap` so only one fallback `/api/system/info` request can be active at a time.
- Ignore polling ticks while the current request is running instead of cancelling it.
- Increase the WebSocket inactivity timeout from 5 to 10 seconds.
- Increase the WebSocket retry delay from 2 to 5 seconds.
- Enable HTTPD TCP keepalive to improve cleanup of stale client connections.

### Result

Slow fallback requests can now finish normally, and temporary WebSocket interruptions produce fewer reconnect attempts. This reduces connection churn while preserving existing polling intervals and connection-state behavior.

### DIFF
--- a/main/http_server/axe-os/src/app/services/live-data.service.ts
+++ b/main/http_server/axe-os/src/app/services/live-data.service.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@angular/core';
 import { BehaviorSubject, Observable, Subject, EMPTY, timer, merge, fromEvent } from 'rxjs';
-import { catchError, retry, share, tap, switchMap, startWith, scan, shareReplay, map, timeout, bufferTime, filter, distinctUntilChanged } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


r3mko:
Might also fix issue: https://github.com/bitaxeorg/ESP-Miner/issues/1793

mutatrum (approved):
LGTM, good fix!

PR #1931, opened 2026-08-30
fix: keep pool password and Noise session keys out of the log buffer

padddi:
## Problem

`log_buffer.c` installs a global `esp_log_set_vprintf()` hook, and the resulting ring buffer is served by `GET /api/system/logs` and streamed over the `/api/ws` WebSocket. Both are gated only by `is_network_allowed()`, so **anything logged is readable by any peer on the local network**.

Two secrets were going into it at INFO level.

**1. The pool password.** `STRATUM_V1_authorize()` passed the complete `mining.authorize` line to `debug_stratum_tx()`, putting the credential in the buffer in cleartext. This defeats the masking `system_api_json.c` deliberately applies when it returns `"stratumPassword": "*****"` on the REST API.

**2. The SV2 Noise session keys.** The handshake dumped the ECDH shared secret, the chaining key before and after HKDF, and the derived `temp_k`. Together these are enough for anyone on the LAN to decrypt the session — which removes the confidentiality the Noise channel exists to provide.

## Changes

`STRATUM_V1_authorize()` no longer routes through `debug_stratum_tx()`. It logs the method and user with the credential replaced by a redaction placeholder.

Moving it to debug level alone would not have been enough: someone raising the log level to troubleshoot a pool connection would put the password back onto a network-readable endpoint, which is exactly when they are most likely to do it.

The key-material dumps in `sv2_noise.c` are **removed rather than downgraded**, with a comment explaining why they must not come back.

The `rx:` and `tx:` lines move to `ESP_LOGD`. That is the same confidentiality argument applied to the rest of the Stratum traffic, and it also takes them off the hot path — a `mining.notify` is several hundred bytes, and every log line costs a `vsnprintf`, a `memcpy` into the PSRAM ring buffer, an `esp_cache_msync`, a task notify, and a second copy in the WebSocket log task.

## What is intentionally kept

Three `ESP_LOG_BUFFER_HEX_LEVEL` calls remain in `sv2_noise.c`. Please don't pattern-match these away in a follow-up — check what the buffer holds:

| Location | Contents | Why it stays |
|---|---|---|
| `sv2_noise_selftest()`, 2 calls | `shared_a` / `shared_b` | Derived from private keys hardcoded as `1` and `2` in this file — public constants. Only print when the secp256k1 library is broken. |
| Protocol-hash mismatch path | the handshake hash | At that point it is SHA-256 of the Noise protocol name, a fixed public value. |
| Handshake steps 1, 4, 6 | handshake hash, `e_pub`, `re_pub` | On-the-wire values, not secrets. Moved to debug level to cut handshake log noise, not for confidentiality. |

## Testing

**Not compiled** — no ESP-IDF toolchain was available in the environment this was prepared in, so CI is the first real build. The changes are log-level and log-content only; no control flow is altered. `shared` is still consumed by `hkdf2()`, and no variable is left unused by the removals.

Worth confirming after flashing: `GET /api/system/logs` during a pool connect no longer contains the password, and an SV2 handshake no longer emits key material.


0xf0xx0:
out of the three this one seems to be the most viable, please rewrite yourself and test locally before opening a pr.

PR #1934, opened 2026-08-30
fix critical logic bug for a uint8_t chip_nonce_address_interval = 25…

aaron3481:
when detected_chip_count = 1, uint8_t chip_nonce_address_interval = 256 will cause it value overflow since the max value for a uint8_t is 255.

### DIFF
--- a/components/asic/bm1373.c
+++ b/components/asic/bm1373.c
@@ -313,7 +313,7 @@ uint8_t BM1373_init(GlobalState * GLOBAL_STATE)
     detected_voltage_domains = GLOBAL_STATE->DEVICE_CONFIG.family.voltage_domains;
     chip_command_address_interval = BM1372_SET_ADDRESS_STRIDE >> BM1372_COMMAND_ADDRESS_SHIFT;
     chip_response_address_interval = BM1372_SET_ADDRESS_STRIDE;
-    chip_nonce_address_interval = 256 / detected_chip_count;
+    chip_nonce_address_interval = 255 / detected_chip_count;
     if (chip_nonce_address_interval == 0) {
         ESP_LOGE(TAG, "Invalid nonce address interval for %u ASICs", detected_chip_count);
         return 0;


WantClue:
The correct solution is to widen chip_nonce_address_interval, not reduce the address-space size from 256 to 255

mutatrum:
> when detected_chip_count = 1, uint8_t chip_nonce_address_interval = 256 will cause it value overflow since the max value for a uint8_t is 255.

This is exactly what needs to happen, `chip_nonce_address_interval` will land on 0 if there's only 1 asic.

aaron3481:
> > when detected_chip_count = 1, uint8_t chip_nonce_address_interval = 256 will cause it value overflow since the max value for a uint8_t is 255.
> 
> This is exactly what needs to happen, `chip_nonce_address_interval` will land on 0 if there's only 1 asic.

detected_chip_count = chip_counter;
detected_voltage_domains = GLOBAL_STATE->DEVICE_CONFIG.family.voltage_domains;
chip_command_address_interval = BM1372_SET_ADDRESS_STRIDE >> BM1372_COMMAND_ADDRESS_SHIFT;
chip_response_address_interval = BM1372_SET_ADDRESS_STRIDE;
chip_nonce_address_interval = 256 / detected_chip_count;
if (chip_nonce_address_interval == 0) {
    ESP_LOGE(TAG, "Invalid nonce address interval for %u ASICs", detected_chip_count);
    return 0;
}

But then this code may cuase it return an error because `chip_nonce_address_interval == 0`. Is this logic correct for a single chip?

aaron3481:
also at line 515 `uint8_t asic_nr = (uint8_t)((nonce_h >> 17) & 0xff) / chip_nonce_address_interval;` If `chip_nonce_address_interval == 0`, it also causes problem?

mutatrum:
There is currently no single chip BM1373 device. The code for that chip was added recently and after a bit more experience with these devices out in the field we can probably re-use a lot of the code from the other ASICs. These issues will then be resolved.

PR #1938, opened 2026-08-31
clean up configs into config dir

WantClue:
moves all the config files into their own dir, it get's quiet busy around root dir


mutatrum:
While you're moving them, can we finally call them `.csv` instead of `.cvs`?

0xf0xx0 (approved):
lgtm

PR #1945, opened 2026-09-03
Keep saved fan mode in sync after reboot

vortexopenclaw:
## Summary

- Load the Settings form from the persisted system configuration API.
- Keep the Automatic Fan Control checkbox synchronized with the saved fan mode after a miner reboot.
- Preserve the saved manual fan-speed value when automatic control is disabled.

## Validation

- Confirmed on a live miner that manual mode and a 65% fan setting persist across reboot.
- Added a regression test for loading the persisted manual fan state.
- All 57 AxeOS tests pass.
- AxeOS production build passes.
- Full ESP32-S3 firmware build passes with 35% of the smallest application partition free.

### DIFF
--- a/main/http_server/axe-os/src/app/components/edit/edit.component.spec.ts
+++ b/main/http_server/axe-os/src/app/components/edit/edit.component.spec.ts
@@ -1,10 +1,12 @@
-import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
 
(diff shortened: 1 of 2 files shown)


vortexopenclaw:
Closing as superseded by #1946. #1945 updates AxeOS state loading, but the confirmed bug was on the Turbo Touch display: it needed the miner's persisted fan mode over BAP plus a companion display parser/UI change (bitaxeorg/BAP-GT-TOUCH#10). #1945 had no reviews or comments.

PR #1948, opened 2026-09-04
BAP: keep SET credentials out of the LAN-readable log; fix CR LF double parse

cbyam:
Two bug fixes to the BAP accessory UART, one commit each. Neither changes what an accessory can do or see; both hold with the accessory fully trusted.

**1. Don't log message bodies on the duplicate path.** The duplicate-message warning echoed the raw message, so a repeated `SET` for the Wi-Fi or pool password put the credential in the log buffer, which is served to the LAN over `/api/system/logs` and the log WebSocket. The event is now logged without the body. This is a LAN-side leak, not a UART one.

**2. Stop parsing a message twice on CR LF.** The reader parsed on `\r` but left the message open for a possible `\n`; when one arrived it was appended and the buffer parsed again. The second string differs by one byte, so the duplicate filter missed it, and `SET` handlers ran twice, including the ones that call `esp_restart()`. The message now ends on either terminator and a trailing `\n` is dropped.

Built against ESP-IDF 6.0.2.


WantClue:
the wifi password is needed for the bitaxe touch display to connect to wifi and pull other data that is not available on the bitaxe itself. This is by design, and while this is only pushing that on a physical connection it's very unlikely this can be exploited. Please remove that part as this would break the GT Touch @cbyam

WantClue:
The checksum not being required by every function is there on purpose so people can test and play around with BAP and some doesn't actually need a checksum check.

cbyam:
Understood on both, thanks. I've dropped the `wifi_password` change (didn't know the GT Touch joins Wi-Fi itself over BAP) and the checksum change, and rebased to just the two bug fixes: the credential echo into the log buffer, which is reachable over the LAN via `/api/system/logs` regardless of who's on the UART, and the CR LF double parse that runs `SET` twice. Title and description updated to match.

main/bap/bap_handlers.c
```diff
@@ -40,7 +40,9 @@ void BAP_parse_message(const char *message) {
     uint32_t current_time = esp_timer_get_time() / 1000;
     if (strcmp(message, last_processed_message) == 0 &&
         (current_time - last_message_time) < 1000) {
-        ESP_LOGW(TAG, "Duplicate message detected, ignoring: %s", message);
+        // Do not echo the body: a SET can carry the Wi-Fi or pool password, and
```
WantClue, line 43: please remove the unnecessary breaks

WantClue (changes requested):

main/bap/bap_handlers.c
cbyam: Trimmed it to a single line, and did the same to the comment in bap_uart.c since it had the same problem. The full reasoning lives in the commit messages instead.

Also rebased onto current master (606da3b), since the base predated the Stratum refactor. That PR did not touch main/bap, so the code change is byte for byte what you reviewed.

PR #1949, opened 2026-09-04
Fix memory leaks in inactive tabs

mutatrum:
One of my inactive Axe-OS tabs had 6GB of memory allocated, triggering the oomkiller.

Chartdata was bypassing visibility check
Some listeners were not properly destroyed
Disable animations on chart update


WantClue (commented):
ACK

WantClue (approved):

PR #1955, opened 2026-09-08
Dynamic ASIC difficulty

mutatrum:
All ASICs had a default ticket mask for diff 256. On the Gamma Hex with 8TH/s, this would result in ~8 shares/second. This adds pressure on the ESP32 as each share needs to be verified, on the logging and on the serial bus from the ASIC.

This PR sets the ticketmask to the lowest of either around 1 share every 5 seconds or whatever the pool decides. On most pools, this vardiff will push the difficulty up so it wants around 1 share every 5 to 10 seconds, on some pools the minimum difficulty is much higher.

If for whatever reason the difficulty is lowered, the ticketmask is updated immediately. If the difficulty is raised, there's a grace period before the asic ticket mask is updated. This prevents ticket mask oscillations on changing conditions, such as multi-pool mining.

The initial ticket mask is now sent after the ASIC init. As I've not seen any weirdness while changing the ticket mask on the fly, my assumption is this is fine, but this does need to be tested on all ASICs.

Self-test also needs verification, as a similar diff scaling is done here. Previously the ticketmask for self-test had a diff of 16. It now sets the ticketmask difficulty to have ~250 shares over the measurement period of 30 seconds. False negatives on hashrate measurements should be minimal without saturating the serial bus.


0xf0xx0 (approved):
works on my gamma

adammwest:
lowest diff that i could get 100% expected was diff 16 on a gamma, less than 16 i did not get all the shares.
current gamma (diff 256) is 1 share per second, 5s is different behavior
I think this function should return < pool_diff-1 at least always 

when the following block runs effective diff == pool diff (float problems?)
```c
if (pool_difficulty > 0.0 && pool_difficulty < target_diff) {
        target_diff = fmax(pool_difficulty, MIN_ASIC_DIFFICULTY);
        // target_diff = pool_difficulty trivially
    }
```

> If for whatever reason the difficulty is lowered, 

typically for low hashrate devices like bitaxe familly the pool schedule starts too high then lowers, I think there is a small risk in assuming downward diff updates are instantaneous.

I think it all comes down to are the current fixed difficulty wrong in some way?
I do think there is a strong case for having better minimum diffs per device

mutatrum:
> I think it all comes down to are the current fixed difficulty wrong in some way? I do think there is a strong case for having better minimum diffs per device

Yes, as we're approaching 10 TH/s machines, a 256 diff is a lot to handle, we don't need that many shares. And we don't need that much logging either. I was on the fence on aiming for a share per 2 seconds or per 5 seconds. I was tempted to also reduce the fan controller logging so they both log once every 5 seconds or so. But I can image some people like it more chatty :-)

I'll look at the other issues you raised, thanks!

PR #1957, opened 2026-09-10
Fix dashboard pool dropdown after automatic failover

johnny9:
Fixes a regression from #1897: the dropdown stayed on Primary after failover. It now follows the active pool.

Adds six tests for failover, recovery, and manual switching. All 74 frontend tests and the production build pass locally.


mutatrum (approved):
LGTM, handles edge case of selecting fallback pool when fallback is not working better that #1962.

PR #1960, opened 2026-09-10
fix(quicklink): update CKPool stats URL and match all regional stratum endpoints

Spicoul:
CKPool consolidated its user statistics to a single central endpoint (stats.ckpool.org). The current regex causes broken links (e.g. eusolostats.ckpool.org).

This PR updates the QuicklinkService regex to:
- Point directly to [https://stats.ckpool.org/users/](https://stats.ckpool.org/users/){$user} regardless of the mining server.
- Support all current regional stratum V1 endpoints (uwsolo, uesolo, eusolo, sgsolo, ausolo, solo).
- Support the default/recommended Stratum V1 and Stratum V2 endpoint (stratum.ckpool.org).


main/http_server/axe-os/src/app/services/quicklink.service.ts
```diff
@@ -33,7 +33,7 @@ export class QuicklinkService {
       { search: 'solo.stratum.braiins.com', url: `https://solo.braiins.com/stats/${user}` },
       { search: 'parasite.wtf', url: `https://parasite.space/user/${user}` },
       { search: 'sololuck.io', url: `https://sololuck.io/users/${user}` },
-      { regex: /^(eu|au)?solo[46]?.ckpool\.org/, url: `https://$1solostats.ckpool.org/users/${user}` },
+      { regex: /^((uw|ue|eu|sg|au)?solo|stratum)\.ckpool\.org$/, url: `https://stats.ckpool.org/users/${user}` },
```
0xf0xx0: nit: anything against `^(.*solo|stratum)\.ckpool\.org$`?

Spicoul: Using ^(.*solo|stratum)\.ckpool\.org$ is more permissive.
Restricting the match explicitly to (uw|ue|eu|sg|au)?solo ensures we only whitelist the actual official regional servers documented by CKPool.
The only downside is that if CKPool adds a new regional server in the future, we will need to update the regex.

0xf0xx0: yeah that downside is what i was thinkin of, the regex could get unweildly quick  
maybe `.{,2}` instead of `.*`?

Spicoul: That's a good compromise.
If a new regional server is added, it will most likely follow the two-letter convention.
Plus, it keeps the regex from becoming overly permissive.

0xf0xx0 (approved):

PR #1961, opened 2026-09-10
fix(axe-os): derive low-frequency warnings from device presets

### LINKED ISSUE #1902: AxeOS reports normal BM1372/BM1373 frequency as low
**Describe the bug**

AxeOS shows the warning banner:

> Device frequency is set low - See settings

on a Naja Duo 1201 running at its normal BM1372/BM1373 frequency of 327 MHz. The ASIC frequency is configured correctly and the miner operates normally.

The warning condition is currently hard-coded as `frequency < 400` in both:

- `main/http_server/axe-os/src/app/components/home/home.component.ts`
- `main/http_server/axe-os/src/app/components/swarm/swarm.component.ts`

BM1372/BM1373 uses a default frequency of 327 MHz and currently exposes these supported options: 327, 350, 375, 380, 400, and 410 MHz.

**To reproduce**

1. Run the Naja Duo 1201 firmware from the current Naja PR series.
2. Leave the BM1372/BM1373 frequency at the factory default of 327 MHz.
3. Open the AxeOS dashboard.
4. Observe the low-frequency warning even though the configured frequency is valid and mining is operating normally.

**Expected behavior**

AxeOS should not display a low-frequency warning when the configured frequency is within the supported range for the active ASIC.

**Suggested fix**

Replace the global 400 MHz threshold with an ASIC-aware threshold. AxeOS already obtains the device-specific settings from `/api/system/asic`, including `frequencyOptions` and `defaultFrequency`.

A robust condition would preserve the warning for a missing or zero frequency, but compare nonzero frequencies against the minimum supported value for the active device, for example:

```ts
const minimumFrequency = Math.min(...asicSettings.frequencyOptions);
const frequencyIsLow = !info.frequency || info.frequency < minimumFrequency;
```

Apply the same logic to both the main dashboard and swarm view so their warning behavior remains consistent. If ASIC settings are unavailable, AxeOS can retain a conservative fallback.

**Hardware**

- Device: Naja Duo
- Board version: 1201
- ASIC: 2x BM1372/BM1373
- Tested frequency: 327 MHz
- Voltage: 1000 mV
- Tested firmware: integrated Naja branch at `3d84e44`

Related implementation work: #1890 and #1892.

vortexopenclaw:
## Problem and solution

Naja Duo’s valid 327 MHz default was being flagged as “low frequency” because AxeOS used one global 400 MHz threshold. This PR makes the advisory device-aware by deriving the minimum from that miner’s advertised presets, so valid Naja settings stay clear while genuinely below-preset or invalid values still warn.

Fixes #1902.

## Overview

AxeOS currently treats any frequency below 400 MHz as low, regardless of the device. That produces a false warning on Naja Duo at its normal 327 MHz setting and can affect other devices whose supported presets begin below 400 MHz. The same hard-coded threshold is used on both the Home dashboard and Swarm view.

This change makes the warning device-aware. AxeOS now compares the configured frequency with the lowest valid preset advertised by that specific device. A normal Naja Duo at 327 MHz therefore stays clear, while a genuinely missing, invalid, or below-preset frequency still produces the warning. Home and Swarm use the same shared check so their behavior remains consistent.

This only changes when the existing advisory appears. It does not change mining settings, firmware safety limits, backend APIs, or dependencies.

The behavior was also exercised against local Bitaxe Supra, Bitaxe Gamma, and NerdQAxe++ miners. Each stayed clear at its own minimum preset, showed the warning below that preset, and returned to a clear state after its original setting was restored.

## Implementation details

- Read the minimum from each device's `frequencyOptions` instead of using a global threshold.
- Preserve warnings for missing, zero, negative, and otherwise invalid frequency values.
- Skip the advisory when compatible ASIC preset metadata is unavailable instead of inventing a device threshold.
- Keep Home telemetry and other notifications live while ASIC settings load or if that request fails.
- Share the helper through `utils/common-functions.ts`.
- Test low/normal frequency behavior with synthetic presets rather than copying device preset catalogs into frontend tests.

## Verification

Hardware-confirmed:

- Physical Naja Duo at 327 MHz: the false low-frequency warning no longer appears.
- Physical Naja Duo at a custom 320 MHz, below its advertised 327 MHz minimum preset: Home showed `Device frequency is set low`, and Swarm showed `Frequency Low`.
- Restoring the original frequency cleared both warnings. The check changed frequency only; voltage, fan, and pool settings were unchanged.

Hardware-backed UI checks:

- The Home view was rendered against live local Bitaxe Supra and Bitaxe Gamma APIs at 400/375 MHz. The Swarm view was tested with a mixed Bitaxe Supra, Bitaxe Gamma, and NerdQAxe++ group, including NerdQAxe++ at 500/490 MHz.
- Bitaxe Supra, Bitaxe Gamma, and NerdQAxe++ each stayed clear at its own minimum preset, warned below that preset, and cleared again after restoration.
- All temporary frequency changes were restored; accepted shares advanced during the successful test stages.

Software/build checks:

- 26 focused Home, Swarm, and shared-helper tests pass in Chrome Headless.
- Production AxeOS build passes.
- Full ESP-IDF firmware build from this commit passes with clean ESP-IDF v6.0.2.
- `git diff --check` passes.

Current-head physical validation (2026-09-17):

- Built and OTA-tested the exact PR commit `955915a` on a physical Naja Duo (board 1201, two ASICs).
- Both Home and Swarm were clear at 475 MHz and at the advertised 327 MHz minimum, warned at 320 MHz, and cleared after restoring 475 MHz: eight rendered UI checks.
- Accepted shares advanced at every stage; voltage, fan, pool, and display settings were preserved.
- Returned the miner to its original known-good firmware slot and original operating settings after the temporary test; verified mining and accepted-share advancement again.
- Application SHA256: `f68d42c264672f7dc0f660e31f89784c2d326e9e84a254ea0e5a4961e199a427`. All 12 built UI assets were verified embedded in the tested application.

The earlier Supra/Gamma/NerdQAxe++ checks above are historical hardware-backed UI checks, not fresh flashes of this revision.

CI reporting follow-up: #1985 handles missing artifacts in the results publisher. That workflow-only correction is separate from this merged firmware/UI fix.

### DIFF
--- a/main/http_server/axe-os/src/app/utils/frequency-warning.spec.ts
+++ b/main/http_server/axe-os/src/app/utils/frequency-warning.spec.ts
@@ -1,0 +1,53 @@
+import { isFrequencyLow } from './frequency-warning';
+
+describe('isFrequencyLow', () => {
+  // Device presets, not a lookup used by production code. BM1370 alone does not
+  // distinguish Gamma, Gamma Duo, Gamma Hex, and NerdQAxe++ operating points.
+  const devices = [
+    { name: 'Max', defaultFrequency: 425, options: [400, 425, 450, 475, 485, 500, 525, 550, 575, 600] },
+    { name: 'Ultra / Hex', defaultFrequency: 485, options: [400, 425, 450, 475, 485, 500, 525, 550, 575] },
+    { name: 'Supra / Supra Hex', defaultFrequency: 490, options: [400, 425, 450, 475, 485, 490, 500, 525, 550, 575] },
+    { name: 'Gamma / GT', defaultFrequency: 525, options: [400, 490, 525, 550, 600, 625, 690] },
+    { name: 'Gamma Duo', defaultFrequency: 400, options: [350, 375, 380, 400, 410] },
+    { name: 'Gamma Hex', defaultFrequency: 690, options: [400, 490, 525, 550, 600, 625, 690] },
+    { name: 'Naja Duo', defaultFrequency: 327, options: [327, 350, 375, 380, 400, 410] },
+    { name: 'NerdQAxe+', defaultFrequency: 490, options: [400, 425, 450, 475, 490, 500, 525, 550, 575] },
+    { name: 'NerdQAxe++', defaultFrequency: 600, options: [500, 515, 525, 550, 575, 590, 600] }
+  ];
+
+  for (const device of devices) {
+    it(`accepts the default and all presets for ${device.name}`, () => {
+      for (const frequency of [device.defaultFrequency, ...device.options]) {
+        expect(isFrequencyLow(frequency, device.options)).withContext(`${frequency} MHz`).toBeFalse();
+      }
+    });
+
+    it(`warns below the lowest preset for ${device.name}`, () => {
+      expect(isFrequencyLow(Math.min(...device.options) - 0.5, device.options)).toBeTrue();
+    });
+  }
+
+  it('accepts custom frequencies between presets and above the preset range', () => {
+    expect(isFrequencyLow(333.3, [327, 350, 410])).toBeFalse();
+    expect(isFrequencyLow(600, [327, 350, 410])).toBeFalse();
+  });
+
+  it('ignores invalid options and does not assume sorted presets', () => {
+    const options = [410, null, 0, -1, NaN, Infinity, '100', 327, 350];
+    expect(isFrequencyLow(327, options)).toBeFalse();
+    expect(isFrequencyLow(326, options)).toBeTrue();
+  });
+
+  it('does not guess a threshold for missing, malformed, or empty metadata', () => {
+    for (const options of [undefined, null, [], {}, '400', [0, -1, null, NaN, Infinity, '327']]) {
+      expect(isFrequencyLow(327, options)).toBeFalse();
+    }
+  });
+
+  it('warns about missing or invalid frequency even without metadata', () => {
+    for (const frequency of [undefined, null, 0, -1, NaN, Infinity, '327']) {
+      expect(isFrequencyLow(frequency, undefined)).toBeTrue();
+      expect(isFrequencyLow(frequency, [327, 350])).toBeTrue();
+    }
+  });
+});
--- a/main/http_server/axe-os/src/app/components/home/home.component.spec.ts
+++ b/main/http_server/axe-os/src/app/components/home/home.component.spec.ts
@@ -167,6 +167,7 @@ const mockLiveDataService = {
 };
 
 const mockSystemApiService = {
+  getAsicSettings: () => of({ frequencyOptions: [400, 490, 525] }),
   getStatistics: () => of(mockSystemStatistics),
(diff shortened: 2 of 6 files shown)


main/http_server/axe-os/src/app/utils/frequency-warning.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { isFrequencyLow } from './frequency-warning';
+
+describe('isFrequencyLow', () => {
+  // Device presets, not a lookup used by production code. BM1370 alone does not
+  // distinguish Gamma, Gamma Duo, Gamma Hex, and NerdQAxe++ operating points.
+  const devices = [
```
mutatrum: It's a bit overzealous to copy the complete list of all devices here. This can only drift away.

main/http_server/axe-os/src/app/utils/common-functions.ts
```diff
@@ -0,0 +1,19 @@
+/** Compare the configured frequency with this device's presets, not its ASIC name.
+ * Presets are advisory: they are not hardware limits, and custom values are allowed.
+ */
+export function isFrequencyLow(frequency: unknown, frequencyOptions: unknown): boolean {
```
mutatrum, line 4: Does this really warrant it's own file?

mutatrum (changes requested):
Seems overly verbose and IMO it's not needed to make a separate utility class for a single function.

0xf0xx0:
> Seems overly verbose and IMO it's not needed to make a separate utility class for a single function.

we should prolly make a util class to move common funcs into tho

0xf0xx0:
@vortexopenclaw 
- remove limits in frontend test, only test low and normal freq
- rename frequency-warning code to common-functions

vortexopenclaw:
Thanks for the feedback, @mutatrum and @0xf0xx0. Updated in 955915a:

- Renamed the shared helper to `utils/common-functions.ts` and updated both consumers.
- Removed the copied device/preset catalog. The frequency tests now use small synthetic low/normal cases instead of maintaining hardware limits in frontend fixtures.
- Kept focused regression coverage for missing metadata, invalid frequency, notification priority, and Home telemetry continuing when settings fail to load.

Validation: 26 focused Chrome Headless tests, the production AxeOS build, and the full clean ESP-IDF v6.0.2 build passed. I also OTA-tested the exact commit on a physical two-ASIC Naja Duo: Home and Swarm were clear at 327 MHz, warned at 320 MHz, and cleared again after restoring the original 475 MHz setting. All eight rendered checks passed, and accepted shares advanced at every stage. The miner was returned to its original firmware and settings afterward, with mining verified again.

The PR description now reflects this exact-head hardware validation and distinguishes the earlier mixed-device UI checks. Warning behavior remains unchanged by this review revision.

WantClue (approved):
ACK

main/http_server/axe-os/src/app/utils/common-functions.ts
mutatrum: It's fine for now. It'll be superceded by #1584 at some point.

mutatrum (approved):

PR #1962, opened 2026-09-10
Fix pool configuration edits, probe on primary update, and test environment

mutatrum:
Fixes the primary/fallback dropdown on the dashboard when esp-miner switches to fallback
Triggers immediate poll on primary on pool edit
Restores stratum v1 logging


johnny9 (approved):
ACK 13d7246b593865b2d8f8cf2600c8c6deb221c59d

Pool edits, Save/reload, and normal failover validated on Gamma-02 with this PR firmware. 7/8 cases passed; silent-pool failover timed out after 180s. Its receive loop is unchanged by this PR.

PR #1966, opened 2026-09-11
fix-unit test

WantClue:


0xf0xx0:
i always forget that file exists lmfao
we gotta autogenerate it somehow

PR #1971, opened 2026-09-13
Add NexusPool to quicklink service

sweethashio:
Adds NexusPool (nexuspool.io) to the quicklink list, same pattern as the other solo pools here.

The target is a public per-address page, no login or API key needed, same address-as-scope pattern SoloLuck uses right above it. `${user}` is the raw BTC address.

`solo.nexuspool.io` is the pool's current connection string, a proximity-routed name that resolves to the closest of our servers.


0xf0xx0 (approved):
ack

PR #1972, opened 2026-09-14
refactor: carry common ASIC jobs through submission and results

johnny9:
Job creation, active-job storage, nonce validation, and pool submission use `asic_job_t`, which owns the Bitcoin header fields and submission metadata. Bitmain drivers build their final wire packets immediately before transmission, removing `bm_job` and the conversion back to the common job.

Each nonce result carries the matched common job, captured under the active-slot lock. Its rolled version is derived from that same snapshot. Validation and pool submission consume the result directly, removing the public job getter and the second slot lookup that could observe a replacement job. The result's job snapshot survives active-slot reuse, and hardware job IDs stay inside the drivers.

BM1366, BM1368, BM1370, and BM1373 share a packet encoder. BM1397 has a separate encoder that generates its software midstates directly into the packet. Chip packet formats are preserved. Version rolling also now carries and wraps within the negotiated mask, fixing the sparse-mask and wraparound bug exposed by the new regression tests.

This keeps Stratum independent of Bitmain internals and provides a common job interface for other ASIC implementations.

The QEMU test task uses an 8 KB stack with the stack-end watchpoint enabled. Job-building and hashing tests exceed ESP-IDF's 3,584-byte default; that overflow could corrupt heap metadata and surface later as a JSON parser allocation crash. A diagnostic run with heap integrity checks passed with about 6 KB peak stack usage.

Validation: all 145 QEMU tests pass, including exact packet bytes for all five drivers, BM1397 midstate/version mapping, result snapshots across slot replacement for every driver and job protocol, invalid response IDs, and allocation failure recovery. Result-task tests run without an active-job store. The ESP32-S3 firmware builds with ESP-IDF 6.0.2 using existing AxeOS assets. Hardware validation has not been run for this revision.


components/asic/private_include/bm_job.h
```diff
@@ -0,0 +1,42 @@
...
+    uint32_t version;
+    uint32_t version_mask;
+    uint8_t prev_block_hash[32];
+    uint8_t merkle_root[32];
+    uint32_t ntime;
+    uint32_t nbits;
+    uint32_t starting_nonce;
+
+    uint8_t num_midstates;
+    uint8_t midstates[BM_JOB_MAX_MIDSTATES][32];
```
mutatrum, line 20: Nit: this should ideally have `__attribute__((aligned(4)))` due to cast to `uint32_t` in `reverse_32bit_words`.

components/asic/bm_job.c
```diff
@@ -0,0 +1,88 @@
...
+        .version_mask = source->version_mask,
+        .nbits = source->nbits,
+        .ntime = source->ntime,
+        .starting_nonce = source->starting_nonce,
+        .pool_diff = source->pool_diff,
+        .pool_id = source->pool_id,
+        .job_type = source->source_type,
+    };
+    job.job_id = strdup(source->job_id);
+    job.extranonce2 = strdup(source->extranonce2);
```
mutatrum: If these would also be defined as
`
    char job_id[ASIC_JOB_ID_LEN];
    char extranonce2[ASIC_JOB_EXTRANONCE2_HEX_SIZE];
`
these strdups can also be eliminated.

components/asic/include/asic.h
```diff
@@ -16,7 +17,11 @@ typedef struct {
 uint8_t ASIC_init(GlobalState * GLOBAL_STATE);
 task_result * ASIC_process_work(GlobalState * GLOBAL_STATE);
 int ASIC_set_max_baud(GlobalState * GLOBAL_STATE);
-void ASIC_send_work(GlobalState * GLOBAL_STATE, bm_job * next_job);
+/* The caller retains ownership and may release the job after this call. */
+void ASIC_send_job(GlobalState *state, const asic_job_t *job);
+/* Copies a valid job under the slot lock. The snapshot survives slot reuse
+ * and is unchanged on failure. */
+bool ASIC_get_job_snapshot(GlobalState *state, uint8_t job_id, asic_job_t *snapshot);
```
johnny9: need to double check that this is necessary. likely ownership related

components/asic/include/asic_common.h
```diff
@@ -7,6 +7,9 @@
 #include "freertos/FreeRTOS.h"
 #include "freertos/task.h"
 
+/* Convert the Bitmain hash counter register to GH/s. */
```
johnny9: this looks bitmain specific but the file is asic common

johnny9 (commented):
double check the correctness of stratum->bm_job

components/asic/bm_job.c
```diff
@@ -0,0 +1,80 @@
...
+    *destination = job;
+    return true;
+}
+
+void free_bm_job(bm_job *job)
+{
+    free(job);
+}
+
+bool bm_job_to_asic_job(const bm_job *source, asic_job_t *destination)
```
mutatrum: Conceptually I don't understand why we would need a transformation from bm_job to asic_job. The pipeline should only go in one direction? Maybe I'm misreading the code, not sure :-)

johnny9: It was a storage decision which I also don't think i agree with. Need to do another round to see if I can get away with not duplicating both. But it's going back to asic job for share submissions

johnny9: I think Im going to remove bm job and just convert from the generic asic job before sending

PR #1976, opened 2026-09-15
ci: add QEMU UBSan tests and fix unaligned memory access

johnny9:
Adds a separate UBSan job for QEMU tests, with alignment and shift checks enabled.

Fixes unaligned hash access and allocations for cJSON and Stratum timing records. Firmware and tests use the same cJSON allocator.

Run locally with `bash tools/run_qemu_tests.sh --ubsan`.

All 105 tests pass with and without UBSan at both `-Og` and `-O2`, including mixed alignment and overlapping buffers.


components/stratum/utils.c
```diff
@@ -201,31 +201,23 @@ void midstate_sha256_bin(const uint8_t *data, const size_t data_len, uint8_t des
...
-    d[1] = s[6];
-    d[2] = s[5];
-    d[3] = s[4];
-    d[4] = s[3];
-    d[5] = s[2];
-    d[6] = s[1];
-    d[7] = s[0];    
+    uint8_t reversed[32];
+
+    for (size_t i = 0; i < 8; i++) {
```
mutatrum: Maybe add a comment that this loop and memcpy will all be compiled away into register operations

johnny9: I dug into this and checked the ESP32-S3 assembly. ESP-IDF sets `-fno-builtin-memcpy`, so the loop was keeping its copy calls.

I restored the unrolled assignments, aligned the midstate buffers, and added a checked fast path using `__builtin_memcpy`. At `-O2`, both copies disappear, leaving eight word loads and eight stores. Unaligned buffers keep a safe fallback.

All 105 tests pass with and without UBSan at both `-Og` and `-O2`, including mixed alignment and overlapping buffers.

PR #1977, opened 2026-09-16
fix(axe-os): make meter colour carry meaning, and fix the near-white meter track

SurefireStudios:
Two dashboard colour problems, three files.

**The meter track was `#dee2e6`** — a near-white bar sitting on a near-black card, so every meter read as almost-full before any value was drawn. It is now `rgba(255, 255, 255, 0.10)` on dark and a matching dark tint on the white theme.

**Every meter was the same brand red**, whether it showed 20% or 100%, so a Bitaxe at its thermal ceiling looked exactly like one idling. Temperature and power meters now shift colour only as they approach their maximum; everything else keeps the theme colour.

## Before / after

Power at 19.4 W of 20 W, ASIC at 72 °C of 75 °C:

| Before | After |
| --- | --- |
| Power card before | Power card after |
| Heat card before | Heat card after |

Note that *ASIC Frequency* stays brand red in the "after" shot: it has no danger threshold, so it keeps the theme colour.

At normal values nothing changes colour — only the track:

| Before | After |
| --- | --- |
| Normal before | Normal after |

## How it works

`app-progressbar` gains a `tone` input: `brand` (the default), `auto`, `warn` and `critical`.

`auto` is brand up to 85% of maximum, warn to 95%, critical above that — a working Bitaxe idles near 80% of its temperature ceiling, so the thresholds sit above normal running.

Warn and critical are Okabe-Ito amber `#e69f00` and vermillion `#d55e00`, which stay distinguishable from each other and from the brand red under deuteranopia, protanopia and tritanopia.

The voltage-regulator meter uses the same 105 °C threshold as the warning text directly below it, rather than `auto`.

## Review changes

Per @0xf0xx0: the card gradient, the 10px border radius, the `color-mix` severity derivation and the tabular-figures rule are all reverted, and safe or threshold-less bars now use `--color-primary` instead of a separate palette entry — which removed the `ok` and `neutral` tones entirely.

The chart fill fix is pulled out of this PR and will be raised separately.

That took it from 7 files to 3.

## Checks

`ng build` passes.

Screenshots are from this branch and its merge base, rendered against the fixture in `home.component.spec.ts` with the API stubbed, so before and after differ only by the commit.


0xf0xx0:
mmm nack on
- the bar color changes, using the defined theme color for "safe" values and bars without thresholds would be preferred
	using `color-mix` is overcomplicating, just use the existing colors
	also, warning and critical colors arent colorblind-friendly
- gradient on cards, just looks out of place imo
	flat is fine, theyre already sitting on a darker surface with a thin light border
also
- chart color fix could be split out

not sure how i feel about the border radius change, why?

SurefireStudios:
all fair, pushed. 7 files down to 3.

- safe + no-threshold bars use `--color-primary` now. that killed `--color-status-ok`, `--color-meter-neutral` and the `neutral` tone entirely — a bar with no threshold is just the default, so it needs no markup. tone is `brand | auto | warn | critical`, four bars opt in
- no `color-mix` anywhere, message severity is back to the colours that were already there
- you're right on the palette, amber next to red is basically one colour under deuteranopia. swapped to okabe-ito orange `#e69f00` / vermillion `#d55e00`, darker pair for the white theme
- gradient gone, plus the hover shadow and the card transition
- chart fix is #1979

border radius: no good reason, i was just softening the cards while i was in there. reverted, along with a tabular-nums rule that didn't belong either.

what's left is the tone input, four bindings, and `--slider-bg` — it was `#dee2e6` declared once and never overridden, so a near-white rail on a near-black card.

builds clean on node 24 / ng 19.

SurefireStudios:
Chart fix split out as #1981, as asked.

Also re-shot the before/after here — the old images still showed the gradient and the 10px radius, so they no longer matched the diff. The new ones are the Power and Heat cards at 19.4 W of 20 W and 72 °C of 75 °C. `ASIC Frequency` staying brand red in the after shot is the "no threshold keeps the theme colour" case.

PR #1978, opened 2026-09-16
feat(axe-os): add Midnight and Slate colour schemes, from a single scheme registry

SurefireStudios:
Stacked on #1977 — the meter-colour commits appear here until that merges.

AxeOS ships five surface options, but each was a hand-maintained pair: a `theme-*` block in `_theme.scss` plus a matching entry in the config component, with "is this dark?" written as `id !== 'white'`.

`color-schemes.ts` makes the list the single source: one entry defines the id, label, whether it is dark, and a one-line description. The config component renders from it, `LayoutService` clears classes from it, and `isDarkScheme()` replaces the equality check.

Two schemes on top of that.

**Midnight** is true black, for OLED panels and dark rooms where the default deep navy reads as grey haze against the bezel.

**Slate** is a neutral mid-grey with lower contrast, for a dashboard left on a lit screen.

| Midnight | Slate |
| --- | --- |
| Midnight | Slate |

Each option now carries a description, since "Light" being a navy was not obvious from the label alone.

## Checks

`ng build` passes.

Rebased onto the reviewed #1977, so the card gradient, border radius and typography changes that were dropped there are not reintroduced here — the two new scheme blocks define only variables the stylesheet still reads.

That took this branch from 11 files to 5 of its own.


SurefireStudios:
Rebased onto the reviewed #1977, so this no longer carries the card gradient, the border-radius change or the typography edit that were dropped there. The two new scheme blocks now define only variables the stylesheet still reads — 11 files down to 5 of its own. Screenshots re-shot from this branch.

PR #1979, opened 2026-09-16
fix(axe-os): stop the chart stacking a fill per series

SurefireStudios:
Split out of #1977 at @0xf0xx0's request — no colour-meaning changes here, just the chart.

### Stacked fills

`createChartDatasets` passes the same `fill` flag to every series in a group:

```ts
return labels.filter(...).map((labelKey, index) => {
  ...
  return { type: 'line', label, data: ..., fill, backgroundColor, borderColor, ... };
});
```

`backgroundColor` is `color-mix(in srgb, ${borderColor}, transparent 81%)` — 19% opacity. With four supported sensors in a group that is four washes drawn on top of each other, and the stacked alpha makes the y1 area read as a solid block rather than a chart. Boards with fewer sensors show it less, which is probably why it survived.

Only the leading series in a group is filled now. One line:

```ts
const filled = fill && index === 0;
```

### Second axis colour

Separately, the y2 group was built with `textColorSecondary` as its base and `'black'` as its mix colour:

```ts
...this.createChartDatasets('chartY2Unit', textColorSecondary, 'black', false, 'y2')
```

On the dark theme that mixes each trailing series *toward the background*, so the further down the group a sensor sits the closer it gets to invisible. Adds `--chart-axis2-color`, mixes toward `--color-text-main` instead, and points the y2 tick colour at the same variable so the axis matches the series it scales.

If you would rather take only the fill fix and leave the axis colour alone, say so and I will drop the second half — they are independent.

### Verified

`npm run build` in `main/http_server/axe-os` completes clean on Node 24 / Angular 19.

### Before / after

| Before | After |
| --- | --- |
| Chart before | Chart after |

Those show the axis change: the right-hand scale and the *ASIC Temp* legend swatch go from the same grey as the gridlines to the group's own colour.

They do **not** show the fill change, and I would rather say so than imply otherwise — the shots are rendered against the fixture in `home.component.spec.ts` with the API stubbed, and I could not get the stub to feed the three rolling averages data that diverges from the raw figure, so only one series is drawn in both. On a device the four differ and the stacking is what you see.

`--chart-axis2-color` falls back to `--color-text-secondary` when unset, so a theme that does not define it renders exactly as today.


0xf0xx0:
nack in current form, it should pick one of the active datapoints to fill instead of just the furst

SurefireStudios:
Good catch. With `index === 0` the fill disappeared entirely when that series was hidden, instead of moving to one still on screen. Both default groups happen to lead with a visible series (`hashrate`, `asicTemp`), so it only bit once you toggled the leading one off from the legend.

Now picks the first entry with `hidden === false`. The legend handler mutates `hidden` in place without rebuilding the datasets, so `refreshChartFill()` re-picks the target on toggle and calls `ci.update()` to repaint; `borderColor` still keys off the original index, so a series keeps its colour whatever is hidden. Checked the selection across all-visible, first-hidden, first-two-hidden, all-hidden, single-series and the unfilled y2 axis — never more than one fill, none when everything is hidden. `npm run build` clean.

PR #1981, opened 2026-09-16
fix(axe-os): stop the hashrate group stacking four fills, and colour the second axis

SurefireStudios:
Split out of #1977 at @0xf0xx0's request. Two files, no behaviour outside the chart.

## The hashrate group stacks four fills

`createChartDatasets` passes its `fill` argument to every series in a group. The hashrate group has four — raw, 1m, 10m, 1h — so four translucent reds stack on one another and merge into a wash that hides the lines and the grid behind it.

Only the leading series in a group is filled now:

```ts
const filled = fill && index === 0;
```

The temperature group already passed `fill: false`, so it is unaffected.

## The temperature group mixed toward black

Each successive series in a group is mixed toward `mixColor`, and the temperature group used `'black'` — on an already near-black card, so the later series went invisible. They now mix toward the body text colour, which is legible on both the dark and white themes.

## The second axis had no colour of its own

That group also drew in `--color-text-secondary`, the same grey as the axis furniture, so on a dual-axis chart nothing tied the lines to the numbers on the right.

A new `--chart-axis2-color` gives the group and its axis one colour — `#7bd8c0` on dark, `#0b7f68` on the white theme — distinct from the accent that draws the hashrate group.

| Before | After |
| --- | --- |
| Chart before | Chart after |

Those shots show the axis change: the right-hand scale and the *ASIC Temp* legend swatch go from the same grey as the gridlines to the group's own colour.

They do **not** show the fill change, and I would rather say so than imply otherwise — the screenshots are rendered against the fixture in `home.component.spec.ts` with the API stubbed, and I could not get the stub to feed the three rolling averages data that diverges from the raw figure, so only one series is drawn in both. On a device the four differ and the stacking is what you see.

## Checks

`ng build` passes.

`--chart-axis2-color` falls back to `--color-text-secondary` if it is ever missing, so a theme that does not define it renders exactly as today.


0xf0xx0:
dupe of #1979

SurefireStudios:
Duplicate of #1979 — same fix, same two files, opened a few hours earlier. My mistake; closing this one and keeping #1979, which I've updated with the before/after shots from here.

Sorry for the noise, @0xf0xx0.

PR #1984, opened 2026-09-18
fix(http_server): add headroom to URI handler table

nokiexow:
max_uri_handlers is 25 and the server registers exactly 25 handlers today (23 in http_server.c plus the 2 theme endpoints), so the table has no headroom left. The /* catch-all is registered last, so the next handler anyone adds ahead of it fails to register silently -- the return value of httpd_register_uri_handler is not checked at that call site -- and the device then serves the captive-portal redirect instead of the UI. Raising the limit to 32 costs nothing at runtime (the table is a single array of handler pointers) and restores the margin.


mutatrum:
Would make sense to also check the failure on all `httpd_register_uri_handler` call sites.

awi81:
Field data point in favour of this: we hit it on a Gamma 601 running a v2.15.0rc1-based build. Adding one endpoint made 26 registrations against the limit of 25, and the one that failed was the last, `/*`. Its return value isn't checked, and unmatched requests fall through to the captive-portal 404 handler, so the result was a dashboard stuck in a 302 loop to `/` while the API kept working, with nothing in the log.

We've run with 32 since 2026-08-21. +1 to checking every `httpd_register_uri_handler()` return and logging the URI that failed. That log line would have saved us the search.

PR #1985, opened 2026-09-18
fix(ci): handle missing unit test reporting artifacts

vortexopenclaw:
Prevent the test-results publisher from crashing when a run produces no artifacts. Upload the event file before building, and skip publishing with an explanatory notice if the event or XML reports are missing. Actual test failures still publish normally. Missing-artifact cases tested; all CI checks pass.


johnny9 (approved):
ACK 6312d9abdc3768a499f117ebec477b71cdbb6a67

Looks good. Missing artifacts are handled correctly, failing test reports still publish, and CI passes.

0xf0xx0 (approved):
ack

PR #1989, opened 2026-09-20
fix(asic): bound thermal recovery and reject invalid PLL settings

nokiexow:
## Summary
Fix the frequency-recovery arithmetic and handle unsatisfiable PLL requests without programming zero divisors or replacing the last valid transition frequency with zero.

## Reproduction
The existing overheat recovery expression, `f > 100 ? f - 100 : 400`, can increase frequency precisely when cooling is needed (`50 → 400`, `100 → 400`) or produce an unsupported low clock (`100.25 → 0.25`, `125 → 25`).

For BM1370, targets such as 1, 25 or 47 MHz have no tuple within the existing PLL search bounds. Previously `pll_get_parameters()` returned zero frequency/dividers without a failure indication, and the driver continued constructing register values.

## Changes
- Recovery subtracts 100 MHz with a floor of **50 MHz**, reusing the clock already used by the existing shutdown ramp. This is not a newly certified electrical operating limit.
- Recovery must strictly reduce frequency. At/below the floor or for invalid values, leave the ASIC powered down, latch a hardware fault, and keep overheat mode enabled rather than restarting at a higher clock. Cooling inspection/operator intervention is required.
- Make PLL lookup return success/failure; leave outputs untouched on failure. Reject invalid targets, pointers and divider bounds; check rounded divider values before narrowing to an integer.
- All five ASIC drivers return before register writes when PLL lookup fails.
- Frequency transitions stop on an invalid callback result and retain the last successfully applied clock. Reject nonfinite/out-of-range transition inputs before integer step calculations.
- BM1366/BM1368/BM1370 initialization uses the applied frequency, not the requested value, for nonce-space timing.

## Regression coverage
- Valid PLL settings across all three feedback-divider ranges; invalid/nonfinite/very large targets leave outputs unchanged.
- UART-spy tests for all five ASIC families: invalid targets produce no register packets; ordinary valid targets still program successfully.
- Failed short/intermediate/final frequency transitions preserve the last valid clock.
- Recovery boundary vectors and a quarter-MHz sweep verify strict decrease and the floor; invalid/floor inputs cannot request a restart.

## Validation and scope
- ESP-IDF v6.0.2: `test-ci` build and full firmware build.
- ESP32-S3 QEMU: **135 tests, 0 failures**, on this independent branch.
- Frontend production bundle built and embedded through the CI build path.
- No physical ASIC/thermal validation or flashing. Hardware follow-up is still needed, particularly for recovery at low clocks.
- This is complementary to #1864: it does **not** duplicate user-setting clamping or establish model-specific overclocking limits. A rejected ramp can have already applied earlier valid steps; preserving the last accepted step is not a guarantee that every user-selected frequency is electrically safe.


mutatrum:
I don't know if you're working on more PRs, but please pause. If you have more suggestions, either first open an issue or join Discord to discuss. We cannot process unlimited code changes.

nokiexow:
ok

